PR #14, opened 2026-02-06
fix: run MCP tools in the caller project directory

BP602:
## Why
When the proxy was started from its own repository, MCP tools (read/write/edit/bash/glob/grep) inherited the proxy process cwd. OpenCode sessions launched from a different project were then operating on the wrong directory, causing confusion and incorrect file/tool behavior.

## What changed
- MCP tools now resolve relative paths from `CLAUDE_PROXY_WORKDIR` when set
- fallback remains `process.cwd()` for backward compatibility
- README updated to document `CLAUDE_PROXY_WORKDIR`

## Result
- proxy can run from one location while tools correctly target the intended project
- predictable per-project behavior without relocating the proxy checkout


BP602:
@rynfar I realize now the readme change might have been a bit too light, see #18

PR #52, opened 2026-03-20
docs: improves README

skipships:
Tweaked language, cut down verbosity, fixed broken links, ran through Markdown formatter


skipships:
The introduction of the per-instance proxy is significant so I can take another pass at explaining the differences and organizing the README. Would be cool to figure out a way to use Desktop with the per-instance-proxy too.

Issue #46, opened 2026-03-20
suggestion: create new proxy per terminal


calebdw:
Hello!

I created a fish function below that starts a new proxy in the current directory, this should prevent "concurrent requests" and keeps the opencode instances segregated---do with it what you will

```fish
# Start a dedicated proxy per OpenCode instance so multiple projects can run
# at once without colliding on a shared fixed port or workdir.
function oc
    set -l proxy "$HOME/sources/opencode-claude-max-proxy/bin/claude-proxy.ts"

    if not test -f $proxy
        echo "Proxy script not found: $proxy" >&2
        return 1
    end

    set -l port (python3 -c 'import socket; s = socket.socket(); s.bind(("127.0.0.1", 0)); print(s.getsockname()[1]); s.close()')

    env CLAUDE_PROXY_IDLE_TIMEOUT_SECONDS=255 CLAUDE_PROXY_PORT=$port CLAUDE_PROXY_WORKDIR="$PWD" bun run "$proxy" &
    set -l proxy_pid $last_pid

    # Wait briefly for the proxy before pointing OpenCode at it.
    for attempt in (seq 1 50)
        curl -fsS "http://127.0.0.1:$port/" >/dev/null 2>&1
        and break
        sleep 0.1
    end

    env ANTHROPIC_API_KEY=dummy ANTHROPIC_BASE_URL="http://127.0.0.1:$port" opencode $argv
    set -l opencode_status $status

    kill $proxy_pid >/dev/null 2>&1
    wait $proxy_pid >/dev/null 2>&1

    return $opencode_status
end
```

rynfar:
> Hello!
> 
> I created a fish function below that starts a new proxy in the current directory, this should prevent "concurrent requests" and keeps the opencode instances segregated---do with it what you will
> 
> # Start a dedicated proxy per OpenCode instance so multiple projects can run
> # at once without colliding on a shared fixed port or workdir.
> function oc
>     set -l proxy "$HOME/sources/opencode-claude-max-proxy/bin/claude-proxy.ts"
> 
>     if not test -f $proxy
>         echo "Proxy script not found: $proxy" >&2
>         return 1
>     end
> 
>     set -l port (python3 -c 'import socket; s = socket.socket(); s.bind(("127.0.0.1", 0)); print(s.getsockname()[1]); s.close()')
> 
[...]

PR #53, opened 2026-03-20
feat: per-terminal proxy launcher and shared session store

rynfar:
## Summary

Two features that work together to solve concurrency and session resume:

### Per-Terminal Launcher (`bin/oc.sh`)
Each terminal gets its own proxy on a random port. No shared port, no concurrent crashes, no supervisor needed.

```bash
./bin/oc.sh        # starts proxy, launches OpenCode, cleans up on exit
```

### Shared Session File Store
Sessions persist to `~/.cache/opencode-claude-max-proxy/sessions.json` so they can be resumed across terminals — even when each terminal has its own proxy.

- Atomic writes for safety
- 24-hour TTL with auto-pruning
- In-memory cache as fast path, file as fallback
- `clearSessionCache()` also clears file store (no test leaks)

### Who Uses What

| Use Case | Approach | Concurrency | Session Resume |
|---|---|---|---|
| **CLI (multiple terminals)** | `oc.sh` per-terminal | ✅ No crashes | ✅ Shared file store |
| **OpenCode Desktop** | Shared proxy + config file | ⚠️ Use supervisor | ✅ Same file store |
| **Single terminal** | Either | ✅ No issue | ✅ Both work |

### Verified
- Two proxies on different ports: Proxy A creates session → Proxy B resumes it ✅
- Launcher starts proxy, connects OpenCode, cleans up on exit ✅
- 114 tests (8 new for session store), all passing ✅

Credit to @calebdw (#46) for the per-terminal proxy idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,28 +102,41 @@ curl http://127.0.0.1:3456/health
 
 ### Connect OpenCode
 
-#### Environment Variables
+#### Per-Terminal Launcher (recommended)
 
 ```bash
+./bin/oc.sh
+```
+
+Each terminal gets its own proxy on a random port. No port conflicts, no concurrency crashes. The proxy starts automatically, connects OpenCode, and cleans up when you exit. Sessions resume across terminals via a shared session file.
+
+Add to your shell config for easy access:
+
+```bash
+# ~/.zshrc or ~/.bashrc
+alias oc='/path/to/opencode-claude-max-proxy/bin/oc.sh'
+```
+
+#### Shared Proxy
+
+If you prefer a single long-running proxy:
+
+```bash
+# Terminal 1: start the proxy
+CLAUDE_PROXY_PASSTHROUGH=1 bun run proxy
+
+# Terminal 2+: connect OpenCode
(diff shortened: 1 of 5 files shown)


TxHawks:
Does this make [`opencode-with-claude`](https://github.com/ianjwhite99/opencode-with-claude) redundant? If not, what’s the essential difference in functionality other than opencode spinning up the proxy server vs the proxy server spinning `opencode`,

rynfar:
Not redundant — they solve different problems.

Meridian is a standalone HTTP proxy. It translates Anthropic API requests into Claude Max SDK calls. Any tool that speaks the Anthropic API format can point at it — OpenCode, Cursor, Claude CLI, custom scripts, whatever. It doesn't care what's on the other end.

 opencode-with-claude bundles both together for convenience — it spins up the proxy and launches OpenCode in one command so you don't have to manage two processes. It's an opinionated wrapper for people who just want OpenCode + Claude Max with zero setup.

 If you only use OpenCode, opencode-with-claude is simpler. If you use multiple clients or want the proxy running as a service (launchd, Docker, systemd), Meridian standalone is what you want.
 
 Like in my case im using droid, opencode and cline so i dont use the opencode-with-claude plugin.  @TxHawks  let me know if this makes sense.

PR #63, opened 2026-03-20
fix: replace ubuntu base image with multi-stage node:22 build to fix Docker build failures

ianjwhite99:
## Problem

`docker compose up -d` fails during the build with `apt-get` errors on certain networks:

```
E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/noble/InRelease
  Clearsigned file isn't valid, got 'NOSPLIT' (does the network require authentication?)
E: The repository 'http://archive.ubuntu.com/ubuntu noble InRelease' is not signed.
```

The `ubuntu:24.04` base image requires `apt-get` to install every dependency (curl, unzip, git, ca-certificates, nodejs, npm). Any network that intercepts or modifies HTTP responses (corporate proxy, VPN, firewall, Docker Desktop DNS quirks) corrupts Ubuntu's GPG-signed APT metadata and makes the build completely unusable.

## Solution

Eliminate `apt-get` entirely by switching to `node:22` (which already ships node, npm, curl, git, ca-certificates) and installing bun via `npm install -g bun` instead of the shell script (which needed `unzip` + `curl`).

### Additional optimizations

- **Multi-stage build** — `node:22` build stage + `node:22-slim` runtime stage, reducing the final image size by ~800MB
- **Production-only deps** — `bun install --production` skips devDependencies (typescript, @types/*) in the image
- **Selective COPY** — only `bin/`, `src/proxy/`, `src/plugin/`, `src/logger.ts`, `src/mcpTools.ts` are copied instead of `COPY . .`, excluding tests and docs
- **Merged RUN layers** — bun install, npm cache cleanup, and user creation consolidated into one layer
- **Bun-based healthcheck** — `curl` replaced with `bun -e "fetch(...)"` in both `Dockerfile` and `docker-compose.yml`, so `curl` is no longer needed in the runtime image
- **Better `.dockerignore`** — excludes `.github/`, `*.md`, `tsconfig.json`, `src/__tests__/`, and Docker config files from the build context

## Files changed

- `Dockerfile` — multi-stage build, no apt-get, production deps, selective COPY
- `docker-compose.yml` — healthcheck switched from curl to bun fetch
- `.dockerignore` — expanded exclusions for smaller build context

## Test plan

- [ ] `docker compose up -d --build` completes successfully
- [ ] `docker compose ps` shows proxy healthy
- [ ] Proxy responds on port 3456 (`curl http://localhost:3456/health`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,15 +1,16 @@
 services:
   # Fix volume permissions on first run (volumes created as root)
   init:
-    build: .
+    image: busybox
     user: root
     volumes:
       - claude-auth:/home/claude/.claude
-    entrypoint: ["bash", "-c", "chown -R claude:claude /home/claude/.claude"]
+    entrypoint: ["sh", "-c", "chown -R 1000:1000 /home/claude/.claude"]
     restart: "no"
 
   proxy:
     build: .
+    image: claude-max-proxy
     user: "claude"
     depends_on:
       init:
@@ -18,14 +19,6 @@ services:
       - "3456:3456"
     volumes:
       - claude-auth:/home/claude/.claude
-    environment:
-      - CLAUDE_PROXY_PASSTHROUGH=1
-      - CLAUDE_PROXY_HOST=0.0.0.0
-    healthcheck:
-      test: ["CMD", "bun", "-e", "const r=await fetch('http://127.0.0.1:3456/health');process.exit(r.ok?0:1)"]
-      interval: 30s
-      timeout: 5s
-      retries: 3
     restart: unless-stopped
 
 volumes:


rynfar:
Nice fix — the multi-stage build and dropping apt-get entirely is a solid improvement.

One thing: the selective COPY in the runtime stage is missing `src/plugin/`. That directory has `claude-max-headers.ts` (the OpenCode session header plugin), which is referenced in the README and listed in package.json `files`. Docker users who try to use the plugin won't find it in the container.

Should just need an additional line:
```dockerfile
COPY --chown=claude:claude src/plugin/ ./src/plugin/
```

Everything else looks good.

ianjwhite99:
Should already be covered with line 38 on the Dockerfile

```dockerfile
COPY --chown=claude:claude src/plugin/ ./src/plugin/
```

ianjwhite99:
Made some optimizations to the `docker-compose.yml` as well. [a737190](https://github.com/rynfar/opencode-claude-max-proxy/pull/63/commits/a737190449d1e0feaa05c6b6d23c1affda05e08f)

rynfar:
Excellent work!  Thank you for working on this!

PR #77, opened 2026-03-21
fix(proxy): convert blocking execSync calls to async

conversun:
## Problem
\`execSync\` calls block the event loop — \`which claude\` at startup, \`claude auth status\` on every health check (up to 5s).

## Fix
- Convert to async \`exec\` with \`util.promisify\`
- Lazy-init + cache for claude path resolution
- Health endpoint now async with preserved timeout

## Test
\`bun test\` — all existing + new \`proxy-async-ops.test.ts\`

### DIFF
--- a/bin/cli.ts
+++ b/bin/cli.ts
@@ -1,7 +1,10 @@
 #!/usr/bin/env node
 
 import { startProxyServer } from "../src/proxy/server"
-import { execSync } from "child_process"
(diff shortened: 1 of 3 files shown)


rynfar:
great catches - going to merge all 3 of yours, i have a couple other changes ill make afterwards but conceptually these are great improvements thank you!

PR #80, opened 2026-03-22
fix: pass OpenCode system prompt via SDK appendSystemPrompt

### LINKED ISSUE #74: global AGENTS.md not being read?
Hello!

I just noticed that some of my commands in my global opencode AGENTS.md were not being followed... is this being passed to the api?

Thanks!

rynfar:
Fixes #74

**Problem:** OpenCode's system prompt (which includes AGENTS.md and custom instructions) was being injected into the prompt text as a string prefix or synthetic user message. Claude treats user messages with less authority than system-level instructions, so AGENTS.md directives were effectively ignored.

**Fix:** Pass the system context via the SDK's documented `systemPrompt` option:
```ts
systemPrompt: { type: 'preset', preset: 'claude_code', append: systemContext }
```

This:
- ✅ Preserves the full Claude Code default system prompt (unchanged)
- ✅ Appends user instructions at the system-prompt level where they carry proper weight
- ✅ Uses the SDK's public API (`preset` + `append`), equivalent to the `--append-system-prompt` CLI flag
- ✅ Applied to both streaming and non-streaming paths

149/150 tests pass.

### DIFF
--- a/src/__tests__/proxy-multimodal.test.ts
+++ b/src/__tests__/proxy-multimodal.test.ts
@@ -149,7 +149,7 @@ describe("Multimodal content", () => {
       messages.push(msg)
     }
 
-    // Should have system context + all 3 messages
+    // Should have all 3 messages (system context now in SDK option, not in prompt)
     expect(messages.length).toBeGreaterThanOrEqual(3)
     // All should have the user type wrapper (SDK requirement)
     for (const msg of messages) {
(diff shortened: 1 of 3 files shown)


calebdw:
Just be careful that there's none of OpenCode specific stuff included that would cause Anthropic to block. Have you inspected the prompt?

PR #87, opened 2026-03-22
Enable 1M context window for Opus via opus[1m] alias

pablogtzgileta:
Currently the proxy maps all Opus requests to just `"opus"`, which defaults 
to a 200K context window. The problem is that OpenCode gets stuck when it 
hits compaction at 200K — you get "Conversation history too large to compact - 
exceeds model context limit" and the session is effectively dead.

Claude Max subscribers already have access to the 1M context window for 
Opus 4.6 through the `opus[1m]` alias, but the proxy strips it in 
`mapModelToClaudeModel`. Using the full 1M context avoids hitting compaction 
entirely for most sessions, which keeps things running smoothly.

The change is small — just return `"opus[1m]"` instead of `"opus"` in the 
two model mapping functions.

### DIFF
--- a/src/__tests__/proxy-agent-definitions.test.ts
+++ b/src/__tests__/proxy-agent-definitions.test.ts
@@ -95,8 +95,8 @@ describe("buildAgentDefinitions", () => {
 
 describe("mapModelTier", () => {
   it("should map opus models", () => {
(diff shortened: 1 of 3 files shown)


rynfar:
Great catch thank you for this.

PR #90, opened 2026-03-23
fix: remove mock.module leak that breaks session store tests

conversun:
## Summary

- `session-lineage.test.ts` used top-level `mock.module("../proxy/sessionStore", ...)` which replaced the sessionStore module globally in Bun's module registry
- This caused **10 test failures** in `proxy-session-store.test.ts` (6) and `proxy-session-store-locking.test.ts` (4) when running the full suite — tests passed individually but failed together
- Root cause: `lookupSharedSession` always returned `undefined` because the mocked version was served to all test files

## Fix

Replace the module mock with a real tmpDir-backed sessionStore (matching the pattern used by other test files):
- Set `CLAUDE_PROXY_SESSION_DIR` to an isolated temp directory
- Call `clearSharedSessions()` in `beforeEach` to prevent cross-test interference
- Clean up tmpDir and env var in `afterAll`

## Verification

`bun test` — 177 pass, 0 fail (was 167 pass, 10 fail before fix)

### DIFF
--- a/src/__tests__/session-lineage.test.ts
+++ b/src/__tests__/session-lineage.test.ts
@@ -6,6 +6,9 @@
  */
 
 import { afterAll, beforeEach, describe, expect, it, mock } from "bun:test"
+import { mkdtempSync, rmSync } from "node:fs"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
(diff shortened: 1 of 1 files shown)


rynfar:
excellent catch thank you for this!

PR #92, opened 2026-03-23
fix: prevent cross-project session contamination in fingerprint cache

conversun:
## Summary

First of all — thank you for building this project. Claude Code on its own left a lot to be desired, and this proxy has genuinely saved my workflow. I've been using it daily across multiple projects and it's been rock solid.

While running multiple OpenCode instances pointed at the same proxy (different projects, different terminals), I noticed sessions occasionally resuming in the wrong project context. This PR fixes the root cause.

## Problem

Two issues combined to cause intermittent cross-project session contamination:

1. **getConversationFingerprint() had no project scoping** — it hashed only the first user message. Different projects with similar/identical first messages (e.g., system-generated greetings) produced the same fingerprint, causing the proxy to resume the wrong SDK session.

2. **storeSession() always wrote to both caches** — even when a header-based session ID existed, it also wrote a fingerprint entry. These phantom fingerprint entries could match requests from a completely different project.

## Fix

1. Include systemContext (which contains project-specific content like CLAUDE.md paths, working directory, custom instructions) in the fingerprint hash — different projects now get different fingerprints even with identical first messages.

2. Write to sessionCache OR fingerprintCache, never both — fingerprint is a fallback for when the OpenCode plugin isn't installed, not a parallel store.

**Net delta: -2 lines across 2 files. All 110 existing tests pass.**

### DIFF
--- a/src/proxy/server.ts
+++ b/src/proxy/server.ts
@@ -114,8 +114,10 @@ setInterval(() => {
   }
 }, 60 * 60 * 1000)
 
-/** Hash the first user message to fingerprint a conversation */
-function getConversationFingerprint(messages: Array<{ role: string; content: any }>): string {
+/** Hash the first user message + system context to fingerprint a conversation.
+ *  Including system context prevents cross-project collisions when different
+ *  projects happen to start with the same first message. */
+function getConversationFingerprint(messages: Array<{ role: string; content: any }>, systemContext?: string): string {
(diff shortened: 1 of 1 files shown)


rynfar:
Absolutely!  I'm glad it's proving effective for people!  You've been a big contributor so thank you for submitting issues so we can make this a quality replacement!

rynfar:
Excellent catch once again!

PR #95, opened 2026-03-23
fix: pass systemContext to storeSession for consistent fingerprinting

### LINKED ISSUE #94: Tool call leakage: raw tool_use blocks appear as text instead of being executed
## Description

Tool calls from the model are leaking as raw text into the response instead of being properly intercepted and executed by OpenCode. The tool call markup appears inline in the assistant's output rather than triggering actual tool execution.

This is not related to OpenCode — reproduces on both v1.2.27 and v1.3.0.

## Steps to Reproduce

1. Use proxy v1.13.3
2. Start an OpenCode session through the proxy and perform a task that requires tool calls (e.g. file edits, bash commands)
3. Observe that tool calls appear as raw text in the response instead of being executed

## Expected Behavior

Tool calls should be intercepted by OpenCode and executed normally, with only the tool results shown to the user.

## Actual Behavior

Tool call blocks (e.g. `[Tool Use: bash({...})]`) are rendered as plain text in the assistant's response. The tools are not actually invoked.

## Workaround

Pinned proxy back to v1.12.2 which does not exhibit this issue.

## Environment

- Proxy version: v1.13.3 (works on v1.12.2)
- OpenCode versions tested: 1.2.27, 1.3.0
- OS: macOS

davdroman:
## Summary

- Fixes fingerprint mismatch between `storeSession` and `lookupSession` that prevented session resume via fingerprint
- Adds missing `systemContext` argument to the stream-path `storeSession` call

## Problem

`lookupSession` computes fingerprints with `systemContext` (introduced in v1.13.0 for cross-project collision prevention), but `storeSession` computes them without it. The hashes never match, so fingerprint-based session resume silently fails. Every request is treated as new, causing the full message history — including tool_use blocks serialized as `[Tool Use: ...]` text — to be re-sent, which the model then echoes back as raw text instead of executing.

Fixes #94

### DIFF
--- a/src/proxy/server.ts
+++ b/src/proxy/server.ts
@@ -225,7 +225,8 @@ function lookupSession(
 function storeSession(
   opencodeSessionId: string | undefined,
   messages: Array<{ role: string; content: any }>,
-  claudeSessionId: string
+  claudeSessionId: string,
+  systemContext?: string
 ) {
   if (!claudeSessionId) return
   const lineageHash = computeLineageHash(messages)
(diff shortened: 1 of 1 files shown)


rynfar:
Merging then im going to add test coverage to prevent regression on this in the future. Thank you for catching this.

PR #96, opened 2026-03-23
feat!: return ProxyInstance from startProxyServer for graceful shutdown and library usage

ianjwhite99:
## Summary

- **Refactors `startProxyServer()`** to return a structured `ProxyInstance` object (`{ server, config, close() }`) instead of a raw `http.Server`, enabling programmatic consumers to gracefully shut down the proxy and access resolved config.
- **Adds `silent` config option** to suppress console output, useful for library/embedded usage and testing.
- **Extracts session cleanup into `startSessionCleanup()`** with per-instance timers that are properly cleaned up on `close()`, preventing timer leaks.
- **Moves `process.exit(1)` for EADDRINUSE from library to CLI layer** — the library now throws instead, letting callers decide how to handle port conflicts.
- **Exports types via `./types` subpath** in `package.json` so consumers can import types directly.

## Breaking Changes

`startProxyServer()` return type changed from `http.Server` to `ProxyInstance`. Consumers need to update:

```ts
// Before
const server = await startProxyServer({ port: 3456 })
server.close()

// After
const proxy = await startProxyServer({ port: 3456 })
await proxy.close()        // graceful shutdown
proxy.server.close()       // or access raw server
```

## Files Changed

| File | Change |
|------|--------|
| `src/proxy/types.ts` | Added `ProxyInstance` interface, `silent` to `ProxyConfig` |
| `src/proxy/server.ts` | Refactored return type, extracted cleanup, silent mode, error handling |
| `bin/cli.ts` | Handles `ProxyInstance`, EADDRINUSE exit at CLI layer |
| `package.json` | Added `./types` subpath export |
| `src/__tests__/proxy-async-ops.test.ts` | Updated for `ProxyInstance` API |

### DIFF
--- a/src/proxy/server.ts
+++ b/src/proxy/server.ts
@@ -1405,8 +1405,6 @@ export async function startProxyServer(config: Partial<ProxyConfig> = {}): Promi
       console.error(`\nOr use a different port:`)
       console.error(`  CLAUDE_PROXY_PORT=4567 claude-max-proxy`)
     }
-    // Don't re-throw — let the error propagate on the server's event emitter.
-    // Callers can handle it via proxy.server.on("error", ...).
   })
 
   return {


rynfar:
Thanks for this — really appreciate you pushing the project toward proper library ergonomics. The `ProxyInstance` return type, `silent` mode, and moving `process.exit` out of the library are exactly the right direction.

Two things I'd like addressed before merging:

**1. Double cleanup timer**
`_defaultCleanupTimer` fires at module load (backward compat), and `startSessionCleanup()` is called again inside `startProxyServer`. So after startup there are two intervals pruning the same caches. It's harmless (idempotent) but wasteful. Could you either clear the module-level timer when `startProxyServer` starts its own, or remove the module-level one entirely?

**2. EADDRINUSE throw from event handler**
The `throw err` inside `server.on("error")` will become an uncaught exception since it's in an event callback, not in the async flow. Library consumers would crash unexpectedly. Better to just let the error propagate on the server's event emitter (don't re-throw) — the CLI already handles it with `proxy.server.on("error")` + `process.exit(1)`. Library consumers can do the same.

Everything else looks good — tests updated, types exported, clean interface.

ianjwhite99:
No problem, I've got that cleaned up. I have a couple thoughts on modularizing some of the code in `server.ts` to break things out into more serviceable components. I'll probably share my thoughts on that in a separate issue for you to take a glance at.

rynfar:
Excellent I appreciate it.  I have some big refactoring I want to do at some point to reduce the ai slop surface area.  But I want to be careful not to break what is currently working well for people.

ianjwhite99:
For sure. Appreciate all your hard work!

PR #103, opened 2026-03-23
fix: export TypeScript declaration files from distFix/types export

ianjwhite99:
## Summary

- Add `tsconfig.build.json` for declaration-only emit, scoped to the `server.ts` entry point
- Append `tsc -p tsconfig.build.json` to the build script to generate `.d.ts` files alongside JS bundles
- Add `types` field and conditional `exports` with `types` condition in `package.json`
- Add `ProxyServer` interface to avoid leaking `hono` types (a devDependency) in the public API
- Postbuild now verifies `dist/proxy/server.d.ts` exists

Consumers can now import types directly:

```ts
import { createProxyServer, startProxyServer } from "opencode-claude-max-proxy"
import type { ProxyConfig, ProxyInstance, ProxyServer } from "opencode-claude-max-proxy"

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,10 @@
     "claude-max-proxy": "./dist/cli.js"
   },
   "exports": {
-    ".": "./dist/server.js",
-    "./types": "./dist/types.js"
+    ".": {
+      "types": "./dist/proxy/server.d.ts",
+      "default": "./dist/server.js"
+    }
   },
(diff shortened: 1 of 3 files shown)


ianjwhite99:
@rynfar My bad. Types weren't exported in #96. Resolved in #103.

rynfar:
no problem!

PR #139, opened 2026-03-25
fix: use subscription type to determine sonnet model variant

jaker-4c:
# Observed behavior

Team subscription works with `claude -p --model sonnet ...`
The proxy maps Sonnet requests to `sonnet[1m]`
`claude -p --model 'sonnet[1m]' ...` returns `API Error: Rate limit reached`

- Proxy rewrites that into `Claude Max rate limit reached. Wait a moment and try again.`
- `claude auth status` reports `subscriptionType: "team"` in this environment
- `opus[1m]` works here, while `sonnet[1m]` does not

## Proposed fix

- Read `subscriptionType` from `claude auth status`
- If `subscriptionType === "max"`, prefer `sonnet[1m]`
- Otherwise use plain `sonnet`
- If a preferred `sonnet[1m]` request fails with access/rate-limit/model-availability errors, retry once with plain `sonnet`
- Preserve upstream error details instead of always rewriting to a Max rate-limit message

## Simpler fix

```ts
export function mapModelToClaudeModel(model: string): ClaudeModel {
  if (model.includes("opus")) return "opus[1m]";
  if (model.includes("haiku")) return "haiku";
  return "sonnet"; // "[1m]" removed
}
```

### DIFF
--- a/src/__tests__/models.test.ts
+++ b/src/__tests__/models.test.ts
@@ -1,10 +1,17 @@
 /**
  * Unit tests for model mapping and utility functions.
  */
-import { describe, it, expect } from "bun:test"
-import { mapModelToClaudeModel, isClosedControllerError } from "../proxy/models"
+import { afterEach, describe, it, expect } from "bun:test"
+import { mapModelToClaudeModel, isClosedControllerError, resetCachedClaudeAuthStatus } from "../proxy/models"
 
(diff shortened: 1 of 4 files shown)


rynfar:
great catch thank you!

I am going to merge it and add a test.

PR #149, opened 2026-03-25
fix: detect rate-limited accounts and fall back from 1m models

ianjwhite99:
## Problem

The proxy previously hard-coded `[1m]` variants for all Opus and Sonnet models, regardless of version. This caused two problems:

1. **Older models broke entirely** — requesting `claude-opus-4-5` or `claude-sonnet-4-5` would map to `opus[1m]` or `sonnet[1m]`, which don't exist for 4.5 models, causing requests to fail outright.
2. **Rate-limited accounts couldn't fall back** — international users and accounts with org-level restrictions receive a `rate_limit_event` with `status: "rejected"` when requesting `[1m]` variants, with no recovery path.

## Solution

This change adds version-aware model mapping (only 4.6+ models get `[1m]`) and automatic fallback when the `[1m]` variant is rejected at runtime.

### Changes

**`src/proxy/models.ts`**
- Added `supports1mContext()` — only Claude 4.6+ models get the `[1m]` suffix; 4.5 and older map to the base variant
- `mapModelToClaudeModel()` now uses version-aware mapping while preserving subscription-type checks for sonnet
- New exports: `stripExtendedContext()`, `hasExtendedContext()`

**`src/proxy/errors.ts`**
- Added `isRateLimitError()` — detects rate limit errors from error messages, used to trigger fallback from `[1m]` to base model

**`src/proxy/server.ts`**
- Both streaming and non-streaming paths detect rate limit errors during queries
- On rejection: if the current model has `[1m]`, strips it and retries with the base model
- Compatible with the existing stale-session retry wrapper

**`src/__tests__/models.test.ts`**
- Tests for version-aware mapping (4.5 → base, 4.6 → `[1m]`)
- Tests for `stripExtendedContext()` and `hasExtendedContext()`

## How it works

1. **Version check at map time**: `claude-opus-4-5` → `opus`, `claude-opus-4-6` → `opus[1m]`
2. **Runtime fallback**: if `opus[1m]` gets a rate limit rejection, retry with `opus`

The detected event format:
```json
{"type": "rate_limit_event", "rate_limit_info": {"status": "rejected", "overageDisabledReason": "org_level_disabled"}}
```

### DIFF
--- a/src/__tests__/models.test.ts
+++ b/src/__tests__/models.test.ts
@@ -2,7 +2,7 @@
  * Unit tests for model mapping and utility functions.
  */
 import { afterEach, beforeEach, describe, it, expect, mock } from "bun:test"
-import { mapModelToClaudeModel, isClosedControllerError, resetCachedClaudeAuthStatus, getClaudeAuthStatusAsync } from "../proxy/models"
+import { mapModelToClaudeModel, isClosedControllerError, resetCachedClaudeAuthStatus, getClaudeAuthStatusAsync, stripExtendedContext, hasExtendedContext } from "../proxy/models"
 
 describe("mapModelToClaudeModel", () => {
   const originalSonnetModel = process.env.CLAUDE_PROXY_SONNET_MODEL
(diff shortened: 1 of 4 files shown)


ianjwhite99:
See issue [#43](https://github.com/ianjwhite99/opencode-with-claude/issues/43#issuecomment-4130229348)

PR #161, opened 2026-03-26
feat: add profile-based routing foundation

davetha:
## Summary
- add a profile-based routing foundation so Meridian can distinguish between multiple Claude contexts instead of assuming one global account/session space
- scope session lookup and resume by profile, and allow profile selection via `x-meridian-profile`
- add a local example for running one Meridian instance with personal/company profiles on the same port

## Use Case
I have both a personal and a company Claude account/plan, and I want to use OpenCode from several systems without having to install and auth Meridian separately on each machine.

The goal is to run a single Meridian instance that I can reach over my local network or a private Tailscale tunnel, then route requests to the right Claude profile from each client. That gives me better control over which systems use which account, while still feeling similar to how each machine might otherwise have its own `claude auth` setup.

This PR only adds the profile-routing foundation. It does not add request authentication or admin-route protection yet.

## Testing
- `bun test src/__tests__/proxy-profiles.test.ts`
- `bun test src/__tests__/proxy-stale-uuid-retry.test.ts`
- `npm run build`

## Notes
Follow-up PRs will keep the maintainer review small:
1. config file loading + optional request API keys
2. optional protection for `/health` and `/telemetry*`

### DIFF
--- a/examples/profile-routing-local.ts
+++ b/examples/profile-routing-local.ts
@@ -1,0 +1,33 @@
+import { startProxyServer } from "../src/proxy/server"
+
+const port = Number.parseInt(process.env.CLAUDE_PROXY_PORT || "3456", 10)
+const host = process.env.CLAUDE_PROXY_HOST || "127.0.0.1"
+
+const personalDir = process.env.MERIDIAN_PERSONAL_CLAUDE_DIR
+const companyDir = process.env.MERIDIAN_COMPANY_CLAUDE_DIR
+
+if (!personalDir || !companyDir) {
+  throw new Error("Set MERIDIAN_PERSONAL_CLAUDE_DIR and MERIDIAN_COMPANY_CLAUDE_DIR before running this example")
+}
+
+const proxy = await startProxyServer({
+  port,
+  host,
+  profiles: [
+    { id: "personal", claudeConfigDir: personalDir },
+    { id: "company", claudeConfigDir: companyDir },
+  ],
+  defaultProfile: "personal",
+})
+
+const stop = async () => {
+  await proxy.close()
+  process.exit(0)
+}
+
+process.on("SIGINT", () => { void stop() })
+process.on("SIGTERM", () => { void stop() })
+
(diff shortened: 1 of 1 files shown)


rynfar:
I'll review this in the morning!  Thank you for. the submission and ideas!

davetha:
Thanks for the feedback.  Whenever I get some free time I'll get the suggestions added in, and continue testing.

rynfar:
This is a genuinely useful feature and I spent some time redesigning this to make it more meaningful.  I've added support for multiple profiles.  There is also a UI to help you manage your profiles with clear instructions in how to authenticate each profile and how to switch bertween them easily - there is also a cli for this if you prefer the CLI.

Right now you can only have 1 profile active at a time, i may extend this to allow 2 profiles under different ports running at the same time depending on if this is a mode people want.

Right now you will be able  to create propfiles and simply swap between them at the proxy level, this will allow you to swap without losing resumability for a session you currentlyl have active.

Give it a spin it will be in the next release.  I am doing some testing and then i will release it.

@davetha

davetha:
I haven't had time, but I'll give it a whirl.

One thing I should mention is the refresh token wasn't working on the secondary Claude/non-default path.

Let me know if you run into the same issue.  It may have been a docker issue or something I had misconfigured.

rynfar:
> I haven't had time, but I'll give it a whirl.
> 
> One thing I should mention is the refresh token wasn't working on the secondary Claude/non-default path.
> 
> Let me know if you run into the same issue. It may have been a docker issue or something I had misconfigured.

Yea I noticed that while i was testing and am fixing this as part of it.  I should have a release for this later today once I've had enough time to stress test it a bit.

The UI will keep track of the last successful refresh for each profile and will refresh both profiles.

You still have to manually log out of oauth before adding a second account.  This part is janky and really hard to code around so its not really worth it.  For now you just have to manage the login yourself.  But once logged in the proxy will keep your profiles refreshed.  I put clear instructions in the UI so its easy to follow.

PR #163, opened 2026-03-26
docs: add same-port Claude Code profile launcher example

davetha:
## Summary
- add a small shell example for launching Claude Code against the same Meridian port while selecting `personal` or `company` via `ANTHROPIC_CUSTOM_HEADERS`
- keep the example intentionally simple so users can adapt it to local shell aliases or wrappers

## Use Case
I have both a personal and a company Claude account/plan, and I want to reach one Meridian instance from several systems over my local network or a private Tailscale tunnel instead of installing Meridian and repeating `claude auth` on every machine.

This example shows the last mile for that setup: keep one Meridian port, then launch Claude Code with the right profile header so the client feels separated in a similar way to having per-system `claude auth`, while still centralizing control.

## Dependency
Draft because this stacks on top of:
- #161

GitHub cannot express the exact stacked base across forks, so this draft PR includes earlier profile-routing work until #161 merges.

## Verification
- `bash -n examples/claude-code-profile.sh`

### DIFF
--- a/examples/claude-code-profile.sh
+++ b/examples/claude-code-profile.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+PROFILE="${1:-personal}"
+BASE_URL="${ANTHROPIC_BASE_URL:-http://127.0.0.1:4567}"
+
+case "$PROFILE" in
+  personal|company)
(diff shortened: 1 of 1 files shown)


davetha:
Superseded by #165

PR #164, opened 2026-03-26
feat: protect admin routes with optional basic auth

davetha:
## Summary
- add optional protection for `/health`, `/telemetry`, and `/telemetry/*`
- support a separate admin key set or fall back to the existing request keys
- add optional Basic Auth so browser access to telemetry/health can prompt instead of only returning JSON 401s

## Use Case
My goal is to run one Meridian instance for multiple machines and access it over my private network or Tailscale. In that setup, I still want tighter control over who can see `/health` and `/telemetry`, not just who can call `/v1/messages`.

Basic Auth helps for browser access, while admin API keys still work for scripts and tooling.

## Dependency
Draft because this currently stacks on top of:
- #161
- the config/request-auth branch

GitHub cannot express the exact stacked base across forks, so this draft PR includes earlier work until the lower PRs merge.

## Testing
- `bun test src/__tests__/proxy-admin-auth.test.ts src/__tests__/proxy-auth.test.ts src/__tests__/configLoader.test.ts`
- `npm run build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ Example:
 {
   "port": 4567,
   "defaultProfile": "personal",
+  "protectAdminRoutes": true,
   "requiredApiKeys": ["env:MERIDIAN_LAPTOP_KEY", "env:MERIDIAN_DESKTOP_KEY"],
   "profiles": [
     { "id": "personal", "claudeConfigDir": "~/.claude" },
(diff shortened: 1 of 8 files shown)


davetha:
Superseded by a new PR that includes a health check fix for multi-profile Docker setups.

PR #162, opened 2026-03-26
feat: add config-based request authentication

davetha:
## Summary
- add optional inbound request authentication for `/v1/messages` and `/messages` using configured API keys
- add JSON config loading from `~/.config/meridian/config.json`, `./meridian.config.json`, or `CLAUDE_PROXY_CONFIG`
- document local and Docker config patterns, including env-backed secrets and plaintext JSON keys

## Use Case
I have both a personal and a company Claude account/plan, and I want to use Meridian from several systems over my local network or a private Tailscale tunnel instead of installing Meridian and repeating `claude auth` everywhere.

This PR is part of that shared-instance setup: it adds a better config story and a way to restrict which clients can call the proxy.

## Dependency
Draft because this currently stacks on top of:
- #161

GitHub cannot express the exact stacked base across forks, so this draft PR includes the earlier profile-routing commits until #161 merges.

## Testing
- `bun test src/__tests__/proxy-auth.test.ts src/__tests__/configLoader.test.ts src/__tests__/proxy-async-ops.test.ts`
- `npm run build`

PR #165, opened 2026-03-27
feat: multi-profile routing with auth and admin protection

davetha:
## Summary
- Add profile-based routing so Meridian can distinguish between multiple Claude accounts/contexts
- Add config file loading with optional request API key authentication
- Add optional admin route protection (API keys + Basic Auth) for `/health` and `/telemetry`
- Fix health check to use the default profile's `CLAUDE_CONFIG_DIR` instead of the unmounted default path
- Add Docker two-profile example and Claude Code profile launcher docs

## Use Case
Run a single Meridian instance for multiple machines over a private network or Tailscale, routing requests to the right Claude profile from each client. Tighter control over who can see health/telemetry vs who can call `/v1/messages`.

## Health Check Fix
The `/health` endpoint was calling `getClaudeAuthStatusAsync()` with no env overrides, so it checked `~/.claude` instead of the configured profile's `CLAUDE_CONFIG_DIR`. In multi-profile Docker setups where only profile dirs are mounted, this always returned "degraded" even though requests worked fine.

## Test plan
- [ ] `npm test` — all tests pass
- [ ] `npm run build` — clean build
- [ ] Deploy with two-profile Docker config and verify `/health` returns healthy
- [ ] Verify profile routing via `x-meridian-profile` header

Supersedes #164, #163, #162. Builds on #161.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ FROM node:22-alpine
 
 RUN deluser --remove-home node 2>/dev/null; \
     adduser -D -u 1000 claude \
-    && mkdir -p /home/claude/.claude \
+    && mkdir -p /home/claude/.claude /home/claude/.config/meridian \
     && chown -R claude:claude /home/claude
 
 RUN npm install -g @anthropic-ai/claude-code \
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ Example:
 {
   "port": 4567,
   "defaultProfile": "personal",
+  "protectAdminRoutes": true,
   "requiredApiKeys": ["env:MERIDIAN_LAPTOP_KEY", "env:MERIDIAN_DESKTOP_KEY"],
   "profiles": [
     { "id": "personal", "claudeConfigDir": "~/.claude" },
@@ -196,6 +197,34 @@ Plaintext keys also work if you want a fully self-contained local config:
 ```
 
 That is supported, but safer practice is to keep secret values in env vars and reference them from JSON.
+
+To lock down telemetry and health endpoints separately from message traffic, you can turn on admin-route protection and optionally use a different key set:
+
+```json
+{
+  "protectAdminRoutes": true,
+  "requiredApiKeys": ["env:MERIDIAN_CLIENT_KEY"],
+  "adminApiKeys": ["env:MERIDIAN_ADMIN_KEY"]
+}
+```
+
+For browser-friendly access, you can also enable Basic Auth on protected admin routes:
+
+```json
+{
+  "protectAdminRoutes": true,
+  "adminUsername": "admin",
+  "adminPassword": "env:MERIDIAN_ADMIN_PASSWORD"
+}
+```
+
+When `protectAdminRoutes` is enabled:
+
+- `/health` requires an admin key
+- `/telemetry` and `/telemetry/*` require an admin key
+- Basic Auth also works when `adminUsername` and `adminPassword` are configured
+- `/` remains public
+- if `adminApiKeys` is omitted, Meridian falls back to `requiredApiKeys`
 
 String values support:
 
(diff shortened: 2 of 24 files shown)


rynfar:
Same here i'll review in the morning!

rynfar (changes requested):
## Review

Nice work on the overall design — clean module extraction, good test coverage structure, and fully backwards compatible. The profile routing and config loading are well thought out. A few things to address before this can merge:

### 🔴 Tests will fail

1. **WWW-Authenticate realm mismatch** — server sends `Basic realm="Admin"` but the test in `proxy-admin-auth.test.ts` expects `Basic realm="Meridian Admin"`. One needs to match the other.

2. **Landing page is not actually public** — the README says `/ remains public` when `protectAdminRoutes` is enabled, and the test `"keeps the landing page route public"` asserts `status === 200` with no credentials. But `server.ts` applies `app.use("/", validateAdminAccess)` when `protectAdminRoutes` is true, which will return 401 when API keys are configured. Either remove the middleware registration on `"/"` or change the test/docs.

The test plan checkboxes are all unchecked — please run `npm test` to verify.

### 🟡 Security

3. **API key comparison is not constant-time** — `auth.ts` uses `normalizedKeys.includes(providedApiKey)` which is a plain string comparison vulnerable to timing attacks. Same for the Basic Auth `===` comparison. For network-exposed auth, use `crypto.timingSafeEqual()` with a fixed-length buffer comparison.

4. **No rate limiting on auth failures** — might be fine for a private-network use case, but worth noting. Could be a follow-up.

### 🟡 Architecture

5. **`getProfileId()` doesn't belong on `AgentAdapter`** — the `x-meridian-profile` header is a proxy-level (Meridian infrastructure) concern, not an agent-specific one. Every adapter would implement it identically. This should be read directly in `server.ts` rather than forced through the adapter interface. The adapter pattern is for abstracting *agent differences* (OpenCode vs. future agents).

6. **`resolveProfile()` throws on unknown profile** — if a client sends `x-meridian-profile: nonexistent`, this throws into the generic catch handler and returns a 500 with a stack trace. Should be caught and mapped to a 400 with a clear error message.

### 🟠 Suggestions

7. **"requested" vs "effective" profile** — the distinction between these two isn't documented. If a session was started with profile A but the client now sends profile B, the cached profile silently wins. Should this warn or error on conflict?

8. **`storeSession()` is at 7 positional params now** — consider an options object to keep it readable.

9. **Silent `undefined` for missing env references** — `"env:MISSING_VAR"` silently resolves to `undefined`. For `apiKey`/`authToken`, this will cause a confusing runtime error later. Consider validating and warning on startup.

10. **Minor indentation issue** — the stale-session retry block in `server.ts` has misaligned indentation after the edit.

davetha:
Thanks for the feedback.  I'll work on it as time permits.

rynfar:
FYI - I haven't forgotten about this.  I just haven't had time to circle back to it.  I am going to try and get to this in the next couple of days.  @davetha

rynfar:
Hey @davetha — thanks for this PR and the work you put into it. The core multi-profile routing from this PR inspired what we shipped in v1.32.0, and profiles are now fully supported with CLI management, web UI switching, and env var configuration.

What we just added from your PR that was missing: **optional API key authentication** for network-exposed deployments. Setting `MERIDIAN_API_KEY=your-secret` now protects all API and admin routes while keeping `/health` open for monitoring. Clients just set their `ANTHROPIC_API_KEY` to the shared secret — no workflow changes needed.

What we're not adding:
- **Config file loading** — we use `MERIDIAN_*` env vars consistently, which works cleanly with Docker, launchd, and systemd. A config file adds complexity without solving problems the env vars don't.
- **Basic Auth on admin routes** — the API key middleware covers this. All telemetry and profile routes are protected when `MERIDIAN_API_KEY` is set.

I understand the multi-profile routing as it exists today may not fully cover your use case. If there are gaps — especially around the Docker/Tailscale multi-machine setup you described — let me know and we can look at how to expand its abilities.

Closing this PR since the profile routing and auth features are now shipped. Thanks again for pushing us in this direction — the auth feature in particular was a gap we needed to fill for network-exposed deployments.

Issue #189, opened 2026-03-29
Build command lacks file change visibility


endre82:
## Problem
When running build agent, there's no feedback showing which files are being modified, added, or deleted during the build process. Users only see generic completion messages without details of actual file operations.

## Current behavior
- Run build agent
- Build completes
- Terminal shows success in text only
- No information about files created, modified, or deleted

KevMac19:
I’m facing the same issue with OpenCode.

During usage, it breaks intermittently and most of the times returns raw response output directly in the thread, which disrupts the expected flow.

Also, the behavior becomes inconsistent during editing or tool execution, making it difficult to rely on the responses.

It indicates that the file was “edited successfully”, but the changes are not actually reflected in the codebase.

<img width="990" height="196" alt="Image" src="https://github.com/user-attachments/assets/e51400c2-9832-4d45-a337-ae092f5e0d04" />

rynfar:
Looking into this!

braydenbabbitt:
Not sure if related, but when I started experiencing this, I also noticed that when bash/terminal commands being run, they don't show up either

rynfar:
This is shipping in the next release (PR #192).

Responses now include a **Files changed** summary when the model writes or edits files:

```
I've fixed the bug in the add function.

Files changed:
- edited src/math.ts
- wrote src/__tests__/math.test.ts
```

### What's tracked

- **`write` / `edit` / `multiedit` tools** — always tracked
- **Bash redirects** — `echo X > file`, `>> file`, `cat > file << EOF`, `tee file`, `sed -i` are all detected
- **Read-only operations** — correctly excluded, no false positives

### A note on coverage

There are some cases where a file change happens but won't appear in the summary — for example, if the model uses `cp`, `mv`, or runs an arbitrary script that writes files. The file change still happens normally and your code is modified as expected. The only thing missing is the visibility in the summary.

This is a limitation of where Meridian sits in the stack. The proxy only sees API messages, not the filesystem. To get true 100% coverage, the client (OpenCode, Cline, etc.) would need to report which files it touched back to the proxy. That's something we'd welcome as a client-side feature if there's interest, but it's outside our control today.

In practice, agents use `write`/`edit` tools and `echo >` redirects for the vast majority of file operations, so the summary captures most real-world changes.

Works across all supported agents (OpenCode, Crush, Cline, Droid) in both streaming and non-streaming modes.

KevMac19:
[...]

PR #200, opened 2026-03-31
feat: add env var to disable file change summaries

davdroman:
## Summary

- Adds `MERIDIAN_NO_FILE_CHANGES` env var to disable the "Files changed" summary appended to responses
- Guards all three injection points: PostToolUse hook registration, non-streaming summary, and streaming SSE block emission

## Motivation

The file change summary introduced in #189 is too noisy for some setups and should be opt-in, or at the very least let users opt out. This adds the opt-out mechanism via env var.

## Usage

```bash
MERIDIAN_NO_FILE_CHANGES=1 npm start
```

### DIFF
--- a/src/proxy/server.ts
+++ b/src/proxy/server.ts
@@ -413,7 +413,8 @@ export function createProxyServer(config: Partial<ProxyConfig> = {}): ProxyServe
       // PostToolUse hook tracks file changes from MCP tools (internal mode only).
       // Catches write, edit, AND bash redirects (>, >>, tee, sed -i).
       const mcpPrefix = `mcp__${adapter.getMcpServerName()}__`
-      const fileChangeHook = createFileChangeHook(fileChanges, mcpPrefix)
+      const trackFileChanges = !process.env.MERIDIAN_NO_FILE_CHANGES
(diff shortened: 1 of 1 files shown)


rynfar:
thanks for this submission, i am going to adopt this with one minor change.

rynfar:
Thanks for this — the motivation and implementation were solid. We landed it in #209 with three small adjustments:

**`CLAUDE_PROXY_NO_FILE_CHANGES` fallback added**
Every other env var in the codebase uses `MERIDIAN_X ?? CLAUDE_PROXY_X` so users on the older prefix can still use it. Updated to:
```typescript
const trackFileChanges = !(process.env.MERIDIAN_NO_FILE_CHANGES ?? process.env.CLAUDE_PROXY_NO_FILE_CHANGES)
```

**4 regression tests added**
The project requires tests for all behaviour changes. Added to `proxy-file-changes.test.ts`: hook suppression, non-streaming summary suppressed, streaming SSE block suppressed, and the `CLAUDE_PROXY_` fallback.

**README env var table updated**
`MERIDIAN_NO_FILE_CHANGES | CLAUDE_PROXY_NO_FILE_CHANGES` added to the configuration table so users can discover it.

Closing in favour of #209, now merged to `main`.

Issue #199, opened 2026-03-31
Add LiteLLM passthrough adapter for proxy


endre82:
Allow Meridian proxy to detect and handle LiteLLM-style requests via x-litellm-* headers.
When LiteLLM is configured as a proxy provider, it sends x-litellm-* headers with requests. This change adds detection for these headers and routes such requests to a passthrough adapter that forwards them directly to Claude Code SDK without session management or tool blocking.
- Added `hasLiteLLMHeaders()` detection in `detect.ts`
- Created `passthroughAdapter` in `adapters/passthrough.ts`
- Added `DEBUG_PROXY` env var for cache hit/miss visibility in `cache.ts`
- Added unit tests for passthrough detection

endre82:
I know, this proxy is focusing on OpenCode, but I would like to extend the usage to LiteLLM.

I've a fix developed for that already, I would like to commit to a branch for review.

PR #201, opened 2026-03-31
Feat/litellm passthrough #199

endre82:
#199 

feat: add LiteLLM passthrough adapter with x-litellm-* header detection
- Add tsx as dev dependency and update supervisor to prefer bun > tsx > npx
- Detect LiteLLM by user-agent header (litellm/*) in addition to x-litellm-* headers
- Force stream=false for all LiteLLM requests (healthchecks don't send x-litellm-* headers)
- Increase MAX_CONCURRENT_SESSIONS default from 10 to 50
- Increase rate-limit retry attempts (2→3) and base delay (1s→2s) with exponential backoff
- Allow rate-limit retry even after partial content was yielded
- Add DEBUG_PROXY=true flag for detailed error diagnosis
- Add prefersStreaming() to AgentAdapter interface (unused but available)

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,33 @@ ANTHROPIC_API_KEY=x ANTHROPIC_BASE_URL=http://127.0.0.1:3456 opencode
 
 The API key value doesn't matter — Meridian authenticates through your Claude Max session, not API keys.
 
+## Install from Source
+
+```bash
+git clone https://github.com/rynfar/meridian.git
+cd meridian
+npm install
+```
+
+Requires: Node.js >=22. Choose your preferred runtime:
+
+**Option 1: Build + supervisor (Node.js only)**
+```bash
+npm run build && npm start
+```
+
+**Option 2: bun (native TypeScript execution)**
+```bash
+bun install
+bun run dev
+bun run start   # for auto-restart supervisor
+```
+
+**Option 3: tsx (no build step, Node.js only)**
+```bash
+npm run dev
+```
+
 ## Why Meridian?
 
 You're paying for Claude Max. It includes programmatic access through the Claude Code SDK. But your favorite coding tools expect an Anthropic API endpoint and an API key.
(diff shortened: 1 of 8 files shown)


rynfar:
Thanks for putting this together — the LiteLLM adapter concept is solid and we want to get it in. However the PR bundles several unrelated changes that need to be separated before we can merge anything.

**What we're pulling out and merging separately:**
The core LiteLLM adapter (`adapters/passthrough.ts`, detection in `detect.ts`, tests) is good and we'll get that in as a clean PR.

**What we're not merging from this PR, and why:**

**`MAX_CONCURRENT_SESSIONS` 10→50** — You can already control this yourself via `MERIDIAN_MAX_CONCURRENT` env var, so no code change is needed. That said, if you're hitting concurrency limits with LiteLLM specifically, bumping this is a reasonable thing to try — just be aware higher values risk process instability since each SDK spawn is ~11MB. We've noted this in the LiteLLM docs we're adding in the clean PR.

**Rate-limit retry after partial content** — The original guard ("Never retry after response content was yielded — response is committed") was intentional. Removing it for rate-limit errors risks corrupting or duplicating an in-flight SSE stream. The client has already received partial events — resuming on the same connection isn't safe. This needs more careful thought as a standalone change.

**`tsx` dev dependency** — Adds 500+ lines to the lock file for a `dev` convenience script. The project already has `bun run ./bin/cli.ts` for development. Not needed.

**`DEBUG_PROXY` env var** — The project already has `CLAUDE_PROXY_DEBUG` routed through `claudeLog`. Adding a second debug mechanism that writes raw `console.error` directly is inconsistent, and one of the debug lines isn't guarded by the flag at all (it fires on every rate limit for every user). We'll incorporate the useful cache visibility into the existing debug mechanism instead.

**`prefersStreaming()` on the adapter interface** — Added but never called in `server.ts`. The stream override is done by duplicating the LiteLLM header detection inline in `server.ts` instead. We'll wire it up properly so the adapter controls its own streaming preference.

We'll post the clean PR shortly and reference this one.

rynfar:
Closing this out now that the LiteLLM adapter has landed in #215 (shipped in v1.24.0). The core work here — adapter detection, passthrough mode, session continuity via `x-litellm-session-id` — is all in. Thanks again @endre82, the original PR was a solid foundation.

PR #206, opened 2026-04-01
fix: passthrough mode tool_use broken for multi-turn and streaming

KentLee86:
## Summary

Fixes two bugs in passthrough mode (`MERIDIAN_PASSTHROUGH=true`) that break tool_use flows:

### Bug 1: `maxTurns=1` causes HTTP 500 on multi-turn tool flows

When a client sends tools and the model responds with `tool_use`, the SDK needs at least one more turn to process the tool result. With `maxTurns=1`, the SDK immediately errors with "Reached maximum number of turns (1)".

**Fix**: Changed `maxTurns` from `1` to `10` in passthrough mode to allow multi-step tool flows.

### Bug 2: Streaming path continues after `tool_use` stop, producing incorrect responses

In streaming mode, when the model emits `stop_reason: tool_use`, the SDK continues internally — it executes the passthrough MCP tools (which return placeholder `"passthrough"` text), feeds that back to the model, and the model generates a fallback text response thinking the tools failed. The client never gets a chance to execute the tools.

**Fix**: Break the stream loop immediately when `message_delta` with `stop_reason: tool_use` is detected in passthrough mode, emitting `message_stop` and closing the controller so the client can execute tools and send results back.

## Changes

- `src/proxy/query.ts`: `maxTurns: passthrough ? 1 : 200` → `maxTurns: passthrough ? 10 : 200`
- `src/proxy/server.ts`: Add early stream termination on `tool_use` stop in passthrough mode

## Testing

Tested with a multi-step tool_use flow (weather → recommendation) in both non-streaming and streaming modes:
- Non-streaming: model returns `tool_use` blocks, client sends `tool_result`, model responds correctly
- Streaming: SSE stream ends cleanly at `tool_use` stop, client processes tools, sends results, gets final response

### DIFF
--- a/src/proxy/query.ts
+++ b/src/proxy/query.ts
@@ -61,7 +61,7 @@ export function buildQueryOptions(ctx: QueryContext) {
   return {
     prompt,
     options: {
-      maxTurns: passthrough ? 1 : 200,
+      maxTurns: passthrough ? 10 : 200,
       cwd: workingDirectory,
       model,
       pathToClaudeCodeExecutable: claudeExecutable,
--- a/src/proxy/server.ts
+++ b/src/proxy/server.ts
@@ -965,6 +965,17 @@ export function createProxyServer(config: Partial<ProxyConfig> = {}): ProxyServe
                     }
                     eventsForwarded += 1
 
+                    // Passthrough mode: when the model stops for tool_use, end the stream
+                    // immediately so the client can execute the tools and send results back.
+                    // Without this, the SDK continues processing "passthrough" tool results
+                    // internally, wasting tokens and producing incorrect responses.
+                    if (passthrough && eventType === "message_delta" && (event as any).delta?.stop_reason === "tool_use" && streamedToolUseIds.size > 0) {
+                      safeEnqueue(encoder.encode(`event: message_stop\ndata: ${JSON.stringify({type: "message_stop"})}\n\n`), "passthrough_tool_stream_stop")
+                      streamClosed = true
+                      try { controller.close() } catch {}
+                      break
+                    }
+
                     if (eventType === "content_block_delta") {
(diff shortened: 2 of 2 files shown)


rynfar:
Thank you for this contribution.  I am going to adopt this change, i have a couple of adjustments but i will get this fix in.  Thank you for catching this.

rynfar:
Thanks for the diagnosis here — both bugs were real and the fixes landed in #207. Before merging we made a few adjustments to align with project conventions:

**`maxTurns: 10` → `maxTurns: 2`**
The SDK only needs two turns in passthrough mode: one for the model to generate the `tool_use` blocks, and one for it to process the blocked-tool handoff from the PreToolUse hook. `10` was more generous than necessary and could allow unexpected multi-turn behaviour. `2` is the minimal safe value, and we added a comment explaining why.

**Empty catch block removed**
`try { controller.close() } catch {}` violates the project's style rules (CLAUDE.md: "No empty catch blocks"). Since `streamClosed` is set to `true` immediately before calling `controller.close()`, and no other code path can close the controller at that point, the call won't throw — so the try/catch wasn't needed.

**Existing test updated**
`query.test.ts` had `expect(result.options.maxTurns).toBe(1)` which would have failed with the change to `2`. That assertion and the related comments in `proxy-passthrough-concept.test.ts` and `proxy-transparent-tools.test.ts` were updated.

**Regression tests added**
Five new tests in `proxy-passthrough-tool-use.test.ts` covering both bugs: streaming early termination (single tool, multi-tool, and end_turn negative case) and non-streaming 200 response for both `tool_use` and `end_turn` stops.

**E2E verified**
E17 (non-streaming and streaming passthrough), E1, and E2 were run against a live Claude Max subscription and all passed.

Closing in favour of #207 which is now merged to `main`.

PR #219, opened 2026-04-01
fix: force executable to node in buildQueryOptions

ianjwhite99:
## Problem

The claude-agent-sdk auto-detects Bun via `process.versions.bun` and defaults to spawning `bun cli.js`. Hosts like OpenCode embed Bun in their native binary, so `process.versions.bun` is set even when `bun` is not available in the system PATH. This causes the SDK's `child_process.spawn("bun", ...)` to fail with ENOENT.

## Fix

Add `executable: "node"` to the options returned by `buildQueryOptions()`. This tells the SDK to always use Node to run the CLI subprocess, bypassing the auto-detection.

## Who this affects

Users running meridian under environments that embed Bun (e.g. OpenCode) but don't have `bun` installed system-wide. Common in Docker containers and CI environments.

### DIFF
--- a/src/proxy/query.ts
+++ b/src/proxy/query.ts
@@ -63,6 +63,11 @@ export function buildQueryOptions(ctx: QueryContext) {
   return {
     prompt,
     options: {
+      // Force Node as the executable. The claude-agent-sdk auto-detects Bun
+      // via process.versions.bun and defaults to spawning `bun cli.js`.
+      // Hosts like OpenCode embed Bun, so the check fires even when `bun`
+      // is not in PATH — causing subprocess spawns to fail.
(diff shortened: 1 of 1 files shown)


rynfar:
sorry for the delay had a huge backlog, looking at this one today

ianjwhite99:
All good. See here for more background as well as documentation: https://github.com/ianjwhite99/opencode-with-claude/issues/50

rynfar:
Applied directly to main as commits 6e33926 (fix) + 8a5cc9e (test) — added a missing test asserting `executable === 'node'` per project conventions, then cherry-picked both commits onto main. Thanks @ianjwhite99!

PR #220, opened 2026-04-01
feat: add agent mode support - skip [1m] models for subagents

lnilluv:
## Summary
Adds agent mode support so subagents can skip [1m] models to save rate limit budget.

## Changes
- Read x-opencode-agent-mode and x-opencode-agent-name headers in server.ts
- Updated mapModelToClaudeModel to accept agentMode parameter - subagents get base models
- Added isPremiumSubscription() for subscription-aware model selection  
- Added circuit-breaker for extended context rate limits (recordExtendedContextRateLimit)
- Added example OpenCode plugin meridian-agent-mode.ts

### DIFF
--- a/examples/opencode-plugin/meridian-agent-mode.ts
+++ b/examples/opencode-plugin/meridian-agent-mode.ts
@@ -1,0 +1,63 @@
+/**
+ * OpenCode plugin — agent mode headers for Meridian.
+ *
+ * Sends x-opencode-agent-name and x-opencode-agent-mode headers with every
+ * Anthropic API request so Meridian can make context-aware model decisions:
+ *
+ *   - Primary agents get opus[1m] / sonnet[1m] (full 1M context window)
+ *   - Subagents get opus / sonnet (200k context, saves rate limit budget)
+ *
+ * This plugin works alongside claude-max-headers.ts. Both can be loaded
+ * simultaneously — they write different headers.
+ *
+ * Usage:
+ *   Copy this file into your project and add to opencode.json:
+ *     { "plugin": ["./meridian-agent-mode.ts"] }
+ *
+ *   Or combine with the session headers plugin:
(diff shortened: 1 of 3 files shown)


bnema:
Heads up: the `chat.headers` hook in the plugin sets the headers correctly on the AI SDK layer, but I found that `client.agent.list()` in the plugin init throws `TypeError: client.agent.list is not a function`. The SDK client uses `client.app.agents()` instead, but that also fails at plugin load time because the internal server isn't ready yet.

I ended up parsing the agent `.md` frontmatter directly from `~/.config/opencode/agents/` as a workaround. Not great but it works until opencode adds `parentAgent` to the `chat.headers` hook input (anomalyco/opencode#15412).

I confirmed the custom headers do reach the proxy with the Anthropic provider, but it took some debugging to figure out the agent list API issue.

rynfar:
I found a solution for this.  ill have it deployed by morning

rynfar:
Thanks for this PR — the approach here is exactly right and directly inspired the implementation we landed in #235.

#235 covers everything in this PR and extends it:

- **Subagent model selection** — `x-opencode-agent-mode: subagent` header forces base model (same as here)
- **Combined plugin** — `plugin/meridian.ts` ships with the package, merging session tracking + agent mode into one file
- **`meridian setup`** — one command configures the plugin in `~/.config/opencode/opencode.json`, handles stale paths from previous installs, cross-platform
- **Extra Usage cooldown** — instead of a rate-limit circuit breaker, we track when Extra Usage is confirmed unavailable and skip `[1m]` for 1 hour, then probe once (self-healing if the user enables Extra Usage later)
- **Startup warning** — proxy warns at boot if the plugin is missing
- **38 new tests** — unit + integration coverage for all of the above

Closing in favour of #235. Really appreciate the contribution — this unblocked the whole feature.

Issue #223, opened 2026-04-02
Manual token-auto-refresh


endre82:
Since this proxy runs as a service on an unattended workstation, the authentication token can expire. Therefore, a CLI command is helpful to update the authentication token via cron job or systemd.

I will prepare a fix and a merge request. (This time I'm focusing on the problem without making any additional changes.)

Note: There is an implemented automatic token update, but it doesn't work reliably.

rynfar:
Fixed in #230 — now merged. Three ways the token stays healthy:

1. **Automatic** — when any request hits an expired token, Meridian refreshes it inline and retries transparently. No user action needed.
2. **CLI** — `meridian refresh-token` (exits 0/1, works in scripts)
3. **HTTP** — `curl -X POST http://127.0.0.1:3456/auth/refresh` (while the proxy is running)

Works on macOS (Keychain) and Linux (credentials file).

PR #224, opened 2026-04-02
Feat/token refresh

endre82:
#223

- Simplify tokenRefresh.ts to expose only refreshOAuthToken()
- Remove auto-refresh from request handler (scheduling is ops concern)
- Add 'refresh-token' CLI subcommand for manual/scheduled token refresh
- Exit code: 0 (success), 1 (failure) for integration with cron/systemd
- Update tests: keep OAuth logic, remove scheduling tests
- Document in README with cron and systemd timer examples
- Users can now: meridian refresh-token OR cron schedule it externally

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,32 @@ It works with any Claude subscription that supports the Claude Code SDK. Max is 
 **What happens if my session expires?**
 The SDK handles token refresh automatically. If it can't refresh, Meridian returns a clear error telling you to run `claude login`.
 
+## Token Refresh
+
+The OAuth token expires periodically. While the SDK subprocess refreshes tokens automatically when needed, you can also refresh manually on a schedule:
+
+```bash
+# Manual refresh
+meridian refresh-token
+
+# Schedule with cron (every 6 hours)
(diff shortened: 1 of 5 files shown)


rynfar:
thanks for this submission i will review this

rynfar:
@endre82  - do we really need the cron job if we fix the underlying issue?   If inline refresh is working correctly a cron job doesn't offer any value.  I am fixing the underlying issue.

I am also making it cross platform so it works on mac, windows and linux.

endre82:
@rynfar, yes, I would like to avoid workarrounds. I'm building dockerized environments, for develoment and also some autonomous task worker. when they don't get anything to do over the weekend, the authentication process needs to be repeated. Unfortunately this will run on an unattended workstation, which makes it harder. A CLI refresh command is a very clear interface.

I would be happy to have those lines of code. to set up a cronjob should be anyway setup task. I'm not sure if I could clearly state the value of this little feature.

rynfar:
i've deployed a fix for this please pull latest you'll see instructions on the readme

Issue #225, opened 2026-04-02
Add Request Validation & E2E Test Coverage


endre82:
# Add Request Validation & E2E Test Coverage
## Summary
Improve proxy reliability by validating required request fields and establish end-to-end test coverage against the real Claude API.
## What's included
- Request validation: Catch missing/invalid `messages` field early with proper error responses
- Live test suite: 5 smoke tests verifying health, response shapes, streaming, sessions, and error handling
- Cost-aware testing: Opt-in live tests with low token usage to avoid quota drain
## Outcome
- Clearer error messages for invalid requests
- Confidence in proxy correctness with real Claude API
- Prevention of malformed requests reaching the SDK
## Testing
All tests pass (633 unit tests + 7 live smoke tests). Live tests require Claude Max subscription and are opt-in via `LIVE=true` flag.

rynfar:
We have an e2e test suite, it's more dynamic and will work with other models.  It's an agent driven approach that has proven very reliable for helping me debug difficult issues.  I would suggest you try this and if you want to add tests to it please do.

endre82:
Sure thanks! what about the message validation part (additional 7 lines in server.ts)?

rynfar:
yes absolutley I will make sure this is captured

rynfar:
Hey @endre82 — both pieces from this issue are now in place:

**Request validation** — `messages` field validation landed a while back (server.ts line 207). Missing/invalid `messages` returns a proper Anthropic-format error:
```json
{"type": "error", "error": {"type": "invalid_request_error", "message": "messages: Field required"}}
```

**E2E tests** — We went with an agent-driven approach documented in `E2E.md` that covers health, streaming, sessions, error handling, and now profile lifecycle (P1–P10). It's model-agnostic and runs against the live proxy with curl.

Thanks for pushing on both of these — the validation suggestion was spot on and the E2E conversation helped us formalize our testing approach. 🙏

PR #226, opened 2026-04-02
feat: add live smoke tests + message validation

endre82:
#225


rynfar:
thanks for this one also!  I'll get this reviewed

rynfar:
Thank you for contributing this, i made some c hanges to your branch and am going to merge it.  I already have a solid e2e workflow that has worked relaly well for me, its a bit different of an approach relying on the agents to do them but i found it has caught a lot of issues and they are way less fragile letting the agent run them itself.  it can adapt and update them as needed.

PR #234, opened 2026-04-03
feat: OpenAI-compatible /v1/chat/completions and /v1/models endpoints

luis15pt:
## Summary

- Adds `POST /v1/chat/completions` — translates OpenAI chat format to Anthropic `/v1/messages` internally, with full streaming (SSE) and non-streaming support
- Adds `GET /v1/models` — returns available Claude models in OpenAI-compatible format, with context window based on subscription type
- Enables clients like **Open WebUI** to connect directly to Meridian without needing LiteLLM or other translation proxies

## Motivation

Many popular LLM frontends (Open WebUI, Continue, etc.) only speak the OpenAI protocol. Currently, using Meridian with these tools requires an intermediate proxy like LiteLLM to translate between OpenAI and Anthropic formats. This adds complexity, an extra container to manage, and introduces streaming translation bugs.

These two endpoints make Meridian a drop-in replacement that works with any OpenAI-compatible client out of the box.

## Implementation Details

**`/v1/chat/completions`**
- Extracts system prompts from the OpenAI messages array into Anthropic's top-level `system` field
- For multi-turn conversations, packs prior messages into system context and sends only the last user message — this avoids conflicts with Meridian's session resumption (OpenAI clients send full history with every request, they don't use session continuations)
- Translates Anthropic SSE events (`content_block_delta`, `message_delta`, `message_start`) to OpenAI SSE chunks (`chat.completion.chunk`)
- Forwards `temperature`, `top_p`, `max_tokens`/`max_completion_tokens`

**`/v1/models`**
- Returns `claude-sonnet-4-6`, `claude-opus-4-6`, `claude-haiku-4-5-20251001`
- Sets `context_window` to 1M for Max subscriptions, 200K otherwise
- Response format matches what clients expect from `/v1/models`

## Test plan

- [x] Non-streaming single message
- [x] Streaming single message
- [x] Multi-turn conversation (3+ turns, streaming)
- [x] System prompt handling
- [x] Multiple system messages (concatenated)
- [x] Different models (Sonnet, Opus, Haiku)
- [x] Temperature and top_p passthrough
- [x] Minimal payload (no model, no max_tokens — defaults applied)
- [x] Empty messages array (returns error correctly)
- [x] Unknown model names (graceful fallback to Sonnet)
- [x] `/v1/models` returns correct model list
- [x] Open WebUI integration — full end-to-end chat with streaming

### DIFF
--- a/E2E.md
+++ b/E2E.md
@@ -78,6 +78,9 @@ kill $(lsof -ti :3456)
 | E22 | [OAuth Token Refresh](#e22-oauth-token-refresh) | Expired access token auto-refreshed inline; request succeeds without manual `claude login` | 2026-04-02 |
 | E23 | [Subagent Model Selection](#e23-subagent-model-selection) | `x-opencode-agent-mode: subagent` header selects base model; primary gets 1M; proxy log shows `agent=subagent` | 2026-04-02 |
 | E24 | [Default Non-Streaming](#e24-default-non-streaming) | Omitting `stream` field returns JSON (not SSE), matching Anthropic API spec | - |
+| E25 | [OpenAI Compat: Non-Streaming](#e25-openai-compat-non-streaming) | `/v1/chat/completions` returns valid OpenAI completion shape | - |
+| E26 | [OpenAI Compat: Streaming](#e26-openai-compat-streaming) | `/v1/chat/completions` with `stream: true` returns OpenAI SSE chunks | - |
+| E27 | [OpenAI Compat: Models](#e27-openai-compat-models) | `GET /v1/models` returns Claude model list in OpenAI format | - |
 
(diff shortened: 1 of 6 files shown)


rynfar:
Thanks for this!  I'll review in the morning!

rynfar:
I am going to look at adopting this but im going to need to make some changes to your branch before I can accept it.  Its an excellent base but its lacking somethings.  Rather than push back and make you fix it all im going to make the changes and push them back to your branch.

rynfar:
this is going to need some heavier testing but im going to merge my updated version.  Thank you for the contribution!

luis15pt:
Thanks!

PR #241, opened 2026-04-03
fix: default to non-streaming (JSON) when stream field is omitted

### LINKED ISSUE #239: Non-streaming /v1/messages requests return SSE instead of JSON
## Description

When a client sends a non-streaming request to `/v1/messages` (i.e. without `"stream": true`), Meridian responds with `text/event-stream` (SSE) instead of a single JSON object.

The Anthropic API specification requires that non-streaming requests return a single `application/json` response containing the full `Message` object, while only streaming requests (`"stream": true`) should return SSE.

## Reproduction

```bash
curl -s http://127.0.0.1:3456/v1/messages \
  -H "Content-Type: application/json" \
  -H "x-api-key: X" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-opus-4-5-20250514","max_tokens":100,"messages":[{"role":"user","content":"Hi"}]}'
```

**Expected:** A single JSON response:
```json
{
  "id": "msg_...",
  "type": "message",
  "role": "assistant",
  "content": [{"type": "text", "text": "..."}],
  "stop_reason": "end_turn",
  "usage": {...}
}
```

**Actual:** SSE event stream:
```
event: message_start
data: {"type":"message_start","message":{...}}

event: content_block_start
data: ...
```

## Impact

The Anthropic Python SDK's `client.messages.create()` method expects a JSON response for non-streaming calls. When it receives SSE instead, the SDK returns a raw string rather than a `Message` object, causing downstream `AttributeError: 'str' object has no attribute 'content'`.

This forces all consumers to use the streaming API (`client.messages.stream()`) as a workaround, even when streaming is not needed.

## Environment

- Meridian v1.24.5
- Anthropic Python SDK (via nanobot-ai v0.1.4.post6)
- Node.js v22.22.2

rafaelsales:
## Summary

- Changes the default streaming behavior from `true` to `false` when the `stream` field is omitted from `/v1/messages` requests, matching the Anthropic API spec
- Updates the `prefersStreaming()` JSDoc to reflect the new default
- Adds tests verifying that omitting `stream` returns JSON and that `stream: true` returns SSE

## Context

The Anthropic API specification requires that requests without `"stream": true` return a single `application/json` response. Meridian was defaulting to `true` (SSE) when the `stream` field was absent, which broke SDK clients using `messages.create()` — the Anthropic Python SDK returned a raw string instead of a `Message` object.

The one-line fix: `body.stream ?? true` → `body.stream ?? false` in `src/proxy/server.ts`.

Fixes #239

## Test plan

- [x] Existing streaming tests pass (explicitly set `stream: true`)
- [x] Existing non-streaming tests pass (explicitly set `stream: false`)
- [x] New test: omitting `stream` field returns `application/json`
- [x] New test: explicit `stream: true` returns `text/event-stream`

### DIFF
--- a/E2E.md
+++ b/E2E.md
@@ -77,6 +77,7 @@ kill $(lsof -ti :3456)
 | FC6 | [File Changes: Multiple ops (stream)](#fc6-file-changes-multiple-ops-stream) | Multiple file changes emitted as a text block in SSE stream | 2026-03-30 |
 | E22 | [OAuth Token Refresh](#e22-oauth-token-refresh) | Expired access token auto-refreshed inline; request succeeds without manual `claude login` | 2026-04-02 |
 | E23 | [Subagent Model Selection](#e23-subagent-model-selection) | `x-opencode-agent-mode: subagent` header selects base model; primary gets 1M; proxy log shows `agent=subagent` | 2026-04-02 |
+| E24 | [Default Non-Streaming](#e24-default-non-streaming) | Omitting `stream` field returns JSON (not SSE), matching Anthropic API spec | - |
 
 ---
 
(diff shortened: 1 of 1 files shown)


rynfar:
excellent find thank you for this! im adding a quick e2e test to this to verify and then i will merge it.

PR #261, opened 2026-04-04
fix: use envBool() for passthrough env var detection

beshkenadze:
## Summary
Replace manual `Boolean(process.env.MERIDIAN_PASSTHROUGH ?? process.env.CLAUDE_PROXY_PASSTHROUGH)` checks in `server.ts` with the existing `envBool("PASSTHROUGH")` helper from `src/env.ts`.

`envBool()` uses a whitelist approach — only `"1"`, `"true"`, `"yes"` enable passthrough. Everything else (including `"0"`, `"false"`, `""`, unset) disables it.

## Problem
- `Boolean("0")` is `true` in JavaScript, so `MERIDIAN_PASSTHROUGH=0` does **not** disable passthrough
- Some orchestrators (Portainer GitOps, etc.) strip empty-string env vars, so `MERIDIAN_PASSTHROUGH=""` may not override the Dockerfile default `CLAUDE_PROXY_PASSTHROUGH=1`
- The `envBool()` helper already exists in `src/env.ts` but wasn't being used for passthrough detection

## Changes
- Import `envBool` from `../env` in `server.ts`
- Replace 5 manual `Boolean()` / ternary passthrough checks with `envBool("PASSTHROUGH")`
- No Dockerfile changes — existing default preserved for backward compatibility

## Test plan
- [ ] `MERIDIAN_PASSTHROUGH=1` → `/health` returns `"mode":"passthrough"`
- [ ] `MERIDIAN_PASSTHROUGH=true` → `"mode":"passthrough"`
- [ ] `MERIDIAN_PASSTHROUGH=0` → `"mode":"internal"` (previously broken: was `"passthrough"`)
- [ ] `MERIDIAN_PASSTHROUGH=false` → `"mode":"internal"`
- [ ] No env var set → falls back to `CLAUDE_PROXY_PASSTHROUGH` from Dockerfile → `"mode":"passthrough"`

### DIFF
--- a/src/proxy/server.ts
+++ b/src/proxy/server.ts
@@ -5,6 +5,7 @@ import type { Server } from "node:http"
 import { query } from "@anthropic-ai/claude-agent-sdk"
 import type { Context } from "hono"
 import { DEFAULT_PROXY_CONFIG } from "./types"
+import { envBool } from "../env"
 import type { ProxyConfig, ProxyInstance, ProxyServer } from "./types"
 export type { ProxyConfig, ProxyInstance, ProxyServer }
 import { claudeLog } from "../logger"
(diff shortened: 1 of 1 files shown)


rynfar:
Thanks @beshkenadze! Great catch on the `Boolean('0')` footgun — that's a real bug affecting Docker users. We've taken your change, rebased onto main, and added unit tests (including the `"0"` case that was the root cause) plus headless validation. Merging via #267. Appreciate the contribution! 🙏

Issue #277, opened 2026-04-05
Extra usage error returned via OpenClaw despite Meridian working via curl


rparme:
### Environment
- Meridian: v1.27.1 (installed via npm)
- OpenClaw: 2026.3.28 (f9b1079)
- Claude Code CLI: v2.1.92
- Auth: Claude Max subscription 
- OS: Ubuntu on Raspberry Pi
- Meridian started with: `MERIDIAN_DEFAULT_AGENT=pi meridian`

### Summary

Meridian worked correctly yesterday after the v1.27.1 release. This morning, requests from OpenClaw are again returning the extra usage error from Anthropic. However, **curl requests to the same Meridian endpoint continue to work perfectly** — even with heavy payloads (15 tools, 10 messages, streaming, thinking enabled, 18k cached tokens).

### What works

Direct curl to Meridian — succeeds every time, including heavy payloads:

```
curl http://127.0.0.1:3456/v1/messages \
  -H "Content-Type: application/json" \
  -H "x-api-key: x" \
  -d '{
    "model": "claude-sonnet-4-6",
    "max_tokens": 100,
    "messages": [{"role": "user", "content": "You are an assistant for OpenClaw, say hi"}]
  }'
{"id":"msg_1775400811298","type":"message","role":"assistant","content":[{"type":"text","text":"Hi there! 👋 Welcome to **OpenClaw**! I'm your AI assistant, here to help you with anything you need. What can I do for you today?"}],"model":"claude-sonnet-4-6","stop_reason":"end_turn","usage":{"input_tokens":0,"output_tokens":0}}% 
```

Meridian health check confirms healthy auth:
```json
{"status":"healthy","auth":{"loggedIn":true,"email":"xxxxxxx","subscriptionType":"max"},"mode":"internal","plugin":{"opencode":"configured"}}%
```

### What fails

OpenClaw → Meridian → Claude Code SDK returns:

```
LLM error api_error: Claude Code returned an error result: API Error: 400
{"type":"error","error":{"type":"invalid_request_error","message":"You're out of extra usage. Add more at claude.ai/settings/usage and keep going."}}
Subprocess stderr: Warning: Custom betas are only available for API key users. Ignoring provided betas.
```

### Meridian logs

When OpenClaw sends a request:
```
[...]

PR #281, opened 2026-04-05
fix: strip anthropic-beta headers for Max subscriptions to prevent extra usage billing

### LINKED ISSUE #278: Strip anthropic-beta headers from incoming requests to prevent extra usage billing
## Problem

When OpenClaw 2026.4.2+ routes requests through Meridian, Anthropic detects them and bills as "extra usage" instead of subscription. Direct curl to Meridian works fine — the issue is specifically with OpenClaw → Meridian.

**Root cause:** OpenClaw 2026.4.2 introduced "native-vs-proxy endpoint classification" ([changelog](https://github.com/openclaw/openclaw/releases/tag/v2026.4.2)). When `baseUrl` points to a non-Anthropic host (like `127.0.0.1:3456`), OpenClaw now adds `anthropic-beta` headers that it previously didn't send. Meridian forwards these headers to Claude Code SDK, which triggers:

1. Claude CLI stderr warning: `"Custom betas are only available for API key users. Ignoring provided betas."`
2. Anthropic server-side detection → request billed as extra usage → `400 You're out of extra usage`

**On OpenClaw 2026.3.25** (before the routing change), these headers were not sent to proxy endpoints, so Meridian worked correctly.

## Environment

- Meridian: v1.27.2
- OpenClaw: 2026.3.25 (works) vs 2026.4.4 (fails)
- Claude Code CLI: v2.1.92
- Auth: Claude Max subscription
- OS: macOS (Mac Mini M4)

## Steps to reproduce

1. Configure OpenClaw with Meridian as Anthropic provider (`baseUrl: http://127.0.0.1:3456`)
2. On OpenClaw 2026.3.25 — works
3. Update OpenClaw to 2026.4.2+ — fails with extra usage error
4. Direct curl to Meridian — still works

## Proposed fix

Meridian should strip `anthropic-beta` and `x-anthropic-beta` headers from incoming requests before forwarding to Claude Code SDK. These headers are only valid for API key users and cause Claude SDK to switch to API-key billing mode.

Something like:

```typescript
// In request handler, before forwarding to SDK
const headersToStrip = ['anthropic-beta', 'x-anthropic-beta'];
headersToStrip.forEach(h => delete incomingHeaders[h]);
```

This would make Meridian resilient to upstream clients (OpenClaw, OpenCode, etc.) adding these headers in future versions.

## Workaround

Currently using OpenClaw 2026.3.25 (pinned) to avoid the issue. An nginx reverse proxy stripping the headers also works but adds unnecessary complexity.

Related: #255, #277

rynfar:
## Problem

When upstream clients like OpenClaw 2026.4.2+ route requests through Meridian, they add `anthropic-beta` headers. Meridian forwards these to the Claude Code SDK, which triggers API-key billing mode on Anthropic servers — causing Max subscribers to be charged **extra usage** instead of their subscription.

Closes #278

## Root Cause

OpenClaw 2026.4.2 introduced endpoint classification that adds `anthropic-beta` headers when `baseUrl` points to a non-Anthropic host (like `127.0.0.1:3456`). The Claude SDK warns: *"Custom betas are only available for API key users"* but still forwards them, triggering extra usage billing.

## Fix

- **claude-max profiles** (default): `anthropic-beta` headers are stripped before reaching the SDK, with a log message
- **api-type profiles**: Headers are forwarded normally (betas are valid for API key users)

The check uses the already-resolved `profile.type` so there is zero overhead.

## Changes

- `src/proxy/server.ts` — Conditionally read `anthropic-beta` header based on profile type
- `src/__tests__/proxy-sdk-params.test.ts` — Updated 2 existing tests + added 1 new test for api-type profile forwarding
- `src/__tests__/profile-switch-integration.test.ts` — Fixed `storeSession` call signatures (from multi-profile merge)

## Testing

- 888 tests total, 885 pass, 3 pre-existing failures (unrelated)
- New test verifies api-type profiles still forward betas correctly

### DIFF
--- a/src/__tests__/profile-switch-integration.test.ts
+++ b/src/__tests__/profile-switch-integration.test.ts
@@ -156,23 +156,15 @@ describe("Session cache eviction on profile switch", () => {
     const app = createTestApp(profiles)
 
     // Store a session in the cache
-    storeSession("test-session-123", {
-      claudeSessionId: "claude-abc",
-      lastAccess: Date.now(),
-      messageCount: 5,
-      lineageHash: "hash123",
-      messageHashes: ["h1", "h2", "h3", "h4", "h5"],
-      sdkMessageUuids: ["u1", "u2", "u3", "u4", "u5"],
-    })
-
-    // Verify session exists
-    const before = lookupSession("test-session-123", [
+    const msgs = [
       { role: "user", content: "a" },
       { role: "assistant", content: "b" },
       { role: "user", content: "c" },
-      { role: "assistant", content: "d" },
-      { role: "user", content: "e" },
-    ])
+    ]
+    storeSession("test-session-123", msgs, "claude-abc")
+
+    // Verify session exists
+    const before = lookupSession("test-session-123", msgs)
     expect(before.type).not.toBe("new")
 
     // Switch profile via API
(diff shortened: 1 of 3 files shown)


rynfar:
This also fixes #277 — same root cause (OpenClaw injecting anthropic-beta headers that trigger extra usage billing).

PR #327, opened 2026-04-09
fix: strip orchestration wrapper tags before text prompt flattening

### LINKED ISSUE #167: Transcript wrapper leakage still appears on v1.18.0 (system-reminder / thinking / tool_exec / tool_output / H:A wrappers)
## Summary

There is still a transcript-leakage class on current upstream `main` / `v1.18.0` that is distinct from the earlier raw `tool_use` text and block-index issues.

The proxy can still surface internal orchestration / transcript wrapper text into model-visible content, for example:

- `<system-reminder> ... </system-reminder>`
- `<task_metadata> ... </task_metadata>`
- `<thinking> ... </thinking>`
- leaked `<tool_exec ...>` / `<tool_output ...>` wrappers
- `<!-- OMO_INTERNAL_INITIATOR -->`
- `H:` / `A:` transcript prefixes
- `⚙ background_output [task_id=...]`

When these leak into prompt reconstruction or forwarded text, Claude can start echoing them back as normal conversation text.

## Example symptom

Things like this show up in assistant-visible output instead of being treated as internal transcript/control formatting:

```
Thinking: The user wants me to handle the case...
Tool Use: read({...})
H: [Tool Result for toolu_...: ...]
A: <tool_exec name="bash" />
<system-reminder>...</system-reminder>
```

This is very similar in user impact to #94 and #106, but the currently leaked shapes are broader than raw `[Tool Use: ...]` text alone.

## What I checked

- `v1.18.0` upstream `server.ts` still has no transcript sanitizer for these wrapper forms
- Earlier issues/PRs I checked:
  - #94 raw tool_use leakage
  - #106 talking-to-itself / system prompt leakage
  - #153 missing tools usage/output
  - #159 block index remapping

Those fixes help adjacent problems, but they do not seem to strip these specific wrapper patterns before they become model/user-visible.

## Suggested direction

A small, targeted sanitizer at the proxy boundary for text-only content paths (prompt reconstruction + streamed/non-stream text forwarding), stripping only internal transcript wrappers while preserving real tool semantics.

Patterns that seemed necessary in local testing:

- `<system-reminder>[\\s\\S]*?</system-reminder>`
- `<task_metadata>[\\s\\S]*?</task_metadata>`
- `<thinking>[\\s\\S]*?</thinking>`
- `<tool_output\\b[^>]*>[\\s\\S]*?</tool_output>`
- `<tool_exec\\b[^>]*\\/>` and paired form
- `<!-- OMO_INTERNAL_INITIATOR -->`
- `\\[SYSTEM DIRECTIVE: OH-MY-OPENCODE[^\\]]*\\]`
- `⚙ background_output [task_id=...]`
- stray `H:` / `A:` wrapper prefixes

## Why this matters

These wrappers are orchestration/debug transcript artifacts, not semantic content Claude needs. If they enter normal conversation text, they can confuse the model and create the impression that it is "talking to itself" again.

If useful, I can provide a minimal patch shape that only addresses these wrapper leaks without changing broader session/model logic.

rynfar:
## Problem

Agent harnesses (OpenCode, Droid, ForgeCode, oh-my-opencode) inject internal markup into message content — `<system-reminder>`, `<env>`, `<task_metadata>`, and similar tags. When the proxy flattens messages into a text prompt for the Agent SDK, these tags become model-visible text that causes the model to echo them back — the "talking to itself" symptom described in #167.

## Why not structured messages?

The ideal fix would be to always use `AsyncIterable<SDKUserMessage>` (structured messages) instead of string prompts, keeping tags as inert content in structured blocks. We built and tested this approach but discovered the Agent SDK sets `isSingleUserTurn = false` for AsyncIterable prompts, which changes how `endInput()` is called on the subprocess stdin. This breaks passthrough tool execution — the model sees "Forwarding to client for execution" as a tool result instead of forwarding the tool call to the client.

Until the SDK supports a `singleUserTurn` flag for AsyncIterable prompts, we need the string prompt path.

## Solution

Per-block sanitization: strip known orchestration tags from individual text content blocks **before** flattening to a string prompt — not from the final concatenated string.

This approach:
- **Eliminates cross-message regex risk** — each content block is sanitized independently
- **Uses exact tag names, not prefix patterns** — no `task_*` / `tool_*` / `system-*` wildcards that could eat legitimate content
- **Preserves the string prompt path** — passthrough tool execution continues to work
- **Is a pure leaf module** — no I/O, no imports from server.ts or session/

### Tags stripped (exact names)

| Tag | Harness | Purpose |
|-----|---------|---------|
| `system-reminder` | Droid | CWD/env injection in first user message |
| `env` | OpenCode, Crush | Environment context blocks |
| `system_information` | ForgeCode | System info wrapper |
| `current_working_directory` | ForgeCode | CWD tag |
| `operating_system` | ForgeCode | OS identifier |
| `default_shell` | ForgeCode | Shell path |
| `home_directory` | ForgeCode | Home dir path |
| `task_metadata` | OpenCode | Task tracking |
| `tool_exec` | OpenCode | Tool execution markers |
| `tool_output` | OpenCode | Tool output wrappers |
| `skill_content` | OpenCode | Skill injection |
| `skill_files` | OpenCode | Skill file lists |
| `directories` | OpenCode | Directory context |
| `available_skills` | OpenCode | Skill listing |
| `thinking` | Various | Leaked thinking tags (not structured blocks) |

Plus non-XML markers: `<!-- OMO_INTERNAL_INITIATOR -->`, `[SYSTEM DIRECTIVE: OH-MY-OPENCODE...]`, `⚙ background_output [task_id=...]`, and Meridian's own `Files changed:` summary.

## Why this supersedes #317

PR #317 used prefix-matching (`task_*`, `tool_*`, `skill_*`, `system-*`) and generic `antml:*` patterns that would strip legitimate content like `<tool-tip>`, `<task-list>`, `<system-status>`, `<br/>`, ``. It also had two regex bugs in the `antml:*` patterns where the opening tag matcher was `<\w+` (any tag) instead of `<\w+`.

This PR uses exact tag names only, operates per-block instead of on the full string, and leaves the text prompt path intact for passthrough compatibility.

## Test plan

- [x] 31 unit tests in `src/__tests__/sanitize-unit.test.ts` covering:
  - Every known orchestration tag (Droid, OpenCode, Crush, ForgeCode, oh-my-opencode)
  - Non-XML markers (OMO, background_output, Files changed)
  - Multiple patterns in one block
  - False-positive safety: HTML tags, self-closing tags, web components, angle brackets in code, prose mentions of tag names, legitimate XML with underscores
  - Regression guard: compound leakage pattern from #167
- [x] Full test suite: 1080 tests across 62 files, 0 failures
- [x] Type check clean: `npx tsc --noEmit`
- [x] E2E verified via `opencode run`:
  - Basic text response ✓
  - Tool use (passthrough) ✓
  - Multi-turn session resume ✓
  - `<system-reminder>` injection via curl → stripped, model answered correctly ✓
  - Multi-turn OpenCode TUI session — no wrapper echoing observed ✓

Fixes #167
Supersedes #317

### DIFF
--- a/src/__tests__/sanitize-unit.test.ts
+++ b/src/__tests__/sanitize-unit.test.ts
@@ -1,0 +1,194 @@
+/**
+ * Unit tests for the per-block content sanitizer.
+ *
+ * Verifies that orchestration wrappers injected by agent harnesses are
+ * stripped from individual text content blocks before prompt flattening,
+ * while preserving legitimate user content.
+ *
+ * Related: https://github.com/rynfar/meridian/issues/167
+ */
+
+import { describe, it, expect } from "bun:test"
+import { sanitizeTextContent } from "../proxy/sanitize"
+
+// ── Orchestration tag stripping ──
+
+describe("sanitizeTextContent", () => {
+  // --- Droid ---
+
+  it("strips <system-reminder> blocks (Droid CWD injection)", () => {
+    const input = '<system-reminder>\nUser system info\n% pwd\n/home/user\n</system-reminder>\nactual question'
+    expect(sanitizeTextContent(input)).toBe("actual question")
+  })
+
+  it("strips multiline <system-reminder> with attributes", () => {
+    const input = 'hello\n<system-reminder id="sr-1">\nline one\nline two\n</system-reminder>\nworld'
+    expect(sanitizeTextContent(input)).toBe("hello\n\nworld")
+  })
+
+  // --- OpenCode / Crush ---
+
+  it("strips <env> blocks (OpenCode environment context)", () => {
+    const input = '<env>\n  Working directory: /home/user/project\n  Platform: darwin\n</env>\nwhat is this project?'
+    expect(sanitizeTextContent(input)).toBe("what is this project?")
+  })
+
+  // --- ForgeCode ---
+
+  it("strips <system_information> wrapper and children", () => {
+    const input = '<system_information>\n<operating_system>Darwin</operating_system>\n<current_working_directory>/path</current_working_directory>\n<default_shell>/bin/zsh</default_shell>\n<home_directory>/Users/dev</home_directory>\n</system_information>\ndo the thing'
+    expect(sanitizeTextContent(input)).toBe("do the thing")
+  })
+
+  it("strips standalone <current_working_directory>", () => {
+    const input = '<current_working_directory>/Users/dev/project</current_working_directory>read file.ts'
+    expect(sanitizeTextContent(input)).toBe("read file.ts")
+  })
+
+  // --- OpenCode orchestration ---
+
+  it("strips <task_metadata> blocks", () => {
+    const input = '<task_metadata>{"id":"task-1","status":"running"}</task_metadata>actual content'
+    expect(sanitizeTextContent(input)).toBe("actual content")
+  })
+
+  it("strips <tool_output> wrappers with attributes", () => {
+    const input = 'before<tool_output name="bash">result here</tool_output>after'
+    expect(sanitizeTextContent(input)).toBe("beforeafter")
+  })
+
+  it("strips self-closing <tool_exec> wrappers", () => {
+    const input = 'text<tool_exec name="read" />more'
+    expect(sanitizeTextContent(input)).toBe("textmore")
+  })
+
+  it("strips paired <tool_exec> wrappers", () => {
+    const input = '<tool_exec name="bash">ls -la</tool_exec>output'
+    expect(sanitizeTextContent(input)).toBe("output")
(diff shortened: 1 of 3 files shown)


Zaczero:
skill_content	OpenCode	Skill injection

This disables the ability for OpenCode to load skills, making them useless. I finally figured out why Claude was behaving so oddly in OC with Meridian.

```
  it("strips <skill_content> blocks", () => {
    const input = '<skill_content name="gh">skill instructions</skill_content>rest'
    expect(sanitizeTextContent(input)).toBe("rest")
  })
```

?

PR #331, opened 2026-04-10
feat: add version field to /health endpoint response

lnilluv:
## Problem

Client tools (pi extensions, OpenCode plugins, Aider, etc.) have no way to know what version of Meridian is running. The npm package version and the running process version can diverge — the process was started days ago and the package updated since. This matters for:

- **Update notifications** — clients want to warn users when their running Meridian is outdated. Currently there's no way to compare the running version against the latest npm release without fragile heuristics.
- **Feature detection** — clients adjust behavior based on what the running version supports (passthrough mode, multi-profile, Pi adapter, etc.).
- **Debugging** — knowing the exact running version is critical. `npm list -g` shows what's installed, not what's running.

## Solution

Add a `version` string field to all `/health` response branches:

```json
{
  "status": "healthy",
  "version": "1.34.1",
  "auth": { "loggedIn": true, "email": "...", "subscriptionType": "max" },
  "mode": "passthrough",
  "plugin": { "opencode": "configured" }
}
```

The same field appears in `degraded` and `unhealthy` responses.

## Design decisions

**Version flows through config, not `createRequire`.** The initial approach imported `package.json` directly in `server.ts` via `createRequire`. That fails in the bundled output because `../../package.json` resolves outside the package root when the code runs from `dist/cli-*.js`. Instead, `bin/cli.ts` already reads `../package.json` (which resolves correctly from `dist/cli.js`), so it passes `version` through the `ProxyConfig` to the server.

**Fallback is `"unknown"`.** When `createProxyServer` is called without `version` (tests, library usage), the health response returns `"unknown"` rather than omitting the field. This keeps the response schema consistent.

## Changes

| File | What |
|------|------|
| `src/proxy/types.ts` | Add optional `version?: string` to `ProxyConfig` |
| `src/proxy/server.ts` | Derive `serverVersion = finalConfig.version ?? "unknown"`, include in all `/health` responses |
| `bin/cli.ts` | Pass `version` (already imported from `package.json`) to `startProxyServer` |
| `src/__tests__/proxy-async-ops.test.ts` | Assert `version` field in all health states; fix env var cleanup to `delete` instead of assigning `undefined` |

## Backward compatibility

Non-breaking. Adding a field to an existing JSON response. All existing clients that consume `/health` simply see an extra field they can ignore.

The `/health` response shape is listed as a stable API contract in `CLAUDE.md`. This is an additive change — no fields removed, no types changed.

### DIFF
--- a/bin/cli.ts
+++ b/bin/cli.ts
@@ -166,7 +166,7 @@ export async function runCli(
     enableDiskProfileDiscovery()
   }
 
-  const proxy = await start({ port, host, idleTimeoutSeconds, profiles, defaultProfile })
+  const proxy = await start({ port, host, idleTimeoutSeconds, profiles, defaultProfile, version })
 
   // Handle EADDRINUSE — preserve CLI behavior of exiting on port conflict
   proxy.server.on("error", (error: NodeJS.ErrnoException) => {
--- a/src/__tests__/proxy-async-ops.test.ts
+++ b/src/__tests__/proxy-async-ops.test.ts
@@ -16,30 +16,31 @@ describe("proxy async ops", () => {
     expect(typeof proxyB.server.keepAliveTimeout).toBe("number")
   })
 
-  it("serves async health endpoint with unchanged response schema", async () => {
+  it("serves async health endpoint with correct response schema", async () => {
     const { app } = createProxyServer({ port: 0, host: "127.0.0.1" })
     const response = await app.fetch(new Request("http://localhost/health"))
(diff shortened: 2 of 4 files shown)


rynfar:
Thank you for this.  I thought about adding it last time i touched the telemetry board and then forgot!  Thank you, I'll review and get this merged.

rynfar (approved):
Looks good — clean, additive change. Ship it.

PR #338, opened 2026-04-10
Correct configuration for Pi

murrayju:
The provided example config for Pi does not work. It must be wrapped in a top-level `providers` key, and the `apiKey` must also be specified to activate the provider. I also added the `x-meridian-agent: pi` to the headers, to correctly inform meridian of the adapter.


rynfar:
Thanks for the fix @murrayju! The existing config was wrong — appreciate you catching that. Picked up your commits in #342 with a small cleanup (normalized indentation, simplified the setup text). Your authorship is preserved. Closing in favor of that PR.

PR #360, opened 2026-04-13
fix: stop appending file change summaries for opencode

briankeefe:
## Summary
- add an adapter-level file change summary opt-out and disable it for OpenCode
- stop registering PostToolUse/file-change summary handling for OpenCode responses and streams
- update OpenCode tests to assert the synthetic \`Files changed:\` block is no longer appended

## Verification
- bun test src/__tests__/file-changes-unit.test.ts src/__tests__/proxy-file-changes.test.ts
- bun run typecheck
- bun run build *(fails in this shell during postbuild because the repo requires Node >=22 and this environment has Node 20)*

### DIFF
--- a/src/__tests__/file-changes-unit.test.ts
+++ b/src/__tests__/file-changes-unit.test.ts
@@ -443,6 +443,10 @@ describe("extractFileChangesFromMessages", () => {
 describe("openCodeAdapter.extractFileChangesFromToolUse", () => {
   const { openCodeAdapter } = require("../../src/proxy/adapters/opencode") as typeof import("../proxy/adapters/opencode")
 
+  it("should disable synthetic file change summaries", () => {
+    expect(openCodeAdapter.shouldTrackFileChanges?.()).toBe(false)
+  })
+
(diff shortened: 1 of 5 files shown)


briankeefe:
Closing this since MERIDIAN_NO_FILE_CHANGES=1 already disables the synthetic file change summary globally, so I’m using the existing local config path instead.

briankeefe:
Reopening — confirmed via local testing that the Files changed: block is produced by Meridian (bundled inside the opencode-with-claude plugin). The MERIDIAN_NO_FILE_CHANGES env var alone doesn't help because the plugin loads Meridian internally. The adapter-level opt-out is the clean fix.

rynfar:
Superseded by #363 — same fix by @briankeefe (commits preserved with full attribution), plus additional regression guards: positive-path integration tests using `x-meridian-agent: pi` to prove the summary still works for non-OpenCode adapters, and per-adapter unit assertions that only OpenCode opts out. Thanks for the fix @briankeefe! Closing this in favor of #363.

PR #378, opened 2026-04-15
fix: prevent cross-session context leakage via fingerprint cache for headerless OpenCode requests

jhenderiks:
## Problem

When OpenCode dispatches requests through plugin-less code paths (OMO category routing, title generation), the Meridian plugin's `chat.headers` hook does not fire. These requests arrive at Meridian **without** an `x-opencode-session` header and fall through to fingerprint-based session lookup.

Because `storeSession()` dual-writes to both `sessionCache` (keyed by header) and `fingerprintCache` (keyed by content hash), a header-tracked session pollutes the fingerprint namespace. A subsequent headerless request with the same `SHA256(cwd + "\n" + firstUserMessage)` fingerprint matches the stored session. `verifyLineage` then classifies the new 1-message request as an "undo" of the stored N-message session and **resumes the old Claude SDK session with its full conversation history**, leaking context across independent OpenCode sessions.

## Root cause

Two independent issues combine:

1. **`storeSession` dual-write**: sessions tracked by header are also written to `fingerprintCache`, creating a collision surface for later headerless requests.
2. **Fingerprint "undo" false positive**: when a headerless OpenCode request (always a new session) has fewer messages than a stored session with the same fingerprint, `verifyLineage` classifies it as "undo" and resumes the old session.

### 1. `storeSession` guard (`src/proxy/session/cache.ts`)

Only populate the fingerprint cache when no session header is present. Header-tracked sessions are already keyed by ID in `sessionCache`; dual-writing to `fingerprintCache` creates the cross-session collision surface.

```diff
- if (fp) fingerprintCache.set(fp, state)
+ if (fp && !sessionId) fingerprintCache.set(fp, state)
```

### 2. Undo→diverged guard for headerless OpenCode (`src/proxy/server.ts`)

When the OpenCode adapter has no session header, downgrade "undo" classification to "diverged" (fresh session). Headerless OpenCode requests are always new sessions (category-dispatched or title-generation flows where the plugin hook didn't fire), never legitimate undos.

```diff
- const lineageResult = lookupSession(...)
+ let lineageResult = lookupSession(...)
+ if (lineageResult.type === "undo" && adapter.name === "opencode" && !agentSessionId) {
+   lineageResult = { type: "diverged" }
+ }
```

### 3. `getSessionId` alignment (`src/proxy/adapters/opencode.ts`)

`detectAdapter()` already recognizes both `x-opencode-session` and `x-session-affinity` as OpenCode signals, but `getSessionId()` only read the former. Without this fix, a request detected as OpenCode via `x-session-affinity` would have `agentSessionId === undefined`, causing the undo→diverged guard to fire incorrectly.

```diff
- return c.req.header("x-opencode-session")
+ return c.req.header("x-opencode-session") ?? c.req.header("x-session-affinity")
```

## Trade-off

Legitimate fingerprint-based undo is sacrificed for headerless OpenCode requests. This is acceptable because:

- OpenCode is designed to work with the Meridian plugin which provides session headers
- When the plugin hook doesn't fire, the request is always a new session, never an undo
- OpenCode's own undo mechanism works at the application level (modifies the message array), not through SDK session forking

## Tests

- Updated 3 existing tests (2 in `session-fingerprint-context.test.ts`, 1 in `session-lineage.test.ts`) to expect diverged behavior instead of fork for headerless OpenCode undo — the updated assertions are more consistent with the test names (e.g. "does NOT resume wrong project")
- Added 1 new regression test: headered session stored → headerless request with same fingerprint → must not resume or fork
- Full suite: **1224 pass, 26 fail** (main: 1223 pass, 26 fail — net +1 from the new regression test, 26 pre-existing failures unchanged)

### DIFF
--- a/src/__tests__/session-fingerprint-context.test.ts
+++ b/src/__tests__/session-fingerprint-context.test.ts
@@ -215,9 +215,10 @@ describe("Fingerprint resume: cross-project safety via lineage", () => {
       { role: "user", content: "list the project B files" },
     ], "sdk-project-b")
 
-    // Prefix overlap ("hello") → undo/branch detected → forks from project A
-    expect(getCaptured()?.options?.resume).toBe("sdk-project-a")
-    expect(getCaptured()?.options?.forkSession).toBe(true)
+    // OpenCode without session header: undo is downgraded to diverged to
+    // prevent cross-session fingerprint collisions (headerless requests
+    // from category-dispatched or title-generation flows).
+    expect(getCaptured()?.options?.resume).toBeUndefined()
   })
 
   it("resumes correctly after cross-project rejection creates new session", async () => {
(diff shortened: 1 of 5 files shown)


rynfar:
Thanks catching this.  I made some changes to this area and have conflicts.  Im going to make a couple of minor adjustments to what youve pushed, do some extra manual testing since i also made changes in this area.  But your fix is real and correct, ill get it in today.  Thank you for submitting it!  I will cherry pick your changes to make sure you maintain credit for your work.

Issue #386, opened 2026-04-18
New version still shows the Tools message in the context


padixa:
```
[Tool Use: mcp__oc__Bash]({"command":"grep -n 'PeakEntry\|peakHistory\|waveformPosMs' /Users/files/pages/Time.ets","description":"grep Time key fields"})

H: [Tool Result for toolu_01GrmVFhj65P3FNJDdZL1DYB: 
19:interface PeakEntry {
23:}
...........
```

padixa:
```
Tool Use: Bash({"command":"grep -n 
```
such a disaster.... why always having this issue? I thought long time ago already fixed.....

padixa:
I've used below docker compose file (with claude code extention for VSCode):
```
services:
  meridian:
    image: ghcr.io/rynfar/meridian:latest
    container_name: meridian
    restart: unless-stopped
    environment:
      - MERIDIAN_HOST=0.0.0.0
      - MERIDIAN_PORT=3456
      - MERIDIAN_API_KEY=XXXXXXXXXX
      - MERIDIAN_PASSTHROUGH=1
      - MERIDIAN_DEFAULT_PROFILE=home
      - MERIDIAN_PROFILES=[{"id":"home","claudeConfigDir":"/profiles/home"},{"id":"home2","claudeConfigDir":"/profiles/home2"},{"id":"home3","claudeConfigDir":"/profiles/home3"}]
      - HTTP_PROXY=http://x.x.x.x:x
      - HTTPS_PROXY=http://x.x.x.x:x
      - NO_PROXY=localhost,127.0.0.1,::1
    volumes:
      - /home/meridian/default:/home/claude/.claude      # default
      - /home/meridian/profiles/home:/profiles/home
      - /home/meridian/profiles/home2:/profiles/home2
      - /home/meridian/profiles/home3:/profiles/home3
    ports:
      - "192.168.1.77:3456:3456"
```

padixa:
Could we just do more test with VSCode?  not only OpenCode

5632515:
I met the same problem... even in opencode

rynfar:
> ```
> Tool Use: Bash({"command":"grep -n 
> ```
> such a disaster.... why always having this issue? I thought long time ago already fixed.....

such a disaster?  It's not a simple fix pal.  What we are doing in Meridian is very tricky to get this to be have perfectly in every situation.  Feel free to explore the code yourself and try and solve it.  I will look again.

rynfar:
[...]

PR #390, opened 2026-04-19
Fix multimodal image handling for OpenAI chat completions and nested tool_result content

bardusco:
This PR fixes two related multimodal issues in Meridian's Anthropic-compatible proxy layer.

### Problem

There were two separate cases where image inputs were being lost before they reached the Claude Agent SDK:

1. **OpenAI-compatible `/v1/chat/completions` dropped `image_url` parts**
   - `translateOpenAiToAnthropic()` reduced structured message content to plain text
   - `image_url` blocks were discarded instead of being translated to Anthropic image blocks
   - as a result, the SDK received only text and responded as if no image had been provided

2. **Images nested inside `tool_result.content[]` were not treated as multimodal**
   - multimodal detection only checked top-level content blocks
   - when a tool returned `content: [text, image]`, that message fell back to the text prompt path
   - the nested image was serialized into text/JSON instead of being preserved as structured multimodal content
   - this caused incorrect or hallucinated descriptions in real tool-driven flows (for example, when a client sends a file path, a tool reads the image, and the resulting `tool_result` contains the image)

#### OpenAI translation fixes

- translate `image_url` data URLs into Anthropic `image` blocks with base64 sources
- preserve `[Image attached]` placeholders when flattening prior multimodal turns into packed conversation history
- add tests covering multimodal extraction and OpenAI-compatible prompt translation

#### Nested multimodal `tool_result` fixes

- detect multimodal content recursively inside `tool_result.content[]`
- strip `cache_control` recursively for nested tool result content as well
- normalize multimodal `tool_result` payloads into structured user content (`text` + `image`) before sending them to the SDK
- skip assistant messages composed only of `tool_use` blocks in the structured multimodal path, since once the tool result has been normalized they add no semantic context and only introduce textual noise

### Why skipping assistant-only `tool_use` messages is intentional

In the structured multimodal path, the useful information is the actual tool output (for example `text + image` returned by `read`), not the fact that the assistant previously invoked the tool. Once the multimodal `tool_result` is normalized into structured user content, keeping a synthetic textual summary like `[Tool Use: read(...)]` is redundant and can pollute the prompt without adding semantic value.

This behavior is limited to the structured multimodal path and only applies when the assistant message contains **only** `tool_use` blocks.

### Tests

Added/updated coverage for:

- `extractOpenAiContent()` preserving image placeholders
- `translateOpenAiToAnthropic()` preserving data URL images in the final user message
- `/v1/chat/completions` preserving multimodal prompts through the HTTP compatibility layer
- `/v1/messages` switching to structured prompts when images are nested inside `tool_result.content[]`
- ensuring assistant messages containing only `tool_use` blocks are not injected as synthetic text in that structured multimodal path

### Validation

Validated with:

- `bun test src/__tests__/openai.test.ts src/__tests__/proxy-openai-compat.test.ts src/__tests__/proxy-multimodal.test.ts`
- `bunx tsc --noEmit`
- `bun run build`

Also manually validated against a real screenshot flow where the image was returned from a tool as `tool_result.content = [text, image]`. Before this patch the model hallucinated the image contents; after this patch it described the screenshot correctly.

### Notes

- `image_url` support in this patch is limited to **data URLs**
- remote image fetching is intentionally left out of scope


rynfar:
Thanks for this!  I'll retry to get this reviewed tonight.

rynfar:
Thanks @bardusco for the excellent analysis and fixes on this. I've opened #392 which cherry-picks your three commits (preserving authorship) and adds a small cleanup on top — the `isToolUseOnlyAssistantMessage` guard became redundant once #386 merged (`flattenAssistantContent` already returns `""` for tool-use-only messages, which the `if (assistantText)` guard drops).

Verified live against real Claude (haiku-4.5): both your bugs reproduce on v1.37.4 and are fully fixed on #392 (OpenAI `image_url` → "red"; nested `tool_result` image → "magenta"; unfixed returned "I don't see an image" and hallucinated "blue" respectively). Full suite 1269/1269.

Closing this in favor of #392 so your commits land with your authorship intact.

PR #396, opened 2026-04-19
fix: normalize context-usage to last iteration

jhenderiks:
## Problem

Meridian's `/v1/sessions/:claudeSessionId/context-usage` endpoint currently returns the raw `session.contextUsage` object that was last observed from the Claude Agent SDK.

That works for simple requests, but it can be wrong for long-running Claude Code / OpenCode sessions when the SDK includes an `iterations` breakdown. In those cases, the top-level usage fields can represent cumulative usage across internal iterations, while the current context-window snapshot is the **last iteration**.

The practical symptom is inflated context-usage reporting in clients that consume this compatibility endpoint. In real Meridian-backed Opus sessions, this showed up as context counters jumping far above the actual active context for a turn.

## Root cause

The endpoint was returning `session.contextUsage` verbatim:

- if `iterations` was absent, this was fine
- if `iterations` was present, the endpoint still returned the cumulative top-level usage instead of the last iteration

The Claude Agent SDK itself treats `result.usage` as cumulative query usage, while current context usage is surfaced separately. This compatibility endpoint is intended to report the **current active context snapshot**, not cumulative totals across internal iterations.

## Fix

- extend `TokenUsage` to model optional `iterations`
- normalize `/v1/sessions/:claudeSessionId/context-usage` to `usage.iterations.at(-1)` when present
- fall back to the original top-level usage object when `iterations` is absent

This keeps existing behavior unchanged for normal sessions while fixing inflated context reporting for iterative SDK runs.

This is also an intentional compatibility choice: the endpoint now returns the effective current usage view for clients, not the raw stored SDK payload (including `iterations`) when those two differ.

## Tests

Added regression coverage for both storage paths Meridian uses:

- in-memory session cache path in `src/__tests__/proxy-sdk-params.test.ts`
- shared session store path in `src/__tests__/proxy-context-usage-store.test.ts`

Both tests store a usage object where:
- the top-level fields are intentionally cumulative and very large
- the last `iterations[]` entry is smaller and represents the actual current context snapshot

They assert that the endpoint returns the last iteration, not the cumulative top-level object.

## Validation

- `bun test src/__tests__/proxy-sdk-params.test.ts`
- `bun test src/__tests__/proxy-context-usage-store.test.ts`
- `bun run typecheck`
- `bun run build`

## Notes

This PR only changes the compatibility endpoint response. It does **not** alter how usage is stored internally, so raw SDK usage data is still preserved in session state.

### DIFF
--- a/src/__tests__/proxy-context-usage-store.test.ts
+++ b/src/__tests__/proxy-context-usage-store.test.ts
@@ -74,4 +74,55 @@ describe("GET /v1/sessions/:claudeSessionId/context-usage — shared store", () 
     expect(usage.input_tokens).toBe(77)
     expect(usage.output_tokens).toBe(11)
   })
+
+  it("returns the last usage iteration from the shared session store", async () => {
+    const { app } = createProxyServer({ port: 0, host: "127.0.0.1" })
+    const claudeSessionId = "sess_shared_usage_iterations_001"
+
+    storeSharedSession(
+      "shared-key-usage-iterations",
+      claudeSessionId,
+      1,
+      "lineage",
+      ["msg-hash"],
+      [null],
+      {
+        input_tokens: 9000,
+        cache_creation_input_tokens: 250000,
+        cache_read_input_tokens: 700000,
+        output_tokens: 1200,
+        iterations: [
+          {
+            input_tokens: 9000,
+            cache_creation_input_tokens: 250000,
+            cache_read_input_tokens: 700000,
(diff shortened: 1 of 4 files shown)


rynfar:
Thanks for this sorry for the delay 8in getting to it.  Been working on the plugin system.  Im going to cherry pick your changes so you remain credited for your work, but i have a copule of small adjustments to make but the concept is spot on.

PR #401, opened 2026-04-20
feat: Claude Code adapter with split SDK / client working directory

davetha:
## Summary

Adds a Claude Code CLI adapter and introduces a `clientWorkingDirectory` concept so remote clients (e.g. Claude Code pointing at Meridian over the network) get accurate working-directory context.

Surfaced while running Claude Code on another machine on my home LAN, pointed at a Meridian instance running on the box that is logged into Claude Max. Everything worked great through Meridian — streaming, tools, passthrough — except the model always answered the server's path when asked "what directory are you in?", and all file-path references were framed from the proxy host.

## Root cause

`workingDirectory` was being used for two independent purposes:

1. **SDK subprocess `cwd`** — must exist on the proxy host, passed as `cwd` to the Claude Agent SDK's `query()`.
2. **Fingerprinting + system-prompt env** — should reflect the client's real project for per-project bucketing and path accuracy.

For same-host clients (OpenCode, Crush) both collapse to one value, so the coupling was invisible. For a remote client whose local path doesn't exist on the proxy, using it as SDK cwd fails; *not* using it means the model sees the proxy's cwd.

## Changes

- `AgentAdapter.extractClientWorkingDirectory` (optional) — adapters can return the client-local path (may not exist on the proxy host). Same-host adapters can leave it unset and the two values collapse.
- `server.ts` now resolves two values: `workingDirectory` (SDK cwd, guaranteed valid) and `clientWorkingDirectory` (used for `profileScopedCwd` and the system-prompt addendum).
- `query.ts` exports `buildCwdNote` which emits an `<env>Working directory: …</env>` block in the Claude-Code-CLI-native format when the two differ. This pre-empts the subprocess' own env-block injection so the model sees the client path. Followed by a short `<meridian-note>` clarifying the proxy path.
- New `claudeCodeAdapter` — parses `Primary working directory:` from Claude Code's `# Environment` block, returns `undefined` from `extractWorkingDirectory` (so the SDK keeps a valid cwd on the proxy host).
- User-Agent detection for `claude-cli/*`, plus `x-meridian-agent` header aliases (`claude-code`, `claudecode`).
- README: Claude Code added to the supported-agents list, new Agent Setup subsection, Tested Agents row.

## Test plan

- [x] Typecheck clean (`bun run typecheck`)
- [x] Build clean (`npm run build`)
- [x] Full suite (`npm test`) — all pass
- [x] New tests:
  - `claude-code-adapter.test.ts` (24 tests) — identity, `extractWorkingDirectory`, `extractClientWorkingDirectory` (string / array / case / edge cases), basic config surface, `extractFileChangesFromToolUse`
  - `adapter-detection.test.ts` additions (5 tests) — `claude-cli/*` UA, `x-meridian-agent: claude-code`, `claudecode` alias
  - `query-cwd-note.test.ts` (7 tests) — `buildCwdNote` formatting and no-op cases
- [x] **Live E2E on a home-LAN setup:**
  - Meridian running as systemd service on the host that is logged into Claude Max
  - Claude Code running on a separate machine, `ANTHROPIC_BASE_URL=http://meridian:3457 claude`
  - Fresh session: `what dir are you in?` → returns the client's real path (previously returned the proxy host's `/home/claude/proxy2`)
  - Existing sessions keep their cached prompt until reopened — expected Claude Code behavior

## Notes

Claude Code as a client is a slightly unusual case (Meridian itself uses the Claude Code SDK internally), but it's a legitimate use-case for sharing a single Max subscription across machines on a trusted network. README documents the setup with a note about `ANTHROPIC_AUTH_TOKEN` vs `ANTHROPIC_API_KEY`.

The `<env>` block format in `buildCwdNote` is chosen specifically to match what the Claude Code subprocess uses for its own env injection — so the subprocess sees "env already present" and doesn't add a second one pointing at its actual (proxy) cwd.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ---
 
-Meridian bridges the Claude Code SDK to the standard Anthropic API. No OAuth interception. No binary patches. No hacks. Just pure, documented SDK calls. Any tool that speaks the Anthropic or OpenAI protocol — OpenCode, ForgeCode, Crush, Cline, Aider, Pi, Droid, Open WebUI — connects to Meridian and gets Claude, with session management, streaming, and prompt caching handled natively by the SDK.
+Meridian bridges the Claude Code SDK to the standard Anthropic API. No OAuth interception. No binary patches. No hacks. Just pure, documented SDK calls. Any tool that speaks the Anthropic or OpenAI protocol — OpenCode, ForgeCode, Crush, Cline, Aider, Pi, Droid, Open WebUI, Claude Code — connects to Meridian and gets Claude, with session management, streaming, and prompt caching handled natively by the SDK.
 
 > [!NOTE]
 > ### How Meridian works with Anthropic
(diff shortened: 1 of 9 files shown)


AdamFarrar:
This is great!!

rynfar:
Thanks for this, @davetha — the split-cwd abstraction is clean and the `<env>` pre-emption trick is a nice piece of subprocess-behavior work. Picked up your commit verbatim with authorship preserved in #426 after rebasing on current `main` (your PR conflicted with the stack of changes that landed after April 20 — advisor support in #413, maxTurns in #420, passthrough normalization in #425 — all of which touched `query.ts` and `server.ts` near your edits).

Also folded in one small docs follow-up: an explicit LAN security warning in the README recommending `MERIDIAN_API_KEY` for multi-machine setups. The original "any string works" note was literally true but risks leaking a Max subscription when the proxy binds beyond loopback.

E2E verified on a live proxy: `claude-cli/*` UA routes to `adapter=claude-code` and a fake `Primary working directory: /Users/alice/my-project` in the system prompt causes the model to report that exact path when asked about its cwd — confirms the full chain works end-to-end.

Closing this in favor of #426. Appreciate the careful design and thorough test coverage (36 new tests all pass).

PR #402, opened 2026-04-20
feat: persistent SDK sessions — hold one live query() per logical session

cartwmic:
## Summary

Replaces request-per-process `query({ resume })` with one live `query()` held per logical session and fed via streaming-input mode. Resumed turns now hit the Anthropic prompt cache (93–98% warm-hit) because the SDK's in-memory conversation stays byte-continuous across HTTP turns, instead of being re-serialized with byte drift every resume.

**Primary goal met:** warm-resume cache hit is observable end-to-end through Pi + Claude Max.

Feature-flagged on `ProxyConfig.persistentSessions` (env: `MERIDIAN_PERSISTENT_SESSIONS=1`). Off-flag behavior is bit-identical to today.

## What's in the branch

- **Runtime layer** (`src/proxy/session/`): `runtime.ts` (live-query wrapper, input queue, per-turn mutex, pending-handler registry, prebound-result buffer), `persistentDispatch.ts` (turn dispatcher), `persistentWiring.ts` (SDK-aware construction), `optionsClassifier.ts` (options drift), `turnRunner.ts` (branching entry point that replaces direct `query()` at each `server.ts` call site).
- **Passthrough deferred-handler pattern** (design §D11): MCP handler blocks on a promise meridian resolves when the client returns with the real tool_result. No sentinel pollution, no synthetic "blocked by hook" narrative, no override user message. See `openspec/changes/persistent-sdk-sessions/spike-notes.md` for the four passthrough approaches that were tested and rejected.
- **Server wiring**: 4 `query()` call-sites in `server.ts` routed through `turnRunner.startTurn`; runtime-manager singleton with periodic idle-sweep + LRU eviction; SIGTERM/SIGINT graceful shutdown; continuation-after-pending SSE framing for multi-response assistant turns; mutex-timeout → HTTP 429 + Retry-After.
- **Observability**: `persistent.turn` + `persistent.lifecycle` trace events; `SessionRuntimeManager.counters` for telemetry.
- **CLI env vars**: `MERIDIAN_PERSISTENT_SESSIONS`, `MERIDIAN_PERSISTENT_IDLE_MS`, `MERIDIAN_PERSISTENT_MAX_LIVE`, `MERIDIAN_PERSISTENT_MUTEX_WAIT_MS`, `MERIDIAN_PERSISTENT_PENDING_TIMEOUT_MS`.
- **Per-adapter rollout gate**: `AgentAdapter.usesPersistentSessions?(): boolean | undefined` overrides the global flag.
- **Critical bug fix (this week)**: Parallel `tool_use` blocks in one assistant message used to deadlock because only the first MCP handler was pending when the client's batched tool_results arrived (SDK fires handlers sequentially); the dispatcher diverted the orphan tool_result to `pushContent` which the SDK couldn't consume while blocked in its handler chain. Fixed with a prebound-result buffer (`runtime.prebindPendingResult`) so later-firing handlers resolve from the buffer when they register. See `spike/e-f-repro.ts` for the isolated reproduction (CURRENT=timeout vs BUFFERED=pass with cacheRead≈43k).

## Test plan

Unit tests (`bun test`):
- [x] Full library-layer coverage: `session-runtime-unit.test.ts` (SessionRuntime, AsyncQueue, Mutex, prebound buffer), `options-classifier-unit.test.ts`, `persistent-dispatch-unit.test.ts`, `persistent-wiring-unit.test.ts`, `passthrough-deferred-unit.test.ts`, `mock-query-unit.test.ts`, integration lifecycle tests. 1283 pass / 0 fail. `tsc --noEmit` clean. `openspec validate --strict persistent-sdk-sessions` clean.

Live Pi playbook (real Pi 0.67.68 + real Claude Max, Sonnet 4.6):
- [x] A single-hop plain: `OK.`
- [x] B multi-hop plain text: T2 cacheRead **98%** on resume
- [x] C single-hop single tool: `@rynfar/meridian`
- [x] D multi-hop single tool: 3 coherent turns, cache 73–97%
- [x] E single-hop parallel tools: `ACK read package.json and CLAUDE.md.` (prebind fix target)
- [x] F multi-hop parallel tools: correct answers with cache 61–93% on continuation turns (one F.1 attempt hit an upstream Anthropic API timeout; F.2 + F.3 of the same shape passed)
- [x] G mixed plain/tool: all 4 turns correct
- [x] H cold-reattach after proxy restart: recall `42`, T3 cache 97%
- [x] L LRU idle eviction: session preserved on disk and resumed correctly after eviction
- [x] M mutex 429: second concurrent same-session request returns `HTTP 429` + `Retry-After: 2`
- [x] N graceful shutdown: SIGTERM → clean exit in ~1s, 0 leaked `cli.js` subprocs

Off-flag: no behavior change. Verified by keeping the legacy code path in `turnRunner.startTurn` as the else-branch; 100% of existing tests pass with the flag off.

## Known follow-ups (out of scope for this PR)

- Openspec §5.5c: mutex isn't held during the drift-check + reopen path. Small race window under concurrent close+reopen; doesn't affect single-request correctness.
- Curl-path `messages.0.content.3.type: Field required` on the opencode adapter for non-trivial message arrays — pre-existing (reproduces on upstream main without any changes from this branch), reported separately.

## Docs

- `CLAUDE.md`: new "Persistent-SDK-sessions mode" section.
- `ARCHITECTURE.md`: `session/` module map updated.
- `E2E.md`: "Persistent-SDK-sessions mode" scenarios P1–P4.
- Openspec change `openspec/changes/persistent-sdk-sessions/` (proposal, design with 12 decisions, spec, tasks, spike-notes).

## Branch

22 commits ahead of `upstream/main` at `c586476`. Squash-merge is fine; individual commits also tell a coherent story if you prefer a regular merge.


cartwmic:
Closing — pausing the upstream merge. Will re-open when ready.

cartwmic:
please disregard, opened in error

PR #405, opened 2026-04-20
chore: improve nix opencode integration

Dreaming-Codes:
This PR adds:

- A simple way to enable the Meridian plugins in OpenCode via `services.meridian.opencode.enable = true`, along with `services.meridian.opencode.agentMode` and `services.meridian.opencode.claudeMaxHeaders`
- A `setSessionVariables` option that sets `ANTHROPIC_API_KEY` and `ANTHROPIC_BASE_URL` globally

Notes before merging:

`agentMode` and `claudeMaxHeaders` currently point to plugins inside the example folders, but I believe these plugins are useful enough that they should be moved out of the examples and treated as recommended plugins, since they are not really just examples.

Personally, I think they could even be merged into the main OpenCode plugin required for Meridian to work. I know there is already an OpenCode plugin that bundles everything together, but it also runs its own proxy. My use case is to have a single proxy on my machine that all apps can access, so that setup does not work for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,13 +130,29 @@ Then in `~/.config/opencode/opencode.json`:
 
 The service starts automatically on login. Manage it with `systemctl --user {start,stop,restart,status} meridian`.
 
-The plugin path is also available as `config.services.meridian.opencode.pluginPath` for use in your OpenCode config:
+**Session variables** -- set `ANTHROPIC_API_KEY` and `ANTHROPIC_BASE_URL` globally so any Anthropic-compatible tool points at Meridian:
 
 ```nix
-xdg.configFile."opencode/opencode.json".text = builtins.toJSON {
-  plugin = [ config.services.meridian.opencode.pluginPath ];
+services.meridian = {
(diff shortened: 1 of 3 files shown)


rynfar:
great input - i am actually currently designing a proper plugin surface for meridian and will be merging it today or tomorrow.  Along with an open code plugin, this would be a much better way to extend this and then it can easily be customized for your use case.

rynfar:
Thanks for this, @Dreaming-Codes — the home-manager integration is thoughtful and the `setSessionVariables` option is a nice touch.

Closing this one: the functionality the `claudeMaxHeaders` and `agentMode` example plugins provided has since been folded into the main `plugin/meridian.ts`, so installing them as separate files is no longer the recommended path. The single-plugin install via `services.meridian.opencode.enable` would still make sense as a follow-up, but we'd want it to only wire up the main plugin — not the examples — to match the current architecture. Happy to take a narrower PR along those lines if you're still using Meridian on NixOS!

PR #406, opened 2026-04-20
Use Lix-compatible flake input syntax

Mewski:
This flake breaks when using Lix. Lix does not support the `?tag=` attribute for the `github` scheme. The `/ref` syntax is supported by both Nix and Lix.


rynfar:
I don't have a way to test this so i'll need to rely on you to confirm it's working 😁. I'll review the pr but i'm gonna trust this works.  Hah.

Mewski:
This works for me with both Nix and Lix.

PR #407, opened 2026-04-20
Preserve advisor semantics and reject unsupported advisor requests

WarGloom:
## Summary
- preserve advisor-shaped response semantics when the Claude Agent SDK emits them
- reject native advisor tool requests that Meridian cannot safely execute through the current SDK path
- add regression coverage for advisor beta forwarding, response preservation, and request rejection

## Motivation
Meridian already forwards the `advisor-tool-2026-03-01` beta header, but its proxy behavior still assumes ordinary client tool flows.

That creates two problems:
- native advisor requests can look superficially supported even though Meridian cannot actually execute them through the current Claude Agent SDK boundary
- if upstream emits advisor-shaped response data like `server_tool_use`, `advisor_tool_result`, or `pause_turn`, Meridian can distort that behavior by flattening unknown blocks or recomputing stop reasons locally

This change makes Meridian more honest and safer:
- unsupported native advisor requests now fail explicitly instead of being silently misrouted
- advisor-related response semantics are preserved when they already exist in upstream SDK output

### Request handling
- detect native Anthropic advisor tool definitions in `tools[]`
- reject those requests on `/v1/messages`
- reject those requests on `/v1/chat/completions`

### Response preservation
- preserve upstream non-stream `stop_reason` instead of always recomputing it from local heuristics
- keep advisor-related content blocks compatible with Meridian prompt rebuilding and hashing paths
- preserve `server_tool_use` and `advisor_tool_result` in response handling paths

### Continuity and replay
- add advisor-aware content normalization and prompt summarization helpers
- prevent replay and retry paths from flattening advisor/server-tool blocks into misleading text-only fallbacks

## Testing
Passed locally:
- `bun test src/__tests__/messages.test.ts src/__tests__/proxy-advisor-support.test.ts`
- `bun test src/__tests__/proxy-sdk-params.test.ts --test-name-pattern "advisor-tool anthropic-beta header is forwarded for claude-max profiles"`

## Limitations
This does not add full native advisor execution support.

Meridian runs through `@anthropic-ai/claude-agent-sdk`, and the current SDK path does not expose a safe request-side way to inject a native Anthropic advisor server tool into `query()`.

So the behavior in this PR is intentionally:
- preserve advisor semantics when upstream emits them
- reject unsupported native advisor requests instead of pretending they work

### DIFF
--- a/src/__tests__/messages.test.ts
+++ b/src/__tests__/messages.test.ts
@@ -2,7 +2,7 @@
  * Unit tests for message parsing utilities.
  */
 import { describe, it, expect } from "bun:test"
-import { normalizeContent, getLastUserMessage } from "../proxy/messages"
+import { extractAdvisorToolDefinition, getLastUserMessage, normalizeContent, summarizeContentBlockForPrompt } from "../proxy/messages"
 
 describe("normalizeContent", () => {
   it("returns string content as-is", () => {
@@ -32,6 +32,19 @@ describe("normalizeContent", () => {
     const result = normalizeContent(content)
     expect(result).toContain("tool_result:tu_1:")
     expect(result).toContain('"key":"val"')
+  })
+
+  it("handles server_tool_use blocks", () => {
+    const content = [{ type: "server_tool_use", id: "srv_1", name: "advisor", input: {} }]
+    const result = normalizeContent(content)
+    expect(result).toBe("server_tool_use:srv_1:advisor:{}")
+  })
+
+  it("handles advisor_tool_result blocks", () => {
+    const content = [{ type: "advisor_tool_result", content: [{ type: "advisor_result", text: "Plan first." }] }]
+    const result = normalizeContent(content)
+    expect(result).toContain("advisor_tool_result:")
+    expect(result).toContain("advisor_result")
   })
 
   it("handles mixed content blocks", () => {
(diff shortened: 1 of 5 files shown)


rynfar:
Thanks for this PR @WarGloom — the thinking behind "be honest about what we can't do" is exactly right, and this surfaced a real gap we needed to address.
However, after investigating the Claude Agent SDK, it turns out advisor support is available via the advisorModel query option. The latest SDK (0.2.116) has:
typescript

advisorModel?: string;  // "Advisor model for the server-side advisor tool."
This means instead of rejecting advisor tool requests, we can extract the advisor tool definition from the client's tools[] and map it to the SDK's advisorModel option — giving us full native advisor support through Meridian.


I am going to work toward native support isntead, thank you for submitting this!  It raised a real issue.  If my thinking is off let me know.

rynfar:
Hey @WarGloom — I've got native advisor support drafted on [`feat/advisor-support`](https://github.com/rynfar/meridian/tree/feat/advisor-support) (commit b4841f5a). It extracts `advisor_*` tool definitions from `tools[]` and forwards them to the SDK's `advisorModel` query option, plus preserves non-standard `stop_reason` values like `pause_turn`.

I need a favor before merging: **I can't E2E this on my account.** The SDK gates advisor activation behind the `tengu_sage_compass` Statsig flag (`GQ()` in the bundle), which isn't enabled for my account — when I send advisor requests through my proxy, the SDK silently no-ops and the model reports "no advisor tool available." Since you originally observed advisor-shaped responses in the wild, it sounds like your account *does* have the path active.

Could you:

1. Check out `feat/advisor-support`, build it, and point your client at it
2. Run a request that previously produced advisor content blocks — confirm the response now round-trips cleanly (server_tool_use, advisor_tool_result, pause_turn preserved)
3. **Especially:** watch for `Reached maximum number of turns` errors. I bumped `maxTurns` by `+3` for advisor requests ([query.ts:149](https://github.com/rynfar/meridian/blob/feat/advisor-support/src/proxy/query.ts#L149)) based on a guess of "call + result + final answer." If the model calls advisor multiple times per task (the SDK system prompt encourages it), `+3` may be too low and we'll need to bump it

If you hit max-turns errors, grab the request ID and the task shape and I'll tune the constant. If it all works, I'll merge and close this PR.

Thanks again for surfacing this — the original PR was right about the problem; this branch is just a different angle on the fix.

rynfar:
@WarGloom  did this resolve your problem can i merge it?

WarGloom:
@rynfar Sounds awesome ,  I'm out of limits for now, Will check it tomorrow, thank you

WarGloom:
@rynfar answered in PR 413

rynfar:
Closing this in favor of #413, which landed actual advisor support via the SDK's `advisorModel` query option. The rejection approach here was based on the (correct-at-the-time) assumption that the SDK couldn't execute native advisors — turns out `claude-agent-sdk@0.2.90` already handled it, so we were able to pass through instead of rejecting.

Thanks for the diligence on preserving upstream response semantics — that insight made it directly into #413 (the `stop_reason` preservation for `pause_turn`). Appreciated.

Issue #408, opened 2026-04-20
`passthrough` agent completely ignores `codeSystemPrompt: false` setting


albe-jj:
**Description:**
In version 1.37.4, disabling the "Claude Code Prompt" for the `passthrough` agent (either via the Web UI or by manually editing `sdk-features.json`) is completely ignored by the proxy. Meridian continues to inject the massive ~28k-token `claude_code` preset into passthrough requests, bypassing the token-saving fix introduced in Issue #190.

This appears to be a two-part bug:
1. **UI Bug:** The Web UI (`/settings`) fails to write `"codeSystemPrompt": false` to the `passthrough` block in the config file when toggled off.
2. **Proxy Engine Bug:** Even when the config file is manually corrected to include `"codeSystemPrompt": false`, the proxy engine ignores it and injects the preset anyway when the `x-meridian-agent: passthrough` header is present.

**Steps to Reproduce:**
1. Manually edit `~/.config/meridian/sdk-features.json` to explicitly disable the preset for `passthrough`:
   ```json
   "passthrough": {
     "clientSystemPrompt": true,
     "codeSystemPrompt": false,
     "sdkDebug": true
   }

rynfar:
I'll review this thanks for reporting it

rynfar:
I am sorry for the delay @albe-jj  I am looking into this right now.  I will try to get a release cut with this fix in it today.

rynfar:
Fix landed in #469 — just merged to \`main\`. Will go out in 1.42.0.

The proxy engine bug was exactly what you described — \`server.ts\` had \`sdkFeatures.codeSystemPrompt ? true : undefined\` at four call sites, which collapsed an explicit \`false\` from your config into \`undefined\` and let \`query.ts\`'s historical fallback re-apply the preset. Two-layer fix:

1. \`server.ts\` now passes the value through directly (4 sites, all identical).
2. \`sdkFeatures.ts\` encodes the historical "preset by default" behavior explicitly so changing the server.ts behavior doesn't silently drop the preset for adapters that historically had it: \`DEFAULT_FEATURES.codeSystemPrompt = true\`, with \`passthrough\` overridden to \`false\` via \`ADAPTER_DEFAULTS\` (preserves the #190 token-saving fix).

Behavior matrix after the fix:

| Case | Result |
|---|---|
| Default OpenCode/Crush/Droid/etc. | preset (matches historical) |
| Default passthrough | no preset (#190 preserved) |
[...]

PR #418, opened 2026-04-22
fix: passthrough agent type normalization and fallback routing

Alexander-Ollman:
## Summary

Fixes subagent spawning failures when OpenCode runs through the proxy in passthrough mode. Three related issues addressed across 3 commits:

### 1. Parameter name mismatch (existing commit)
SDK system prompt describes tools with snake_case params (`file_path`), but OpenCode defines them as camelCase (`filePath`). Model generates SDK naming, client rejects with "expected string, received undefined". Added `normalizeToolInput()` to remap snake_case ↔ camelCase against the client's input_schema.

### 2. Unknown agent types
Claude invents `subagent_type` values not in the parsed agent set (e.g., PascalCase `"Explore"`, aliases like `"general-purpose"`). SDK rejects these as unregistered. Added:
- Default agent type injection (`build`, `plan`, `explore`, `general`) with case-variant registration
- `ensureDefaultAgents()` only injects when parsing yielded at least one agent
- `addCaseVariants()` registers PascalCase and alias variants
- Fallback routing to `"general"` for completely unknown types in `fuzzyMatchAgentName()`

### 3. Non-streaming normalization bypass + streaming buffer
**Non-streaming**: SDK content blocks already contain tool_use with raw (un-normalized) input. PreToolUse hook normalizes and captures, but dedup check (`contentBlocks.some(b => b.id === tu.id)`) found the raw block already present and skipped the captured version. Changed to merge strategy using `capturedById` Map — replaces existing content block inputs with normalized versions.

**Streaming**: `input_json_delta` sends JSON in small chunks, so regex on individual deltas can't reliably find `subagent_type`. Buffer all `input_json_delta` chunks for Task tool blocks and emit the complete fixed JSON at `content_block_stop`.

## Changes

| File | Description |
|------|-------------|
| `src/proxy/passthroughTools.ts` | `normalizeToolInput()` — snake_case ↔ camelCase |
| `src/proxy/agentDefs.ts` | `FALLBACK_AGENT_NAME`, `DEFAULT_AGENT_TYPES`, `ensureDefaultAgents()`, `addCaseVariants()` |
| `src/proxy/agentMatch.ts` | Fallback to `"general"` instead of raw lowercased input |
| `src/proxy/server.ts` | PreToolUse subagent_type normalization, capturedById merge, streaming buffer |
| `src/__tests__/*` | Tests for all of the above |

## Test plan

- [x] 1367/1371 tests pass (4 failures are pre-existing 5s timeouts from env contention)
- [x] E2E validated: streaming `"Explore"` → `"explore"` ✓
- [x] E2E validated: non-streaming `"General-Purpose"` → `"general"` ✓  
- [x] Read tool `file_path` → `filePath` parameter correctly normalized ✓
- [x] Unknown agent types fall back to `"general"` instead of SDK rejection ✓
- [x] Full proxy session tested with `era-code start` — all requests returning 200

### DIFF
--- a/src/__tests__/adapter.test.ts
+++ b/src/__tests__/adapter.test.ts
@@ -101,7 +101,7 @@ describe("openCodeAdapter.buildSdkAgents", () => {
     const agents = openCodeAdapter.buildSdkAgents!(body, [])
     for (const [name, def] of Object.entries(agents)) {
       expect((def as any).description).toBeTruthy()
-      expect((def as any).prompt).toContain(name)
+      expect((def as any).prompt.toLowerCase()).toContain(name.toLowerCase())
       expect((def as any).model).toBe("inherit")
     }
   })
--- a/src/__tests__/passthrough-param-normalization.test.ts
+++ b/src/__tests__/passthrough-param-normalization.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, it } from "bun:test"
+import { normalizeToolInput } from "../proxy/passthroughTools"
+
+describe("normalizeToolInput", () => {
+  it("returns input unchanged when all required fields are present", () => {
+    const input = { filePath: "/src/app.ts", offset: 0 }
+    const schema = {
+      properties: { filePath: { type: "string" }, offset: { type: "number" } },
+      required: ["filePath"],
+    }
+    expect(normalizeToolInput(input, schema)).toEqual(input)
+  })
+
+  it("converts snake_case to camelCase when required field is missing", () => {
+    const input = { file_path: "/src/app.ts" }
+    const schema = {
+      properties: { filePath: { type: "string" } },
+      required: ["filePath"],
+    }
+    expect(normalizeToolInput(input, schema)).toEqual({ filePath: "/src/app.ts" })
+  })
+
+  it("converts camelCase to snake_case when required field is missing", () => {
+    const input = { filePath: "/src/app.ts" }
+    const schema = {
+      properties: { file_path: { type: "string" } },
+      required: ["file_path"],
+    }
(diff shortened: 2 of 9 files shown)


Alexander-Ollman:
Issue with current Meridian build was a lot of incorrect formatting for OpenCode tools and sub-agents. Wrote a wrapper to correctly parse tools and subagent classes. Working really well with Claude Max!

Alexander-Ollman:
I think this might even address #408? @rynfar

rynfar:
thank you for this, i will try to look at this tonight!

rynfar:
Thanks for this — really thorough diagnosis across all three issues (param naming, agent registration, and the streaming-delta buffer). I picked up your three commits verbatim with authorship preserved in #425, plus a small refactor on top:

- Centralized the `general-purpose → general` mapping into a single `resolveAgentAlias()` helper in `agentMatch.ts` so it's not duplicated between the non-stream and stream paths
- Replaced the streaming regex with `JSON.parse` / `JSON.stringify` for robustness (the buffer is complete JSON at block_stop, so parsing is safe)

E2E verified all three fixes through live opencode → proxy flow with haiku:
- `file_path` → `filePath` ✅
- `"Explore"` (non-stream) → `"explore"` ✅
- `"general-purpose"` (stream) → `"general"` ✅

Also ran the unit suite: **1395/1395 pass** (3 new tests for the helper).

Re: #408 — the diff here doesn't touch the `codeSystemPrompt` path in `query.ts` that #408 is about, so closing #418 shouldn't close #408.

Closing this in favor of #425. Appreciate the careful PR writeup.

PR #422, opened 2026-04-23
fix(persistent-sdk-sessions): gate onTurnAborted on pending.size === 0 (scenario V)

cartwmic:
## Summary

- Gate `onTurnAborted` firing on `pending.size === 0` inside `consumeTurn`'s finally block so legitimate parallel-tool passthrough pauses don't spuriously drop the persistent runtime.
- Scenario T (mid-text client abort) behavior preserved: still drops cleanly when no tools are pending.
- Regression test covers the new gate.

## Root cause

Commit `83d265c` added `onTurnAborted` to handle scenario T (mid-stream client abort), but the callback couldn't distinguish two different exits-without-terminator paths in `consumeTurn`'s finally block:

- **Legitimate pause** — `turnRunner`'s `message_stop + pendingCount > 0` early-exit for parallel-tool passthrough. The synthetic `result` event is yielded BY turnRunner (not by consumeTurn), so `.return()` cascades to consumeTurn which exits with `terminated=false`.
- **Client abort** — TCP close during text generation. Same exit path, but no tools pending.

Both fired `onTurnAborted` → `manager.drop()`. As a result, every parallel-tool-use turn silently dropped the persistent runtime. The cascade of downstream failures:

- Runtime dropped → next turn's tool_results cold-reattach → pending registry empty → all tool_results classified as `prebind` → nothing pushed to SDK → deadlock.
- Subsequent follow-up HTTP request waits on the held mutex → 30s timeout → SSE `timeout_error` → pi retries 3× → gives up.

Visible symptom was the scenario V cascade in session `2026-04-23T02-02-04-302Z_019db812-aa8e-7628-9aaf-3744bcfa01e7.jsonl`. Hidden regression was cache eviction on every tool-using turn.

## Fix

```ts
finally {
  if (!terminated && !closed && pending.size === 0) {
    try { init.onTurnAborted?.() } catch { /* swallow */ }
  }
}
```

With handlers pending, treat as benign pause (runtime stays warm, the client's next turn resolves pending handlers / drains prebound buffer). Without handlers, still treat as client abort (Scenario T semantics preserved).

## Test plan

- [x] Unit: new regression test `does NOT fire onTurnAborted when pending handlers exist at break (scenario V)` covers the pending-gate discriminator.
- [x] Unit: existing 3 onTurnAborted tests still pass (terminator-reached, closed-runtime, mid-turn-abort).
- [x] Full suite: 1410/1410 pass, typecheck clean.
- [x] Live playbook regression pass via pi TUI + tmux:
  - **A** smoke — "OK" returned
  - **B** multi-hop cache — turn 2 `cache=94%`
  - **C/D** single-tool / multi-hop — cache climbs 21→84→98%
  - **E/F** single-turn / multi-hop parallel tools — ACK_BOTH, PAIR_A/B_DONE, **no spurious `persistent.turn_aborted` on pause**
  - **G** mixed plain/tool/plain/tool — all correct
  - **H** cold-reattach after proxy hard-kill — cache=96%
  - **I** setModel in-place — optionsHash unchanged, no reopen
  - **J** systemPrompt drift — optionsHash changed, close+reopen lifecycle fired
  - **L** LRU idle eviction — 9 evicts, all >20s since last activity
  - **N** SIGTERM graceful shutdown — runtimes closed, process exited, in-flight request surfaced error
  - **O** opus+thinking=high+parallel — O_DONE, no spurious aborts
  - **P/Q** steer-during-tool + queued follow-up — aggregation (bf39203) still works, model replied BYE
  - **T** mid-text abort — ESC fires `persistent.turn_aborted` (pending=0 path), follow-up recovers at `cache=95%`
  - **U** pi --fork cold-start seed — full 14-msg context recapped
  - **V** mid-synthesis post-tool abort — cascade gone, follow-up cache=98%

Scenarios K (pi undo/fork) shares code path with U. M (mutex 429), R (pending timeout), S (subagent cascade) are covered by unit tests or need harness/fixtures outside this PR's blast radius.


cartwmic:
Closing — merging locally to fork instead.

Issue #419, opened 2026-04-22
/v1/models doesn't list claude-opus-4-7


xAlcahest:
The /v1/models endpoint currently returns three models: claude-sonnet-4-6, claude-opus-4-6, and claude-haiku-4-5-20251001.

claude-opus-4-7 isn't in the list, but it works fine when you call it directly through /v1/messages. Tools like OpenCode use the models list to populate their model picker, so 4.7 doesn't show up in the UI even though it's perfectly usable.

Quick repro:

```bash
# only returns 3 models, no claude-opus-4-7
curl http://127.0.0.1:3456/v1/models

# but this works fine
curl http://127.0.0.1:3456/v1/messages \
  -H "Content-Type: application/json" \
  -H "anthropic-version: 2023-06-01" \
  -d '{"model":"claude-opus-4-7","max_tokens":10,"messages":[{"role":"user","content":"hi"}]}'
```

## Environment

- Meridian: 1.38.0
- opencode-with-claude: 1.5.6

xAlcahest:
Dug into this a bit more and found a bigger problem than the missing model listing.

Regardless of which model you request, Meridian always routes to claude-sonnet-4-20250514. Tested by asking the model to self-identify:

```
claude-opus-4-20250514    -> self-identifies as: claude-sonnet-4-20250514
claude-opus-4-1-20250805  -> self-identifies as: claude-sonnet-4-20250514
claude-opus-4-5-20251101  -> self-identifies as: claude-sonnet-4-20250514
claude-opus-4-6           -> self-identifies as: claude-sonnet-4-20250514
claude-opus-4-7           -> self-identifies as: claude-sonnet-4-20250514
```

The response `model` field just echoes back whatever you sent, but the actual model answering is always Sonnet 4. Token usage is also always reported as zero:

```python
import requests
r = requests.post('http://127.0.0.1:3456/v1/messages',
    headers={'Content-Type':'application/json','anthropic-version':'2023-06-01'},
    json={'model':'claude-opus-4-7','max_tokens':50,
          'messages':[{'role':'user','content':'What exact model are you? Reply with just your model ID.'}]})
print(r.json())
# {'model': 'claude-opus-4-7', ..., 'content': [{'text': 'claude-sonnet-4-20250514'}], 'usage': {'input_tokens': 0, 'output_tokens': 0}}
```

Running with `MERIDIAN_PASSTHROUGH=true` (default set by opencode-with-claude plugin). Meridian 1.38.0.

[...]

PR #430, opened 2026-04-24
fix(models): pin SDK model aliases via ANTHROPIC_DEFAULT_*_MODEL env

rynfar:
Main fix for #419.

## Summary

The deeper symptom in #419 — every requested model (opus-4-5 through opus-4-7) silently self-identifying as \`claude-sonnet-4-*\` — stems from stale SDK alias resolution.

\`mapModelToClaudeModel\` in \`src/proxy/models.ts\` collapses every requested model string into one of three aliases: \`sonnet\` / \`opus\` / \`haiku\`. These aliases are passed to the Claude Agent SDK, which resolves them to concrete versions via \`ANTHROPIC_DEFAULT_{TYPE}_MODEL\` env vars. When those env vars are unset, the SDK falls back to its own bundled defaults — which lag real Claude Max availability, so fresh installs route to outdated model pins.

@fbraza confirmed in #419 that setting these env vars by hand in \`.bashrc\` fixed the routing. This PR makes Meridian pin them automatically at the SDK subprocess boundary.

## Design

New \`resolveSdkModelDefaults()\` in \`models.ts\` emits canonical pins for all three aliases:

| Alias | Meridian-pinned default | Constant |
|---|---|---|
| opus | \`claude-opus-4-7\` | \`CANONICAL_OPUS_MODEL\` |
| sonnet | \`claude-sonnet-4-6\` | \`CANONICAL_SONNET_MODEL\` |
| haiku | \`claude-haiku-4-5\` | \`CANONICAL_HAIKU_MODEL\` |

\`server.ts\` spreads the defaults into \`profileEnv\` **before** the inherited \`cleanEnv\`:

\`\`\`ts
const profileEnv = { ...sdkModelDefaults, ...cleanEnv, ...profile.env }
\`\`\`

So the precedence is:
1. **Shell \`ANTHROPIC_DEFAULT_*_MODEL\`** (in \`cleanEnv\`) — always wins; power users can pin manually
2. **\`MERIDIAN_DEFAULT_*_MODEL\`** (proxy-side override) — wins over the canonical default
3. **Canonical pins** — Meridian's baked-in fallback

Also logs the active pins at startup so operators can verify without having to ask the model to self-identify:

\`\`\`
Model pins: opus=claude-opus-4-7 sonnet=claude-sonnet-4-6 haiku=claude-haiku-4-5
\`\`\`

## Scope note

This fix is **necessary but not sufficient** on setups using the bundled SDK cli.js (pre-0.2.117). That bundled cli.js is too old to recognize \`claude-opus-4-7\` regardless of env — it ignores the pin and falls through to its own hardcoded default. The full remedy requires upgrading the SDK to 0.2.117 and switching to the standalone \`@anthropic-ai/claude-code\` binary, which will land as a separate PR.

For all other setups (node-run Meridian using the system \`claude\` binary, which is usually current) this PR is the complete fix.

## Test plan

- [x] Unit tests in \`models.test.ts\` for \`resolveSdkModelDefaults\`:
  - Canonical pins returned when no overrides set
  - \`MERIDIAN_DEFAULT_{TYPE}_MODEL\` overrides win over canonicals
  - Returns only the three \`ANTHROPIC_DEFAULT_*\` keys
- [x] Integration tests in \`proxy-env-stripping.test.ts\` verifying \`capturedQueryOptions.env\`:
  - Canonical pins injected when no shell env set
  - Shell \`ANTHROPIC_DEFAULT_*_MODEL\` values win over Meridian's pins
  - \`MERIDIAN_DEFAULT_*_MODEL\` overrides canonical but not shell
- [x] \`bun test\` — **1444/1444 pass** (7 new)
- [x] **E2E:** launchd proxy respawned on this branch logs the pins correctly at startup:
  \`\`\`
  Model pins: opus=claude-opus-4-7 sonnet=claude-sonnet-4-6 haiku=claude-haiku-4-5
  \`\`\`
- [x] **E2E caveat:** Trevor's launchd runs under bun → uses the bundled cli.js (stale), so live self-identification for opus-4-7 still hits the stale-cli issue. Full self-identification verification blocked on SDK upgrade PR (PR C).

Supersedes fbraza's \`.bashrc\` workaround in #419. Preserves every escape hatch for users who want different pins.

### DIFF
--- a/src/__tests__/models.test.ts
+++ b/src/__tests__/models.test.ts
@@ -3,7 +3,7 @@
  */
 import { afterEach, beforeEach, describe, it, expect, mock } from "bun:test"
 
-import { mapModelToClaudeModel, isClosedControllerError, resetCachedClaudeAuthStatus, stripExtendedContext, hasExtendedContext, recordExtendedContextUnavailable, isExtendedContextKnownUnavailable, resetExtendedContextUnavailable } from "../proxy/models"
+import { mapModelToClaudeModel, isClosedControllerError, resetCachedClaudeAuthStatus, stripExtendedContext, hasExtendedContext, recordExtendedContextUnavailable, isExtendedContextKnownUnavailable, resetExtendedContextUnavailable, resolveSdkModelDefaults, CANONICAL_OPUS_MODEL, CANONICAL_SONNET_MODEL, CANONICAL_HAIKU_MODEL } from "../proxy/models"
 
 describe("mapModelToClaudeModel", () => {
   const originalSonnetModel = process.env.CLAUDE_PROXY_SONNET_MODEL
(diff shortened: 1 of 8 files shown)


fbraza:
Cheers man many thanks for your work :-)

PR #442, opened 2026-04-27
fix(build): add postbuild script to fix Bun bundler export artifacts

Alexander-Ollman:
## Summary

- Adds `scripts/fix-bun-exports.mjs` that runs automatically in `postbuild` to clean up two known Bun bundler artifacts when using `--splitting`:
  - **Duplicate `export {}` blocks** — Bun emits a second export statement with a subset of symbols, causing `SyntaxError: Duplicate export of '<symbol>'` at runtime
  - **`__INVALID__REF__` references** — dead imports/exports referencing a nonexistent symbol

- The script finds all `.js` files in `dist/`, detects duplicate export blocks whose symbols are subsets of the canonical (first) export, removes them, and strips any `__INVALID__REF__` artifacts

## Test plan

- [x] `npm run build` succeeds end-to-end (bun build → fix script → node --check)
- [x] Verified no `__INVALID__REF__` or duplicate exports remain in dist/ after build
- [ ] `npm test` passes (CI will confirm)


rynfar:
im sorry for the delay on this i will get this resolve dtoday

rynfar:
Thanks for catching this @Alexander-Ollman — verified the bug is actually firing on a code-split chunk in this repo's `dist/` (the postbuild script reports `fixed: cli-<hash>.js — patched 1 file(s)` on a clean build), so this is a live fix, not just preventive maintenance.

Carried this forward in #456 with your original commit preserved (so you keep authorship in `git log` on `main`), plus a fixup commit that adds:

- Refactored `scripts/fix-bun-exports.mjs` to expose `patchSource()` and `fixBunExports()` for testability.
- A small formatting fix in the canonical-block cleaner — the chained-regex approach was swallowing the newline before `}` and producing output like `realThing};`.
- Regression tests in `src/__tests__/fix-bun-exports.test.ts` covering the pure helpers and an end-to-end fixture where a file fails to import BEFORE the script runs and loads cleanly AFTER.

Merged with `--merge` (not squash) so both commits land in `main` history. Your contribution is now on `main` — thanks again for the fix!

PR #420, opened 2026-04-23
fix(proxy): allow 4 turns when passthrough has both resume and deferred tools

WarGloom:
## Problem

In passthrough mode, `buildQueryOptions` caps `maxTurns` with:

```ts
maxTurns: passthrough ? ((resumeSessionId || hasDeferredTools) ? 3 : 2) : 200
```

Either single extension (resume *or* deferred tools) gets 3 turns; plain passthrough gets 2. But when **both** are active simultaneously, the SDK walks four phases:

| Phase | Cost |
|---|---|
| Session rehydration (resume) | 1 turn |
| ToolSearch lookup (deferred tools) | 1 turn |
| Real tool call | 1 turn |
| Blocked-tool handoff | 1 turn |

The request blows past the ternary's 3-turn ceiling and `claude-agent-sdk` throws `Reached maximum number of turns (3)` from `QueryEngine`, which surfaces to callers as:

```json
{"type":"api_error","message":"Claude Code returned an error result: Reached maximum number of turns (3)"}
```

I hit this repeatedly running a resumed session with deferred MCP tools registered (the common setup when using `opencode-with-claude` + a plugin that surfaces deferred tools via `ToolSearch`).

## Fix

Raise the ceiling to 4 **only** when both flags are true. Single-extension (3) and plain passthrough (2) budgets are unchanged.

```ts
maxTurns: passthrough
  ? (resumeSessionId && hasDeferredTools
      ? 4
      : (resumeSessionId || hasDeferredTools) ? 3 : 2)
  : 200
```

## Test coverage

Added a new case to `src/__tests__/query.test.ts`:

```ts
it("sets maxTurns to 4 in passthrough mode when resume AND deferred tools are both active", ...)
```

Existing tests for `maxTurns = 2 | 3 | 200` still pass unchanged. The existing test matrix (`proxy-droid-integration`, `proxy-transparent-tools`, `proxy-passthrough-tool-use`, `proxy-passthrough-concept`, `proxy-passthrough-thinking`, `deferred-tool-loading`) was missing the both-true case, which is why the regression slipped through.

## Verification

- `bun test src/__tests__/query.test.ts` — 40/40 pass
- `bun test` on related suites (`deferred-tool-loading`, `proxy-transparent-tools`, `proxy-passthrough-tool-use`, `proxy-droid-integration`, `proxy-passthrough-concept`, `proxy-passthrough-thinking`) — 62/62 pass
[...]

PR #447, opened 2026-04-29
feat(agentDefs): fall back to input_schema enum for Task subagent extraction

WarGloom:
## Summary

Adds a JSON Schema-based fallback to `agentDefs` so Meridian can extract subagent definitions from a Task tool when the textual `description` does not match the existing `Available agent types:` regex.

## Motivation

`parseAgentDescriptions` parses the Task tool's `description` field with this regex:

```
/Available agent types.*?:\n((?:- [\w][\w-]*:.*\n?)+)/s
```

If a client renders the Task tool with a different description style but still declares `subagent_type` as a string field with an `enum` in `input_schema`, Meridian receives no agents. The Claude Agent SDK then falls back to its built-in subagents (`dev`, `explore`), and any user-defined subagents are silently dropped from routing.

JSON Schema's `enum` is the canonical, language-agnostic place to declare a closed set of string values for a tool argument. Reading from there makes Meridian work with any Anthropic-compatible client that follows the JSON Schema convention, regardless of how the description text is phrased.

## Changes

- New `parseAgentNamesFromSchema(taskTool)` helper that reads `taskTool.input_schema.properties.subagent_type.enum`.
- New `buildAgentDefinitionsFromTool(taskTool, mcpToolNames?)` entry point:
  - Tries description-based parsing first (existing behavior, fully preserved).
  - Falls back to schema enum only when description parsing yields zero agents.
  - Returns the same `Record<string, AgentDefinition>` shape, so PreToolUse hooks, system context addendum, default injection, and case variants all keep working unchanged.
- `adapters/opencode.ts` and `transforms/opencode.ts` switch their internal call to `buildAgentDefinitionsFromTool`.
- Original `buildAgentDefinitions(taskDescription, mcpToolNames?)` is kept and unchanged.
- New helper signatures use `unknown` (not `any`) with nested narrowing, so callers passing arbitrary JSON cannot crash the parser.

## Behavior matrix

| Description regex match | Schema enum present | Result |
|---|---|---|
| match | – | description-derived agents (unchanged) |
| miss | present | enum-derived agents + defaults + variants |
| miss | absent | empty (unchanged; SDK builtins kick in) |

Description always wins when present, so existing OpenCode integrations are not affected.

## Testing

```bash
bun test src/__tests__/proxy-agent-definitions.test.ts
```

10 new test cases:

- `parseAgentNamesFromSchema`: happy path, missing enum, missing schema, non-string enum entries, primitive inputs, wrong-shape inputs.
- `buildAgentDefinitionsFromTool`: enum fallback, description precedence, empty inputs, MCP tool propagation, prompt/model/case-variant correctness.

All pre-existing tests in this file still pass (40/40).

## Token impact

Schema enum strings live in a cached prompt prefix (~3-5 tokens per agent name), so the cost amortizes to near-zero across a session.

## Notes

- No breaking changes.
- No new env vars or feature flags.
- Other adapters (`crush`, `droid`, `forgecode`, `pi`, `passthrough`, `claude-code`) are untouched.

### DIFF
--- a/src/__tests__/proxy-agent-definitions.test.ts
+++ b/src/__tests__/proxy-agent-definitions.test.ts
@@ -6,7 +6,14 @@
  */
 
 import { describe, it, expect } from "bun:test"
-import { parseAgentDescriptions, buildAgentDefinitions, mapModelTier, FALLBACK_AGENT_NAME } from "../proxy/agentDefs"
+import {
+  parseAgentDescriptions,
+  buildAgentDefinitions,
+  mapModelTier,
+  FALLBACK_AGENT_NAME,
+  parseAgentNamesFromSchema,
+  buildAgentDefinitionsFromTool,
+} from "../proxy/agentDefs"
 
 const SAMPLE_TASK_DESCRIPTION = `Launch a new agent to handle complex, multistep tasks autonomously.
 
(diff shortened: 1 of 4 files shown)


rynfar:
thanks for this, im reviewing this now im sorry for the delay in getting to this, ive been tied up!

rynfar:
Thanks @WarGloom — exactly the right design instinct. JSON Schema's `enum` IS the canonical place to declare a closed set of string values for a tool argument, and relying on prose parsing in the description was fragile. Verified the fallback works end-to-end:

- Cherry-picked your commit onto current `main` (4 PRs landed since you opened this) — cleanly, no conflicts.
- Live-tested all 5 input shapes (OpenCode-style description, Omo-style enum-only, conflicting both, neither, and defensive parsing) — every case behaves per your behavior matrix.
- Full suite: 1557 pass / 0 fail, typecheck clean.

Carried forward in #457 with `--merge` so your commit lands on `main` with authorship intact. Also seen #420 from you on the maxTurns gap — same instinct was correct there too; we ended up at the same answer in #454 yesterday. Thanks again!

PR #448, opened 2026-04-29
fix(agentDefs): clone definitions when registering case variants and aliases

WarGloom:
## Summary

`addCaseVariants` (in `src/proxy/agentDefs.ts`) registers PascalCase variants and the `general-purpose` / `General-Purpose` aliases by **assigning the source `AgentDefinition` by reference**. Mutating any of those entries (e.g. pushing onto `tools`) silently mutates the lowercase base, and vice versa.

## Reproducing

```ts
const defs = buildAgentDefinitions(SAMPLE_DESCRIPTION, ["mcp__opencode__read"])
defs["Explore"].tools.push("mcp__opencode__poison")
console.log(defs["explore"].tools)
// → ["mcp__opencode__read", "mcp__opencode__poison"]
```

Today no production code path mutates these values, so the bug is latent. But:

- The SDK or future consumers may legitimately tweak per-variant `tools`, `model`, `prompt`, or `disallowedTools`, and would corrupt the base unexpectedly.
- It is the kind of footgun that surfaces months later and is hard to reproduce.

## Change

- New tiny helper `cloneAgentDefinition(def)` — shallow clone with `tools` / `disallowedTools` arrays duplicated.
- Variants and aliases now use `cloneAgentDefinition(...)` instead of direct assignment.
- No behavior change for current call sites: shape and contents of returned record stay identical.

## Tests

Two regression tests in `proxy-agent-definitions.test.ts`:

1. `defs["Explore"].tools.push(...)` does not leak to `defs["explore"]`.
2. `defs["general-purpose"].tools.push(...)` does not leak to `defs["general"]`.

```bash
bun test src/__tests__/proxy-agent-definitions.test.ts
# 30 pass, 0 fail
```

## Notes

- No public API change.
- No new env vars or feature flags.
- Pure defensive fix.

### DIFF
--- a/src/__tests__/proxy-agent-definitions.test.ts
+++ b/src/__tests__/proxy-agent-definitions.test.ts
@@ -319,4 +319,18 @@ describe("Case variant registration", () => {
     const defs = buildAgentDefinitions("No agents here")
     expect(Object.keys(defs)).toHaveLength(0)
   })
+
+  it("PascalCase variant must not share mutable references with base", () => {
+    const defs = buildAgentDefinitions(SAMPLE_TASK_DESCRIPTION, ["mcp__opencode__read"])
+    expect(defs["Explore"]).not.toBe(defs["explore"])
+    defs["Explore"]!.tools!.push("mcp__opencode__poison")
+    expect(defs["explore"]!.tools).not.toContain("mcp__opencode__poison")
+  })
+
+  it("'general-purpose' alias must not share mutable references with target", () => {
+    const defs = buildAgentDefinitions(SAMPLE_TASK_DESCRIPTION, ["mcp__opencode__read"])
+    expect(defs["general-purpose"]).not.toBe(defs["general"])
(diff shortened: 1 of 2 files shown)


rynfar:
great work looking at this one now

rynfar:
Thanks @WarGloom — caught a real footgun before it bit us. Verified:

- Cherry-picked your commit onto current `main` (5 PRs landed since you opened this, including #457 which also touched `agentDefs.ts`) — auto-merged cleanly, no conflicts.
- Full suite: 1559 pass / 0 fail; typecheck clean.
- Your regression tests double as the runtime smoke test (mutate `defs["Explore"].tools`, verify `defs["explore"].tools` unchanged) — exactly the right shape.

Carried forward in #458 with `--merge` so your commit lands on `main` with authorship intact. That's two of yours merged today (+ #457). Thanks again for the careful work!

PR #451, opened 2026-04-29
feat(proxy): Tool call and thinking support for OpenAI proxy

Lumpiasty:
## Summary

Adds support for tool calling via OpenAI request translator, both non-streaming and streaming. Allowing for use cases like agentic web search in Open WebUI.

`/v1/chat/completions` endpoint gained support for `tools` field with `type: "function"`, `role: "tool"` messages and `tool_calls` field in assistant's message. Likewise, response `thinking`, and `tool_use` blocks are now translated to OpenAI responses.

Conversation history is still summarized into system prompt, I didn't change anything there besides adding new block types. Current way it's done invalidates any prompt caching and has side effects on model's responses. I think proper way would be to get rid of this "summarizing" and maintain a claude code session if prefix matches (if I understand the issue from code comment correctly, correct me if I'm wrong), but that's out of scope, my goal was making it work at all. I think I can start working on that if this gets approved.

## Demo
[Nagranie ekranu_20260429_170317.webm](https://github.com/user-attachments/assets/619d6fbf-ddce-457d-95cd-07999010461a)

### DIFF
--- a/src/__tests__/openai.test.ts
+++ b/src/__tests__/openai.test.ts
@@ -9,6 +9,7 @@ import {
   translateOpenAiToAnthropic,
   translateAnthropicToOpenAi,
   translateAnthropicSseEvent,
+  createSseTranslator,
   buildModelList,
 } from "../proxy/openai"
 
@@ -58,7 +59,7 @@ describe("translateOpenAiToAnthropic", () => {
       messages: [{ role: "user", content: "Hello" }],
     })
     expect(result).not.toBeNull()
-    expect(result!.messages).toEqual([{ role: "user", content: "Hello" }])
+    expect(result!.messages).toEqual([{ role: "user", content: [{ type: "text", text: "Hello" }] }])
     expect(result!.system).toBeUndefined()
   })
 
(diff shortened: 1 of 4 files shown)


rynfar:
thanks for submitting this I will review it now.  Sorry for the delay!  

Thank you for the high quality PR.  You clearly took care in submitting this.

rynfar:
Thanks @Lumpiasty — fantastic contribution. Brings the OpenAI translator path to feature parity with the Anthropic passthrough on tool calling and thinking. Verified end-to-end:

- Cherry-picked both your commits onto current `main` (6 PRs landed since you opened this, including SDK termination diagnostics that also touched `server.ts`) — auto-merged cleanly, no conflicts.
- Live-tested all 5 translator scenarios: tool-def translation, tool_use → tool_calls, thinking → reasoning_content (passthrough on/off), and streaming with multiple tools (sequential index advancement).
- Full suite: 1610 pass / 0 fail; typecheck clean.

The `createSseTranslator` factory pattern for per-stream tool-call index counter is exactly the right design — encapsulates the state cleanly without leaking into server.ts.

The discriminated union upgrade for `AnthropicContentBlock` is a nice improvement on the type safety side too.

Carried forward in #459 with `--merge` so both your commits land on `main` with authorship intact. Re: the conversation-history summarization caveat you flagged — agreed that's the right separate follow-up. If you want to take that on, would be happy to review.

Thanks again!

Issue #445, opened 2026-04-28
Latest version fails to install


murrayju:
I'm currently on `1.38.0`, and have been through several installs/updates already via npm. The current version (`1.40.0`) fails to install this way:

```bash
❯ npm install -g @rynfar/meridian
npm error code 1
npm error path /opt/homebrew/lib/node_modules/@rynfar/meridian/node_modules/@anthropic-ai/claude-code
npm error command failed
npm error command sh -c node install.cjs
npm error [@anthropic-ai/claude-code postinstall] Failed to place binary: ENOENT: no such file or directory, open '/opt/homebrew/lib/node_modules/@rynfar/meridian/node_modules/@anthropic-ai/claude-code/bin/claude.exe'
npm error   Fallback: node /opt/homebrew/lib/node_modules/@rynfar/meridian/node_modules/@anthropic-ai/claude-code/cli-wrapper.cjs
npm error A complete log of this run can be found in: /Users/murrayju/.npm/_logs/2026-04-28T17_10_13_444Z-debug-0.log
```

Note that I'm running on MacOS, so the mention of `claude.exe` in the error message is suspicious.

ianjwhite99:
I believe this issue stems from #417

rynfar:
Sorry for the delay im playing catch up on issues.  I will try to get to this today.

rynfar:
Partial fix landed in #463 (just merged to \`main\`). What it changes:

- Even when `@anthropic-ai/claude-code`'s install.cjs fails (the upstream ENOENT bug you hit), meridian's resolver now **skips the broken stub** and falls back to the platform-specific peer package directly. So once meridian is installed, the proxy will boot and work correctly.

What it does **not** change:
- The actual upstream install.cjs failure is in `@anthropic-ai/claude-code`'s postinstall, not meridian. Until that's fixed upstream, you may still see the ENOENT error during install.

**Workaround for now**: try installing with `--ignore-scripts` to bypass all postinstalls, then meridian's runtime resolver will find the binary on its own:

\`\`\`bash
npm install -g @rynfar/meridian --ignore-scripts
\`\`\`

@murrayju — once the next release is out, would you mind verifying:

1. Does `npm install -g @rynfar/meridian --ignore-scripts` succeed for you on macOS + homebrew node?
2. After install, does `meridian` boot and `POST /v1/messages` return a real completion?

If yes to both, we have a working path for the affected install scenario. If you want to test before the release, you can install from the merged commit:

[...]

PR #463, opened 2026-05-01
fix(resolver): Windows + broken-postinstall fallbacks (closes #417, mitigates #445)

### LINKED ISSUE #417: Windows: 'Could not find Claude Code executable' on startup (resolver incompatible with claude-agent-sdk 0.2.x)
## Summary

`meridian` fails to start on Windows with:

```
[PROXY] Uncaught exception (recovered): Could not find Claude Code executable. Install via: npm install -g @anthropic-ai/claude-code
```

The process never binds port 3456. Reproduced on clean install of `@rynfar/meridian@1.38.0` + Node 23.9.0 + Windows 10, with `@anthropic-ai/claude-code@2.1.117` already installed globally and `claude login` completed.

## Root cause

`resolveClaudeExecutableAsync` in `dist/cli-*.js` tries two branches on non-Bun:

1. `exec("which claude")` via Node `child_process.exec` — on Windows this invokes `cmd.exe`, which resolves `which.exe` from Git for Windows' `usr/bin`. That `which` emits mingw-style paths like `/c/nvm4w/nodejs/claude`, which `fs.existsSync` rejects. The `try/catch` swallows it and falls through.
2. Fallback looks for `cli.js` next to `@anthropic-ai/claude-agent-sdk`'s main entry. Current SDK (`0.2.117`, bundled by meridian) no longer ships `cli.js` — the native binary lives in the platform-specific peer packages (e.g. `@anthropic-ai/claude-agent-sdk-win32-x64/claude.exe`). Fallback fails, resolver throws.

## Fix sketch

Minimal patch that works for me:

- Honor a `MERIDIAN_CLAUDE_PATH` env var as an explicit escape hatch at the top of the resolver.
- On `process.platform === 'win32'`, use `where claude` instead of `which claude` and pick the first existing line from multi-line stdout.
- Add a last-resort fallback: look for `claude.exe` (or `claude` on POSIX) in `@anthropic-ai/claude-agent-sdk-<platform>-<arch>` relative to the resolved SDK package.

Diff-style:

```js
cachedClaudePathPromise = (async () => {
  if (process.env.MERIDIAN_CLAUDE_PATH && existsSync(process.env.MERIDIAN_CLAUDE_PATH)) {
    cachedClaudePath = process.env.MERIDIAN_CLAUDE_PATH;
    return cachedClaudePath;
  }
  // ...existing bun cli.js lookup...
  try {
    const whichCmd = process.platform === "win32" ? "where claude" : "which claude";
    const { stdout } = await exec(whichCmd);
    const claudePath = stdout.split(/\r?\n/).map(s => s.trim()).find(s => s && existsSync(s));
    if (claudePath) { cachedClaudePath = claudePath; return claudePath; }
  } catch {}
  if (!runningUnderBun) {
    try {
      const sdkPath = fileURLToPath(import.meta.resolve("@anthropic-ai/claude-agent-sdk"));
      const sdkDir = dirname(sdkPath);
      const cliJs = join(sdkDir, "cli.js");
      if (existsSync(cliJs)) { cachedClaudePath = cliJs; return cliJs; }
      const exe = join(sdkDir, "..", `claude-agent-sdk-${process.platform}-${process.arch}`, process.platform === "win32" ? "claude.exe" : "claude");
      if (existsSync(exe)) { cachedClaudePath = exe; return exe; }
    } catch {}
  }
  throw new Error("Could not find Claude Code executable...");
})();
```

With this patch applied locally, `meridian` boots and `POST /v1/messages` returns a real completion from the user's Max subscription.

## Repro

1. Windows 10, Node 23.9.0, Git for Windows installed (adds `usr/bin/which.exe` to PATH).
2. `npm i -g @rynfar/meridian @anthropic-ai/claude-code`
3. `claude login`
4. `meridian` → exception above, no listener on 3456.

## Environment

- meridian: 1.38.0
- @anthropic-ai/claude-agent-sdk (bundled in meridian): 0.2.117
- @anthropic-ai/claude-code: 2.1.117
- Node: 23.9.0
- OS: Windows 10 19045

rynfar:
Fixes the Claude executable resolver to handle two unrelated failure modes that together leave Windows users (#417) and some macOS users (#445) unable to start the proxy.

## Root causes

**Issue #417** — Windows resolver bugs (analysis credit: @Macawls):
- `exec("which claude")` on Windows invokes Git for Windows' `which.exe`, which emits mingw-style paths (`/c/nvm4w/nodejs/claude`) that Node's `existsSync` rejects.
- No `MERIDIAN_CLAUDE_PATH` env escape hatch for users with unusual setups.
- @BenIsLegit's follow-up: even when the bundled binary exists, `bin/claude.exe` can fail with `spawn UNKNOWN` on Windows; the platform-specific peer package binary works.

**Issue #445** — broken upstream postinstall:
- `@anthropic-ai/claude-code`'s `install.cjs` can fail with ENOENT (e.g. on homebrew npm setups).
- When that fails, `bin/claude.exe` stays as a ~500-byte placeholder stub instead of being replaced with the ~200 MB real binary.
- The old resolver picked up the stub and handed it to the SDK, which then failed at runtime with a cryptic spawn error.

## Fix

Refactored `resolveClaudeExecutableAsync` into a pure `resolveClaudeExecutable(deps)` function with all I/O dependencies injected. Production caller `resolveClaudeExecutableAsync` wraps it with caching (unchanged contract). Tests inject mock deps to simulate Windows / broken-postinstall / etc. without touching the real filesystem.

New resolution order, each step pure and falling through on miss:

| # | Source | Solves |
|---|---|---|
| 0 | `MERIDIAN_CLAUDE_PATH` env var | Universal escape hatch (Macawls' suggestion) |
| 1 | `@anthropic-ai/claude-code/bin/claude.exe` **with stub-size guard** (skip files ≤ 4 KB) | #445 — broken postinstall stub no longer poisons the resolver |
| 2 | `@anthropic-ai/claude-code-<platform>-<arch>/claude[.exe]` (and `-musl` on Linux) | Bypasses bundled wrapper entirely; works for #445 (no postinstall needed) AND #417 BenIsLegit case (wrapper fails but platform binary works) |
| 3 | `where claude` on Windows / `which claude` on POSIX. Multi-line stdout parsed; mingw `/c/...` paths filtered out as defense-in-depth. | #417 main complaint |
| 4 | Legacy SDK cli.js (bun only, SDK < 0.2.98) | Existing fallback, kept |

## Tests

26 new unit tests in `claude-executable-resolver.test.ts` covering:
- Env override priority + presence/absence/empty handling
- Stub-size guard at the 4 KB boundary
- Each platform package variant (darwin-arm64, win32-x64, win32-arm64, linux-x64-musl fallback)
- `where` vs `which` command selection
- Multi-line stdout parsing + first-existing-path selection
- Mingw `/c/...` path filtering on Windows
- Bun-only legacy cli.js gate
- Full priority chain (env beats bundled, bundled real beats platform, platform beats PATH when bundled is stub)

**Result:** 1643 tests pass / 0 fail / typecheck clean.

## Windows CI smoke test

New `windows-smoke` job in `.github/workflows/ci.yml` that on `windows-latest`:
1. `bun install` — exercises claude-code's postinstall on real Windows
2. Runs the resolver tests on the real Windows runtime (path joining, file URLs, etc.)
3. Builds the bundle
4. Spawns `node dist/cli.js`, polls `/health` for 30s, verifies a 200 response with a `version` field

This is the real backstop — it catches Windows-specific regressions in the install + start flow that unit tests with mocks can't see, and runs on every PR.

## Honest verification disclosure

I'm developing on macOS arm64. The unit tests exhaustively cover the resolver's branching logic with mocked deps, but **I can't run a true Windows e2e test from this machine.** The Windows CI smoke job above will do that on this PR. **Please don't merge until that job passes** AND ideally one of the original reporters confirms it works on their setup.

Pinging the reporters who diagnosed this:
- @Macawls — does this match your earlier patch and unblock the install on Windows 10 + nvm4w + Git for Windows?
- @BenIsLegit — does the platform-package fallback resolve the `spawn UNKNOWN` you hit?
- @murrayju — once installed (with `--ignore-scripts` if needed), does the proxy boot for you on macOS? Note the upstream `claude-code` postinstall ENOENT is not directly fixable from meridian, but with this PR a broken postinstall no longer breaks startup.

## Test plan

- [x] `npm test` — 1643 pass / 0 fail
- [x] `npx tsc --noEmit` — clean
- [x] macOS regression: `resolveClaudeExecutable()` returns the real bundled binary as before
- [ ] Windows CI smoke (will fire on push)
- [ ] Real-world verification by reporters before merge

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,3 +24,74 @@ jobs:
       - run: npm run typecheck
       - run: npm test
       - run: bun run build
+
+  # Windows smoke test — catches platform-specific regressions in:
+  #   - claude-code postinstall on Windows (issue #445 class)
+  #   - the resolver picking up the bundled binary or platform-package binary (#417)
+  #   - cross-platform path handling in TS code
+  #   - bundle output spawning under node on Windows
+  # We don't run the full unit suite here — that's covered by the Linux job
+  # above and the resolver's logic is mock-driven so platform doesn't matter
+  # for unit-level coverage. This job exercises the REAL install + start flow.
+  windows-smoke:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+
+      - uses: oven-sh/setup-bun@v2
+
+      # Install — this also runs claude-code's postinstall, which is the
+      # single most common source of Windows install-time bugs.
+      - run: bun install
+
+      # Run only the resolver tests on Windows. Useful even though they use
+      # mocked deps, because the test runner itself exercises path joining,
+      # file URL parsing, etc., on the real Windows runtime.
+      - run: bun test src/__tests__/claude-executable-resolver.test.ts
+        shell: bash
+
+      # Build the bundle.
+      - run: bun run build
+        shell: bash
+
+      # Smoke-start: launch the bundled CLI in the background, hit /health,
+      # verify it responds with a body. We don't have Claude auth in CI, so
+      # the auth check inside /health may report loggedIn:false — that's
+      # fine; what we're verifying is the resolver finds the executable
+      # and the proxy boots without throwing.
+      - name: smoke start + /health probe
+        shell: pwsh
+        run: |
+          $env:MERIDIAN_PORT = "3458"
+          $env:MERIDIAN_HOST = "127.0.0.1"
+          # Start meridian in the background.
+          $proc = Start-Process -FilePath "node" -ArgumentList "dist/cli.js" -PassThru -RedirectStandardOutput "meridian.out.log" -RedirectStandardError "meridian.err.log" -WindowStyle Hidden
+          try {
+            # Poll /health up to 30s.
+            $ok = $false
+            for ($i = 0; $i -lt 30; $i++) {
+              Start-Sleep -Seconds 1
+              try {
+                $r = Invoke-WebRequest -Uri "http://127.0.0.1:3458/health" -UseBasicParsing -TimeoutSec 2
+                if ($r.StatusCode -eq 200) { $ok = $true; break }
+              } catch { }
+            }
+            if (-not $ok) {
+              Write-Host "===== meridian.out.log ====="
+              if (Test-Path "meridian.out.log") { Get-Content "meridian.out.log" }
+              Write-Host "===== meridian.err.log ====="
+              if (Test-Path "meridian.err.log") { Get-Content "meridian.err.log" }
+              throw "Meridian did not respond on /health within 30s — resolver or startup likely failed on Windows"
+            }
+            $body = Invoke-RestMethod -Uri "http://127.0.0.1:3458/health" -TimeoutSec 5
+            Write-Host "/health response: $($body | ConvertTo-Json -Compress)"
+            if (-not $body.version) { throw "/health missing version field" }
+            Write-Host "OK — meridian booted on Windows and /health returned a version."
+          }
+          finally {
+            try { Stop-Process -Id $proc.Id -Force -ErrorAction SilentlyContinue } catch { }
+          }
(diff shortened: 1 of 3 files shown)


Macawls:
I don't mind testing again but it seems like its solved, lmk 

Thank you for solving this

Issue #453, opened 2026-04-30
sharedMemory sets CLAUDE_CONFIG_DIR and breaks Claude Code SDK OAuth lookup


johanes-dev:
Meridian sets `CLAUDE_CONFIG_DIR` for shared memory at:

https://github.com/rynfar/meridian/blob/main/src/proxy/query.ts#L249

## Current code:

```ts
...(sharedMemory ? { CLAUDE_CONFIG_DIR: join(homedir(), ".claude") } : {}),
```

This looks equivalent to leaving `CLAUDE_CONFIG_DIR` unset, but it is not. In Claude Code SDK query mode, explicitly setting `CLAUDE_CONFIG_DIR=/Users/<user>/.claude` changes the Claude Code Keychain lookup key/behavior and can make SDK auth fail.

## Observed behavior:
- `claude auth status` works.
- SDK `query()` works when `CLAUDE_CONFIG_DIR` is unset.
- SDK `query()` fails when `CLAUDE_CONFIG_DIR=$HOME/.claude` is explicitly set.
- Meridian then returns `Not logged in · Please run /login`.
- Removing this env override fixes Meridian.

## Related upstream Claude Code issue:
https://github.com/anthropics/claude-code/issues/20553

## Expected behavior:
Do not set `CLAUDE_CONFIG_DIR` when the intended target is the default `~/.claude`. Only set it for real custom profile/config directories.

rynfar:
thanks for reporting, putting together a fix now

rynfar:
Fix landed in #467 — just merged to \`main\`. Will go out in the next release.

Your diagnosis was on the money. Implemented the slightly broader version:

- **Default profile + sharedMemory:** env unset, SDK uses its own ~/.claude default (the case you reported)
- **Custom profile + sharedMemory:** custom value stripped, SDK falls back to default — actually achieves the "share memory with Claude Code" intent without the explicit-set bug
- **No sharedMemory:** unchanged in either case

Generalized as a tiny `stripConfigDir(env)` helper so the intent is explicit: when the user wants to share memory, we let the SDK pick its own default rather than poisoning Keychain auth by setting the path string ourselves. Pinned by 4 tests covering the full matrix.

Two of yours fixed today — #452 (compact JSON credentials) and now this one. The level of precision in your reports has been outstanding; both were diagnosed line-accurately and the fixes were small.

To verify before the next release:

\`\`\`bash
npm install -g github:rynfar/meridian#main
\`\`\`

[...]

PR #474, opened 2026-05-02
feat: register OAuth token as profile (closes #446)

### LINKED ISSUE #446: Feature: register a `claude setup-token` OAuth token as a profile (single-value, no config dir)
## Background

Anthropic ships `claude setup-token`, which mints a **long-lived OAuth token** (valid ~1 year) for headless / CI use of Claude Code. The Claude Code SDK already honours `CLAUDE_CODE_OAUTH_TOKEN` natively.

I'd like to register such a token as a Meridian profile **the same way `apiKey` works today** — just a single string value in `profiles.json`, no synthetic config directory, no `.credentials.json`, no browser flow.

## Current limitation

`ProfileConfig` (`src/proxy/profiles.ts`) supports only two shapes:

```ts
export interface ProfileConfig {
  id: string
  type?: "claude-max" | "api"
  claudeConfigDir?: string   // claude-max: path to a config dir
  apiKey?: string            // api: ANTHROPIC_API_KEY
  baseUrl?: string
}
```

So the only way to register a Max-subscription identity today is to point at an entire `CLAUDE_CONFIG_DIR`. There's no slot for a single OAuth token string.

## Proposal

Mirror the `apiKey` pattern with a new `oauthToken` field:

```ts
export interface ProfileConfig {
  id: string
  type?: "claude-max" | "api" | "oauth-token"
  claudeConfigDir?: string
  apiKey?: string
  baseUrl?: string
  oauthToken?: string        // NEW — value from `claude setup-token`
}
```

In `buildResolvedProfile()`, expose it as `CLAUDE_CODE_OAUTH_TOKEN` to the SDK subprocess:

```ts
if (profile.oauthToken) {
  return {
    id: profile.id,
    type: "oauth-token",
    env: { CLAUDE_CODE_OAUTH_TOKEN: profile.oauthToken },
  }
}
```

### `profiles.json` example

```json
[
  { "id": "personal",    "claudeConfigDir": "/home/me/.claude" },
  { "id": "ci-token",    "oauthToken": "sk-ant-oat01-..." },
  { "id": "direct-api",  "apiKey": "sk-ant-api03-..." }
]
```

### CLI

```bash
meridian profile add ci-token --oauth-token sk-ant-oat01-...
# or prompt for the token instead of running `claude auth login`
```

## Why this is useful

- **One-line registration** — same UX as `apiKey`, no sentinel directory on disk
- **Cross-platform** — works on macOS without touching Keychain (the only file-based workaround today is Linux-only because of `tokenRefresh.ts`'s Keychain backend)
- **CI / Docker / remote** — no browser flow needed
- **Multi-account headless setups** — drop several long-lived tokens into `profiles.json` and switch with `x-meridian-profile`

Happy to send a PR if the direction looks good.

isac322:
Closes #446

## Summary

Adds an `oauth-token` profile type so Meridian can be configured with a long-lived OAuth token from `claude setup-token`. The new shape mirrors `apiKey`: a single string value in `profiles.json`, no per-profile config directory, no browser handshake. Useful for headless, CI, and Docker setups.

## Why

Per #446, the existing profile shapes both require state on disk:
- `claudeConfigDir` profiles need a Claude Code config directory populated by an interactive `claude auth login` (browser flow).
- `apiKey` profiles work but route through the direct Anthropic API, which is a different billing/quota path than Claude Max.

`claude setup-token` mints a ~1-year OAuth token specifically for headless/CI usage, and the Claude Code SDK natively honors `CLAUDE_CODE_OAUTH_TOKEN`. Surfacing it as a Meridian profile gives a one-line registration that works cross-platform, requires no Keychain access, and unlocks multi-account headless via `x-meridian-profile`.

### `src/proxy/profiles.ts`
- Add `'oauth-token'` to `ProfileType`.
- Add `oauthToken?: string` to `ProfileConfig`.
- Extend `buildResolvedProfile` with an oauth-token branch (checked first): emits `{ env: { CLAUDE_CODE_OAUTH_TOKEN } }`. Presence of `oauthToken` infers the type, so `{ "id": "ci", "oauthToken": "..." }` works without an explicit `type` field. An explicit `type: "oauth-token"` is also accepted.

### `src/proxy/profileCli.ts`
- New exported `profileAddOauthToken(id, tokenArg?)`: validates id, dedupes, prompts when `tokenArg` is omitted, and writes `{ id, type: 'oauth-token', oauthToken }` to `profiles.json` only — no config directory is created.
- New `promptToken(question)` helper: line-buffered when stdin is piped (so `echo $TOKEN | meridian profile add ci --oauth-token` works), and silent (no echo) raw-mode prompt on a TTY. Ctrl+C cancels with non-zero exit.
- `profileList` short-circuits oauth-token entries with a `✓ OAuth token` line and skips the `claude auth status` lookup (these profiles have no config dir).
- `profileLogin` refuses oauth-token profiles with a helpful message pointing at `remove + add --oauth-token`.
- `profileRemove` is unchanged; it was already safe for oauth-token profiles (`existsSync("") === false` skips the `rmSync`).
- Help text updated: `--oauth-token [TOKEN]` example added; `login` annotated as browser-login only.

### `bin/cli.ts`
- `profile add <name>` now accepts `--oauth-token [TOKEN]`. With a value: passes it directly. Without: hands off to the secure prompt.

### `src/__tests__/profiles-unit.test.ts`
- 6 new unit tests in the existing `resolveProfile` and `listProfiles` blocks:
  - resolves oauth-token profile via explicit type
  - infers oauth-token type from `oauthToken` field alone
  - oauth-token wins over `claudeConfigDir` when both are present
  - bare `type: 'oauth-token'` without a value returns empty env
  - header selects oauth-token profile when active points elsewhere
  - listProfiles preserves explicit `oauth-token` type

### `README.md`
- New "Headless / CI: register an OAuth token" subsection under "Adding profiles" with prompt + CLI-arg examples.
- Profile commands table: `--oauth-token [TOKEN]` row added; `login` row qualified as browser-login only.
- "How it works" note clarifies that oauth-token profiles live entirely in `~/.config/meridian/profiles.json` and feed the SDK via `CLAUDE_CODE_OAUTH_TOKEN`.
- `MERIDIAN_PROFILES` example expanded to include all three profile shapes.
- Bottom CLI Commands table mirrors the new row + qualifier.
- New Docker subsection "OAuth-token profiles in Docker (no volume mount)" documenting the simpler env-only path.

## CLI examples

```bash
# Prompted (input is hidden)
meridian profile add ci --oauth-token

# Direct value
meridian profile add ci --oauth-token sk-ant-oat01-...

# Pipe (non-TTY)
echo $TOKEN | meridian profile add ci --oauth-token
```

## Resulting `profiles.json`

```json
[
  { "id": "personal",   "claudeConfigDir": "/home/me/.claude" },
  { "id": "ci-token",   "type": "oauth-token", "oauthToken": "sk-ant-oat01-..." },
  { "id": "direct-api", "apiKey": "sk-ant-api03-..." }
]
```

`profiles.json` continues to be written at mode `0600`.

## Compatibility

- Additive only. Existing `claudeConfigDir` and `apiKey` profiles are untouched and continue to work.
- Profile resolution priority unchanged: header > active > default > first effective.
- `ProfileConfig`, `ResolvedProfile`, `x-meridian-profile`, and the `/profiles/*` endpoints are extended in a backwards-compatible way (no field renames, no removals).
- Leaf-module rule preserved: `profiles.ts` and `profileCli.ts` add no imports from `server.ts` or `session/`.

## Verification

- `npm run build` (bun build + tsc) — exit 0.
- `bun test src/__tests__/profiles-unit.test.ts` — 31/31 pass.
- Full `npm test` — 1586 pass / 3 unrelated pre-existing flakes (`session/cache.ts`, `models.ts`, `server.ts` paths; none touch the profiles surface).
- Manual matrix worth running against your Claude Max account before merge:
  - `meridian profile add ci-test --oauth-token` (prompted) → profiles.json contains `{id, type:'oauth-token', oauthToken}`.
  - `meridian profile list` → shows `✓ OAuth token`.
  - `meridian profile login ci-test` → refused with the `remove + add --oauth-token` hint.
  - `meridian profile remove ci-test` → succeeds without trying to delete a directory.
  - With `x-meridian-profile: ci-test`, request reaches the SDK with `CLAUDE_CODE_OAUTH_TOKEN` set in env.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ meridian profile add ci --oauth-token
 meridian profile add ci --oauth-token sk-ant-oat01-...
 ```
 
-OAuth-token profiles store nothing on disk besides the token in `profiles.json` — there's no per-profile config dir, no Keychain entry, no browser handshake. The Claude Code SDK reads the token from `CLAUDE_CODE_OAUTH_TOKEN` for any request routed to that profile.
+OAuth-token profiles store the token in `profiles.json` and feed it to the SDK via `CLAUDE_CODE_OAUTH_TOKEN` — no Keychain entry, no browser handshake. To prevent the SDK's 401-recovery from silently falling back to the host's `~/.claude` credentials, OAuth-token profiles also pin `CLAUDE_CONFIG_DIR` to an isolated per-profile directory under `~/.config/meridian/profiles/<name>/`. That directory holds only SDK state (sessions, settings) — never `.credentials.json`, since the token is delivered through the env.
 
 ### Switching profiles
 
@@ -278,7 +278,7 @@ You can also switch profiles from the web UI at `http://127.0.0.1:3456/profiles`
 
 ### How it works
 
-Each profile stores its credentials in an isolated `CLAUDE_CONFIG_DIR` under `~/.config/meridian/profiles/<name>/`. OAuth-token profiles are the exception — they live entirely in `~/.config/meridian/profiles.json` and feed the SDK via `CLAUDE_CODE_OAUTH_TOKEN` directly. When a request arrives, Meridian resolves the profile in priority order:
+Each profile stores its credentials in an isolated `CLAUDE_CONFIG_DIR` under `~/.config/meridian/profiles/<name>/`. OAuth-token profiles use the same isolated directory layout — but the token itself lives in `~/.config/meridian/profiles.json` and is fed to the SDK via `CLAUDE_CODE_OAUTH_TOKEN`, so the per-profile dir holds only SDK state (sessions, settings) and never the credential. When a request arrives, Meridian resolves the profile in priority order:
 
 1. `x-meridian-profile` request header (per-request override)
 2. Active profile (set via `meridian profile switch` or the web UI)
--- a/README.md
+++ b/README.md
@@ -239,6 +239,20 @@ meridian profile add work
 
 > **⚠ Important:** Claude's OAuth reuses your browser session. Before adding a second account, sign out of claude.ai and sign into the other account first.
 
+#### Headless / CI: register an OAuth token
+
+When a browser isn't available (containers, CI runners, remote shells), generate a long-lived OAuth token with `claude setup-token` and register it as a profile:
+
+```bash
+# Prompt for the token (input is hidden — paste the value from `claude setup-token`)
+meridian profile add ci --oauth-token
+
+# Or pass it inline
+meridian profile add ci --oauth-token sk-ant-oat01-...
+```
+
+OAuth-token profiles store nothing on disk besides the token in `profiles.json` — there's no per-profile config dir, no Keychain entry, no browser handshake. The Claude Code SDK reads the token from `CLAUDE_CODE_OAUTH_TOKEN` for any request routed to that profile.
+
 ### Switching profiles
 
 ```bash
@@ -256,14 +270,15 @@ You can also switch profiles from the web UI at `http://127.0.0.1:3456/profiles`
 | Command | Description |
 |---------|-------------|
 | `meridian profile add <name>` | Add a profile and authenticate via browser |
+| `meridian profile add <name> --oauth-token [TOKEN]` | Add a headless profile from a `claude setup-token` value (prompts when `TOKEN` is omitted) |
 | `meridian profile list` | List profiles and auth status |
 | `meridian profile switch <name>` | Switch the active profile (requires running proxy) |
-| `meridian profile login <name>` | Re-authenticate an expired profile |
+| `meridian profile login <name>` | Re-authenticate an expired profile (browser-login profiles only) |
 | `meridian profile remove <name>` | Remove a profile and its credentials |
 
 ### How it works
 
-Each profile stores its credentials in an isolated `CLAUDE_CONFIG_DIR` under `~/.config/meridian/profiles/<name>/`. When a request arrives, Meridian resolves the profile in priority order:
+Each profile stores its credentials in an isolated `CLAUDE_CONFIG_DIR` under `~/.config/meridian/profiles/<name>/`. OAuth-token profiles are the exception — they live entirely in `~/.config/meridian/profiles.json` and feed the SDK via `CLAUDE_CODE_OAUTH_TOKEN` directly. When a request arrives, Meridian resolves the profile in priority order:
 
 1. `x-meridian-profile` request header (per-request override)
 2. Active profile (set via `meridian profile switch` or the web UI)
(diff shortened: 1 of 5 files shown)


isac322:
Pushed a follow-up fix: `fix(profiles): isolate oauth-token profile from host ~/.claude` (16438e5).

While testing this PR locally on a machine with an active `claude login`, I noticed `oauth-token` profiles were silently using the host's `~/.claude` credentials instead of the env-provided token. Reproduced with three scenarios:

| Setup | Result |
|---|---|
| **bad** token, no `CLAUDE_CONFIG_DIR` (host has `claude login`) | 200 OK — token ignored, host creds used |
| **bad** token, isolated empty `CLAUDE_CONFIG_DIR` | 401 — token actually used and rejected |
| **valid** token, isolated empty `CLAUDE_CONFIG_DIR` | 200 OK — token genuinely used |

The cause is in the Claude Agent SDK itself. The SDK does prefer `CLAUDE_CODE_OAUTH_TOKEN` for the first request attempt, so the env path works on its own. But when an API call returns 401, the SDK has a recovery path that reads `~/.claude/.credentials.json`, refreshes via the stored refresh token, and overwrites `process.env.CLAUDE_CODE_OAUTH_TOKEN` with the new value before retrying (it emits a `tengu_oauth_401_recovered_from_disk` telemetry event when this fires). On a host with an active login this silently masks any token failure — including the case where the user expected the env token to be the *only* credential in play. So in practice, `oauth-token` profiles on a logged-in host become indistinguishable from "use whatever's logged in".

Fix: in `buildResolvedProfile` (`src/proxy/profiles.ts`), set `CLAUDE_CONFIG_DIR=~/.config/meridian/profiles/<id>/` for `oauth-token` profiles so the SDK can never reach the host's credentials directory. The directory holds only SDK state (sessions, settings); no `.credentials.json` is written there, since auth comes from the env. The path is deliberately *not* `~/.claude` to avoid the Keychain corruption from upstream [claude-code#20553](https://github.com/anthropics/claude-code/issues/20553) that's already documented in `query.ts`.

README updated to reflect the new behavior. Verified end-to-end:

- 32/32 in `profiles-unit.test.ts` pass (with one new test asserting `CLAUDE_CONFIG_DIR` is set + isolated)
- Full suite: same pre-existing flaky failures as `main`, no regressions from this commit (verified by stash + re-run)
- Build succeeds; `dist/cli.js` deployed via systemd drop-in on my host and confirmed working with my real token: the SDK creates only `.claude.json` / `backups` / `projects` / `sessions` in the isolated dir — no `.credentials.json` — and PONG roundtrips succeed.

Side effect: `oauth-token` profiles now get isolated session/memory state per profile, matching `claude-max` profile semantics. CI/Docker behavior is unchanged (no host `~/.claude` exists in those environments anyway).

rynfar:
Great concept here.  Let me review this.

rynfar:
Hey @isac322 — thanks for this. The motivation reads exactly right (browser-flow + apiKey weren't enough; `claude setup-token` is the missing third shape), and the second commit is the strong part of this PR — catching that the SDK's 401-recovery silently reads host `~/.claude` and swaps a refreshed token in for the env-provided one is a non-obvious failure mode that would have burned someone in prod. The fix (pin to `~/.config/meridian/profiles/<id>/`, deliberately not `~/.claude`) is the right one and correctly avoids the keychain-corruption case from #453.

I went through it line-by-line and want to merge it — with two small follow-ups stacked on top.

**1. `sharedMemory: true` undoes your isolation fix.** `stripConfigDir` in `query.ts` unconditionally removes `CLAUDE_CONFIG_DIR` whenever `sharedMemory` is on (the existing #453 workaround). For an oauth-token profile, that strips the pin you just installed, and the SDK falls back to host creds again — the exact failure your second commit was added to prevent. Latent today since `sharedMemory` is per-adapter and defaults off, but worth closing the gap. Fix: skip the strip when `CLAUDE_CODE_OAUTH_TOKEN` is in env.

**2. `profile remove` leaks the isolation dir.** Your second commit creates `~/.config/meridian/profiles/<id>/` (via the SDK populating state at runtime), but `profileRemove` only deletes `profile.claudeConfigDir` — unset for oauth-token profiles — so removing the profile leaves an orphan dir on disk forever. Fix: extract a pure `dirsToRemoveOnProfileRemove` helper, route both shapes through it.

Rather than block your PR on these, I cherry-picked both of your commits as-is (authorship preserved) and stacked one fix commit per issue on top, each with a regression test that fails on unfixed code. Opening that as #483 — see the PR for the diff and the failing-then-passing tests. Closing this one in favor of that.

Plan to **rebase-merge** #483 so all 4 commits — your 2 + my 2 — land on `main` individually with original authorship. `git log` and the contribution graph will reflect your work correctly.

Two further items I noticed but did **not** fix in #483 (out of scope, worth a separate issue if you want):
- Type inference is asymmetric (`oauthToken` infers, `apiKey` does not) — small JSDoc clarification at most.
- `promptToken` shells out to `node -e` for raw-mode TTY — pragmatic given Bun's stdin quirks, but worth a Bun-native rewrite later.

Thanks for this — the headless / Docker path it unlocks is a real gap closer.

rynfar:
Closed in favor of #483, which contains both of your commits (authorship preserved) plus two stacked fixes on top (sharedMemory strip gate + remove cleanup). Will rebase-merge #483 so your work lands on main with your authorship intact.

PR #479, opened 2026-05-04
fix(tokenRefresh): close trigger gap + add traffic-independent refresh

CrazyCoder:
## Problem

After deploying `meridian-v1.41.1` (with #464's compact-JSON fix in place) on two long-running NAS-hosted Meridian instances, both started returning 401 to clients silently after sitting idle:

```
{"type":"error","error":{"type":"authentication_error",
 "message":"Claude authentication expired or invalid. Run 'claude login' in your terminal to re-authenticate, then restart the proxy."}}
```

Symptoms when diagnosed (2026-05-03):

- `/health` reported `loggedIn: true` (the local `claude auth status` only reads `.credentials.json` and trusts `expiresAt` — it doesn't probe Anthropic).
- `.credentials.json` mtime had **not** ticked since the operator's last `claude login` — 25h+ on one instance, ~2 days on the other.
- `docker logs … | grep -E 'token_refresh|OAuth token expired'` was **empty** — neither the inline retry nor any refresh attempt had ever fired despite continuous traffic.
- A manual `POST https://platform.claude.com/v1/oauth/token` with the on-disk refresh_token returned `400 invalid_grant: Refresh token not found or invalid`. Anthropic had server-side invalidated the unused refresh token.

A dev-stack proxy on the *same image*, logged in to the same account 2 days earlier, had no problem refreshing the moment we sent it traffic — implication: the refresh path *can* succeed if exercised regularly. What kills it is letting the access token sit expired long enough that Anthropic GC's the underlying refresh token.

## Two gaps in the existing design

#230 made refresh **inline-on-401**, gated on `isExpiredTokenError(errMsg)` matching `"oauth token has expired"` or `"not logged in"`. That's the only refresh trigger today.

**Gap #1 — trigger surface is too narrow.** Anthropic's API can return a generic 401 for a stale access token without echoing the CLI-specific wording. Such errors fall through to `classifyError`'s broader auth branch (`401 | authentication | invalid auth | credentials`) and surface as `"Claude authentication expired or invalid"` to the client. Refresh is never even attempted. This is the proximate cause of the silent failure above.

**Gap #2 — there is no traffic-independent refresh path.** Even with a fully broadened trigger, an idle proxy that never receives a request never exercises the chain. Anthropic's GC of unused refresh tokens (timeline appears to be ~24-48h based on what we saw, exact policy unknown) eventually invalidates the chain entirely; once that happens, `claude login` is the only recovery.

## Changes (4 commits)

| # | Commit | Layer | What |
|---|--------|-------|------|
| 1 | `fix(errors): broaden isExpiredTokenError to catch generic 401s` | reactive | Adds matchers for RFC 6750 wording (`invalid_token`, `token_expired`) and `"401" + ("authentication" \| "unauthorized" \| "invalid")`. Existing positive cases still match; existing negative cases (rate-limit errors, unrelated auth errors without 401) still return false. False positives are bounded — the existing `tokenRefreshed` guard makes refresh single-shot per request; a misclassified error costs at most one OAuth round-trip and the original error still bubbles up. |
| 2 | `feat(tokenRefresh): proactive ensureFreshToken before SDK call` | proactive | New `ensureFreshToken(store?, bufferMs = 5min)` helper. Reads credentials cheaply; returns true when far from expiry, calls `refreshOAuthToken` when within buffer. Wired into both SDK call sites in `server.ts` (non-stream + stream) before the per-request retry loop. Best-effort `.catch(() => {})` — reactive 401-retry handles anything missed. |
| 3 | `feat(tokenRefresh): background scheduler keeps refresh chain warm` | **traffic-independent** | `startBackgroundRefresh()` arms a self-rescheduling `setTimeout` at `expiresAt - bufferMs`, fires `refreshOAuthToken` regardless of traffic, schedules the next call based on the new `expiresAt` on success, retries every `failureRetryMs` (default 5min) on failure. Booted from `startProxyServer()` after `serve()`; `stopBackgroundRefresh()` wired into the close path. Idempotent (`scheduledRefreshActive` flag), `.unref()`-ed timer so it doesn't hold the process. **This is the load-bearing fix** — the other two close the trigger gap, but this is what keeps the chain warm for unattended deployments. |
| 4 | `feat(tokenRefresh): make scheduler activity visible in default logs` | observability | `claudeLog()` is gated on `OPENCODE_CLAUDE_PROVIDER_DEBUG` (unset in prod), so scheduled-refresh events would have been silent. One `console.error('[token_refresh] scheduled refresh (immediate) ok=…')` line per immediate refresh — kept narrow so a healthy scheduler is mostly silent (logs only fire when refresh actually runs). |

## How this fits with existing patterns

- Builds on #230's platform-aware `CredentialStore` interface — the scheduler uses the same store, so it inherits Keychain-on-macOS / file-on-Linux without re-introducing #224's macOS bug.
- Co-exists with #279's per-profile `authKeepaliveInterval` (which does a *local* `claude auth status` poll). Different concern: that one keeps the local in-memory cache fresh; this one keeps the upstream OAuth chain warm. No overlap.
- Mirrors #439's "401 → refresh → retry" pattern as already implemented in `oauthUsage.ts` — same shape, applied to the `server.ts` refresh sites.
- No new env vars, no new CLI subcommands, no external scheduler required.

## Tests

`bun test src/__tests__/token-refresh.test.ts src/__tests__/errors.test.ts`: 99 pass / 0 fail. The new tests:

- 4 for the broadened `isExpiredTokenError` (new positive matches for generic 401s and RFC 6750 codes; new negative cases for non-401 incidental matches like `"unauthorized scope"`).
- 7 for `ensureFreshToken` (under/over buffer, no credentials, no `expiresAt`, refresh failure, custom buffer).
- 6 for `startBackgroundRefresh` (already-expired immediate refresh, scheduling without refresh, no-credentials polling, retry-on-failure, idempotency, stop() prevents future fires).

Real timers with short delays (50–120 ms) — the scheduler suite runs in ~2 s.

## Soak evidence

Branch deployed to two production NAS instances on 2026-05-03 18:00 UTC. ~21 hours in:

```
$ docker logs meridian-proxy-1 2>&1 | grep '\[token_refresh\]'
[token_refresh] scheduled refresh (immediate) ok=true
[token_refresh] scheduled refresh (immediate) ok=true

$ docker logs meridian-work-proxy-1 2>&1 | grep '\[token_refresh\]'
[token_refresh] scheduled refresh (immediate) ok=false
[token_refresh] scheduled refresh (immediate) ok=true
[token_refresh] scheduled refresh (immediate) ok=true
```

The single `ok=false` line on the work instance is the boot-time probe against an already-invalidated refresh token (the chain that triggered the original investigation) — operator did `claude login`, the next 5-min retry tick read the new credentials, fired a successful refresh, and the chain has been auto-refreshing ever since. `.credentials.json` mtimes are ticking on schedule on both instances; no client-visible 401s since deploy.

## Forced-expiry repro

Reproducible locally with the dev stack:

```bash
docker exec dev-proxy-1 node -e '
  const fs = require("fs"); const p = "/home/claude/.claude/.credentials.json";
  const c = JSON.parse(fs.readFileSync(p, "utf8"));
  c.claudeAiOauth.expiresAt = Date.now() - 3600000;
  fs.writeFileSync(p, JSON.stringify(c));'
docker restart dev-proxy-1
sleep 6
docker logs dev-proxy-1 2>&1 | grep '\[token_refresh\]'
# [token_refresh] scheduled refresh (immediate) ok=true
docker exec dev-proxy-1 stat -c '%y' /home/claude/.claude/.credentials.json
# (mtime ~3 seconds after restart)
```

No traffic involved — the scheduler fires at boot because `expiresAt` is in the past, refreshes successfully, persists, and arms a long-cycle timer.

## Known limitations (out of scope for this PR)

The new call sites in this PR call `refreshOAuthToken()` / `createPlatformCredentialStore()` without a `claudeConfigDir` argument. Under multi-profile setups with `CLAUDE_CONFIG_DIR` overlays, only the default credential store gets refreshed. This mirrors the *existing* upstream behavior at the already-merged refresh sites (`server.ts` ~1062, ~1520, ~2502) — this PR doesn't introduce the gap, but it does extend it to a few more sites. `inflightRefresh` and the scheduler state are also process-global, so two profiles in one process can't refresh independently.

The fix is to thread `claudeConfigDir` through the refresh helpers — `oauthUsage.ts` (#439) already shows the right pattern. Happy to follow up with a separate PR that retrofits all the refresh sites to be profile-aware if you'd like; kept it out of this PR to keep the diff focused on the trigger gap and the missing traffic-independent path.

## Compatibility with open work

#474 (open, oauth-token profile type) doesn't conflict — its tokens have ~1 yr TTL and bypass the refresh path entirely. If it merges, the scheduler should grow a `ProfileType !== "oauth-account"` skip so it doesn't try to refresh long-lived static tokens. One-line guard, easy follow-up.

---

Closes the silent-401 failure mode that #223 originally described and that #230 partially addressed. Tested locally + 21 h prod soak across two long-running unattended instances.

### DIFF
--- a/src/__tests__/token-refresh.test.ts
+++ b/src/__tests__/token-refresh.test.ts
@@ -246,6 +246,101 @@ describe("refreshOAuthToken", () => {
 })
 
 // ---------------------------------------------------------------------------
+// ensureFreshToken — proactive refresh before SDK call
+// ---------------------------------------------------------------------------
+
+describe("ensureFreshToken", () => {
+  let originalFetch: typeof globalThis.fetch
+  beforeEach(() => { originalFetch = globalThis.fetch })
+  afterEach(async () => {
+    globalThis.fetch = originalFetch
+    const { resetInflightRefresh } = await import("../proxy/tokenRefresh")
+    resetInflightRefresh()
+  })
+
+  function makeStoreWithExpiry(expiresAt: number) {
+    return makeStore({ ...MOCK_CREDENTIALS, claudeAiOauth: { ...MOCK_CREDENTIALS.claudeAiOauth, expiresAt } })
+  }
+
+  it("returns true without refreshing when token is far from expiry", async () => {
+    const { ensureFreshToken } = await import("../proxy/tokenRefresh")
+    const fetchSpy = mock(() => Promise.reject(new Error("fetch should not be called")))
+    mockFetch(fetchSpy)
+    const { store, writes } = makeStoreWithExpiry(Date.now() + 60 * 60 * 1000) // +1h, well outside default 5min buffer
+
+    const ok = await ensureFreshToken(store)
+    expect(ok).toBe(true)
+    expect(fetchSpy).not.toHaveBeenCalled()
+    expect(writes).toHaveLength(0)
+  })
+
+  it("refreshes when token is inside the buffer (near-expiry)", async () => {
+    const { ensureFreshToken } = await import("../proxy/tokenRefresh")
+    const fetchSpy = mock(() => Promise.resolve(makeSuccessResponse(MOCK_TOKEN_RESPONSE)))
+    mockFetch(fetchSpy)
+    const { store, getStored } = makeStoreWithExpiry(Date.now() + 60 * 1000) // +1min, well inside default 5min buffer
+
+    const ok = await ensureFreshToken(store)
+    expect(ok).toBe(true)
+    expect(fetchSpy).toHaveBeenCalledTimes(1)
+    expect(getStored()?.claudeAiOauth.accessToken).toBe("new-access-token")
+  })
+
+  it("refreshes when token is already expired", async () => {
+    const { ensureFreshToken } = await import("../proxy/tokenRefresh")
+    const fetchSpy = mock(() => Promise.resolve(makeSuccessResponse(MOCK_TOKEN_RESPONSE)))
+    mockFetch(fetchSpy)
+    const { store, getStored } = makeStoreWithExpiry(Date.now() - 60 * 60 * 1000) // -1h
+
+    const ok = await ensureFreshToken(store)
+    expect(ok).toBe(true)
+    expect(fetchSpy).toHaveBeenCalledTimes(1)
+    expect(getStored()?.claudeAiOauth.accessToken).toBe("new-access-token")
+  })
+
+  it("returns false when no credentials stored", async () => {
+    const { ensureFreshToken } = await import("../proxy/tokenRefresh")
+    const fetchSpy = mock(() => Promise.reject(new Error("fetch should not be called")))
+    mockFetch(fetchSpy)
+    const { store } = makeStore(null)
+    const ok = await ensureFreshToken(store)
+    expect(ok).toBe(false)
+    expect(fetchSpy).not.toHaveBeenCalled()
+  })
+
+  it("returns false when expiresAt missing", async () => {
+    const { ensureFreshToken } = await import("../proxy/tokenRefresh")
+    const fetchSpy = mock(() => Promise.reject(new Error("fetch should not be called")))
+    mockFetch(fetchSpy)
+    // expiresAt = 0 is falsy via `!expiresAt`
+    const { store } = makeStoreWithExpiry(0)
+    const ok = await ensureFreshToken(store)
+    expect(ok).toBe(false)
+    expect(fetchSpy).not.toHaveBeenCalled()
+  })
+
+  it("returns false when refresh request fails", async () => {
+    const { ensureFreshToken } = await import("../proxy/tokenRefresh")
+    mockFetch(() => Promise.resolve(new Response("invalid_grant", { status: 400 })))
+    const { store } = makeStoreWithExpiry(Date.now() - 1000)
+    const ok = await ensureFreshToken(store)
+    expect(ok).toBe(false)
+  })
+
+  it("respects custom bufferMs", async () => {
+    const { ensureFreshToken } = await import("../proxy/tokenRefresh")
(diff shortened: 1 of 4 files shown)


rynfar:
Hey @CrazyCoder — first off, thank you for this work. The diagnosis is the kind of thing that's hard to nail down (silent 401 + invalidated refresh chain after idle, with `loggedIn:true` lying about it), and the three-layer fix (broaden trigger → proactive → background scheduler) lands exactly where it needs to. The 21h soak across two NAS instances is solid evidence.

I went through it line-by-line and want to merge it — with one small adjustment.

During review I found a race in the background scheduler: the single `scheduledRefreshActive` boolean isn't sufficient to invalidate an in-flight `scheduleNext()` chain across a `stop() + start()` (e.g. during a hot reload). When the first chain's `await store.read()` resolves *after* the second `start()` re-flips the boolean, both chains arm follow-up timers — only the latest is tracked in `scheduledRefreshTimer`, the orphan keeps firing in parallel, and the refresh rate doubles forever. Edge case in practice (you'd need a rapid stop/start while a read is parked), but easy to fix and worth catching because it compounds silently.

Rather than block your PR on it, I cherry-picked your 4 commits as-is (authorship preserved) and stacked one fix commit on top with a regression test. Opening that as #482 — see the PR for the diff and the failing-then-passing test. Closing this one in favor of that.

Plan to **rebase-merge** #482 so all 4 of your commits land on `main` individually with your authorship — `git log` and the contribution graph will reflect your work correctly.

Thanks again — this is a real fix to a problem we've been getting bit by.

rynfar:
Closed in favor of #482, which contains all 4 of your commits (authorship preserved) plus one stacked fix on top. Will rebase-merge #482 so your work lands on main with your authorship intact.

PR #480, opened 2026-05-05
fix(plugins): convert Windows paths to file:// URLs for ESM import

heckradr:
On Windows, import() rejects backslash paths like C:\Users\... with 'Received protocol c:'. Use pathToFileURL() to convert absolute paths to file:// URLs before dynamic import.

Fixes plugin loading via plugins.json on Windows.

### DIFF
--- a/src/proxy/plugins/loader.ts
+++ b/src/proxy/plugins/loader.ts
@@ -1,5 +1,6 @@
 import { readdirSync, readFileSync, existsSync } from "fs"
 import { join, isAbsolute, extname } from "path"
+import { pathToFileURL } from "url"
 import type { Transform } from "../transform"
 import type { PluginEntry, PluginConfig, LoadedPlugin } from "./types"
 import { validateTransform } from "./validation"
@@ -93,7 +94,7 @@ export async function loadPlugins(
       // many times within the same millisecond (tests, hot iteration) the
       // timestamp alone collides and we serve the cached build.
       const cacheBuster = `?t=${Date.now()}-${++loadCounter}`
-      const mod = await import(filePath + cacheBuster)
+      const mod = await import(pathToFileURL(filePath).href + cacheBuster)
       const exported = mod.default ?? mod
 
       // Support single Transform or array of Transforms


rynfar:
thanks for submitting this! i'll review this shortly

rynfar:
Hey @heckradr — thanks for catching this. The diagnosis is right (`Received protocol c:` is exactly what ESM does to a `C:\\…` path) and `pathToFileURL` is the correct fix. Pulled it locally and went through it line-by-line.

There's one platform-asymmetry I caught while running the tests against your branch: applying `pathToFileURL` unconditionally **silently breaks `/plugins/reload` on POSIX.** Bun's module cache keys off the full specifier string when given a raw path (so `?t=…` evicts), but it normalizes `file://` URLs and drops the query *before* computing the cache key. After your change, every reload kept serving the previously-loaded module. The existing integration test caught it cleanly:

```
(fail) Plugin integration — end-to-end via HTTP > POST /plugins/reload picks up changes without restart
  Expected: defined / Received: undefined
```

Latent on this PR because fork CI doesn't auto-run, but it would have hit anyone running `/plugins/reload` against the merged version.

Rather than block your fix on that, I cherry-picked your commit as-is (authorship preserved) and stacked one fix on top with a platform gate — `pathToFileURL` only on `win32`, raw path on POSIX — and wired `plugin-loader.test.ts` into the `windows-smoke` job so the buggy code path actually gets exercised on the Windows runner going forward. Opening that as #484. Closing this one in favor of that.

Plan to **rebase-merge** #484 so your commit lands on `main` with your authorship intact.

Small note: your commit was authored with a literal `[EMAIL]` placeholder in the git config (`Heckrath, Andre <[EMAIL]>`), so the commit's `Author:` field won't link to your GitHub avatar in `git log` even though the PR-level credit shows up correctly. Worth fixing your local `git config user.email` for future PRs — not a blocker here.

Thanks for this — the Windows path was a real gap.

rynfar:
Closed in favor of #484, which contains your commit (authorship preserved) plus a platform gate that prevents the silent /plugins/reload regression on POSIX, plus a windows-smoke CI step that catches future regressions in the loader's URL handling on Windows. Will rebase-merge #484.

PR #487, opened 2026-05-06
feat(adapters): cherry-studio chat-client adapter

### LINKED ISSUE #481: Cherry Studio Support + Web Search Issues
Could we get [Cherry Studio](https://github.com/CherryHQ/cherry-studio) added as a harness customizable like OpenCode, Crush, etc? It's a pretty nice chat client that comes with autonomous agents, assistants, etc. I've started using it instead of using the Claude Desktop app.

I was able to get it working by setting the API url as `http://127.0.0.1:3456`, but it doesn't seem to work with Web Searches, saying it doesn't have the tools `WebSearch` or `WebFetch` tool exposed in the session. Cherry Studio has Web Search integrated, but I still can't seem to get Claude to use web searches at all.
Image

It'd be nice to customize things like Client Prompt, Thinking Passthrough, Thinking like other harnesses, too.

rynfar:
Closes #481.

> **Hold on merging.** This needs end-user verification in actual Cherry Studio UI before shipping — the adapter is server-side correct but I can't drive a desktop Electron app from CI. @BenIsLegit / @rynfar to confirm before merging.

## Why

Cherry Studio is a desktop chat client that, until this PR, falls through to the OpenCode adapter (the default fallback). OpenCode blocks `WebSearch` and `WebFetch` because it has MCP equivalents (`websearch_web_search_exa`). Cherry Studio has no MCP equivalent and no client-side web access — when blocked, the user sees:

> WebSearch tool not exposed in the session

and Claude can't search the web. Reported by @BenIsLegit in #481.

## What

A new `cherry-studio` adapter — the first **chat-client** adapter (every existing adapter is a CLI coding agent). The chat-client shape is the inverse of the coding-agent shape: web tools enabled, filesystem tools off-by-default, no MCP, no passthrough.

| Property | Coding-agent adapters (OpenCode, Crush, …) | Cherry Studio |
|---|---|---|
| `WebSearch` / `WebFetch` | blocked (agent has MCP equivalent) | **allowed** |
| `Read` / `Write` / `Bash` / etc. | passed through to agent | **blocked** by default |
| `usesPassthrough` | varies (often true) | false — SDK runs tools, returns results inline |
| `getAllowedMcpTools` | mcp__opencode__read, etc. | none |
| `buildSdkAgents` | subagent routing | none |

### Detection

Cherry Studio's User-Agent is overridden ([CherryHQ#10209](https://github.com/CherryHQ/cherry-studio/issues/10209)) so auto-detection isn't reliable. Routes via:

- `x-meridian-agent: cherry-studio` header (per-request — the documented path)
- `MERIDIAN_DEFAULT_AGENT=cherry-studio` env var (global default)

Aliases: `cherry-studio`, `cherrystudio` (no hyphen).

### Bonus fix: single source of truth for the adapter list

While in there I noticed the settings UI hardcoded `["opencode", "crush", "forgecode", "pi", "droid", "passthrough"]` in two places — both drifted from `ADAPTER_MAP`. `claude-code` was registered for detection but never rendered in `/settings`, so its feature toggles were silently inaccessible. New `ADAPTER_LABELS` / `ADAPTER_NAMES` exports in `detect.ts` are the single source; `sdkFeatures.ts` and `settingsPage.ts` consume them. Adding `cherry-studio` AND fixing the latent `claude-code` gap is automatic. An audit test guards against future drift.

### Why filesystem tools are blocked by default

A chat-style LLM shouldn't be able to enumerate `~/.ssh` or read `/etc/secrets` on the proxy host because the user typed "what's in my home directory?" into Cherry Studio. Even on localhost, the user is rarely thinking "this chat sends Read/Bash to the server." Operators who genuinely want filesystem access can use a coding-agent adapter (where tool calls are surveilled by the calling tool) or open a follow-up issue if there's demand for a "trusted chat client" mode with broader access.

## Tests

21 new tests in `cherry-studio-adapter.test.ts`:

- Identity, session/CWD shape (5)
- **Load-bearing**: `WebSearch` / `WebFetch` NOT in blocked + incompatible lists (2)
- Filesystem tools blocked (2)
- Claude-Code orchestration tools blocked (1)
- Chat-client behavior toggles — `usesPassthrough`, `supportsThinking`, etc. (7)
- `x-meridian-agent` detection incl. alias and case-insensitivity (3)
- Adapter-list audit: every key in `ADAPTER_LABELS` resolves via `ADAPTER_MAP`, gets a feature config from `getAllFeatureConfigs`, and includes `cherry-studio` + `claude-code` (4)

Full suite **1742/0**. Build clean. Typecheck clean.

## End-to-end verification (server-side)

Booted a test proxy on `:3500` (separate from prod on `:3456`, prod undisturbed throughout):

```bash
curl -X POST http://127.0.0.1:3500/v1/messages \
  -H 'x-meridian-agent: cherry-studio' \
  -d '{"model":"claude-haiku-4-5","max_tokens":400,"messages":[{"role":"user",
       "content":"Use WebSearch to find the latest version of node.js. Reply with just the version number and a citation URL."}]}'
```

Result:
- `stop_reason: tool_use` (initial turn) → `WebSearch` invoked with `{"query": "latest node.js version 2026"}`
- 4 content blocks: `thinking` → `tool_use(WebSearch)` → `thinking` → `text("**26**\n\nSources:\n- [Node.js 26.0.0 (Current)](https://nodejs.org/en/blog/release/v26.0.0)")`
- `WebSearch` ran, real result returned with citation, thinking blocks forwarded.

So the proxy / SDK / Anthropic side of the chain is verified. **Cherry Studio's actual UI rendering** of those blocks (citation hyperlinks, streaming, thinking display) can only be verified end-user-side.

## Verification asks before merge

@BenIsLegit / @rynfar — please:

1. Check out this branch (or wait for it to merge to a release).
2. Configure Cherry Studio's Anthropic provider with:
   - **API URL**: `http://127.0.0.1:3456`
   - **Custom request header**: `x-meridian-agent: cherry-studio`
3. Ask Cherry Studio something that should trigger web search — e.g. "what's the latest news about <recent topic>?"
4. Confirm:
   - [ ] Web search actually fires (Cherry Studio shows search results)
   - [ ] Sources/citations render correctly in the UI
   - [ ] No "WebSearch tool not exposed" error
   - [ ] Per-adapter toggles (Client Prompt, Thinking Passthrough, Thinking) are now visible at `/settings` under a "Cherry Studio" section
   - [ ] Streaming works if Cherry Studio uses streaming responses

If anything's off, comment on the PR and we'll iterate.

## Merge strategy

Squash merge. Single commit by @rynfar.

## Out of scope (follow-ups, not blocking this PR)

- Filesystem tool toggle for chat-client adapters (per-feature, like `additionalDirectories`).
- Claude Desktop adapter — same archetype, would alias to `cherry-studio`'s shape under a different name.
- Auto-detection via header sniffing if Cherry Studio adds a stable header in a future release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -537,6 +537,30 @@ Requests flow through the Claude Code adapter which:
 - Leaves the SDK subprocess cwd on the proxy host (Claude Code's local paths don't exist there).
 - Runs in passthrough mode by default — Claude Code executes its own tools on the machine it runs on; Meridian just forwards tool_use blocks.
 
+### Cherry Studio
+
+[Cherry Studio](https://github.com/CherryHQ/cherry-studio) is a desktop chat client (Electron) that talks to any Anthropic-compatible endpoint. It's the first **chat-client** adapter in Meridian — the existing adapters (OpenCode, Crush, Pi, ForgeCode, Claude Code, Droid) are all CLI coding agents with their own MCP tool runtimes, but Cherry Studio is a pure chat UI that wants Claude's server-side tools (especially `WebSearch` and `WebFetch`) to work natively.
+
+In Cherry Studio's provider settings, add a new Anthropic-compatible provider:
+
+- **API URL:** `http://127.0.0.1:3456`
+- **API Key:** any string when `MERIDIAN_API_KEY` is unset, or your key value otherwise
+- **Custom request headers:** `x-meridian-agent: cherry-studio`
+
+Cherry Studio doesn't send a stable User-Agent ([CherryHQ#10209](https://github.com/CherryHQ/cherry-studio/issues/10209)) so the header is required for the adapter to fire — without it, requests fall through to whatever `MERIDIAN_DEFAULT_AGENT` is (default OpenCode), which blocks `WebSearch`/`WebFetch` in favor of OpenCode's MCP equivalents and breaks Cherry Studio's web search.
+
+If Cherry Studio is the *only* tool pointing at this Meridian instance, you can skip the header and set the env var instead:
+
+```bash
+MERIDIAN_DEFAULT_AGENT=cherry-studio meridian
+```
+
+The Cherry Studio adapter:
+- **Allows** `WebSearch` and `WebFetch` (the whole point — chat clients have no MCP equivalent and need Claude's built-in web access).
+- **Blocks** filesystem and shell tools by default (`Read`, `Write`, `Edit`, `Bash`, `Glob`, `Grep`). A chat-style LLM shouldn't enumerate files on the proxy host unsupervised, even on localhost. If you want broader access, point at a coding-agent adapter where tool calls are surveilled by the calling tool.
+- Runs the SDK's tools internally (no passthrough) so results land inline in the assistant turn — exactly what a chat UI wants to render.
+- Supports `thinking` / `thinkingPassthrough` toggles via the settings UI at `/settings`, same as the coding-agent adapters.
+
 ### Any Anthropic-compatible tool
 
 ```bash
--- a/src/__tests__/cherry-studio-adapter.test.ts
+++ b/src/__tests__/cherry-studio-adapter.test.ts
@@ -1,0 +1,171 @@
+/**
+ * Tests for the Cherry Studio chat-client adapter.
+ *
+ * The load-bearing assertion below — that WebSearch / WebFetch are NOT in
+ * the adapter's blocked / incompatible lists — is the actual fix for the
+ * symptom Ben reported in #481 ("WebSearch tool not exposed in the
+ * session"). When this assertion regresses, Cherry Studio's web search
+ * silently breaks again.
+ */
+import { describe, it, expect } from "bun:test"
+import { cherryStudioAdapter } from "../proxy/adapters/cherrystudio"
+
+describe("cherryStudioAdapter — identity", () => {
+  it("has name 'cherry-studio'", () => {
+    expect(cherryStudioAdapter.name).toBe("cherry-studio")
+  })
+})
+
+describe("cherryStudioAdapter.getSessionId", () => {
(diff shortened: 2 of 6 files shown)


BenIsLegit:
Thanks! I'll try it out today

BenIsLegit:
Looks like it's still struggling and using the wrong tools to do web searches. 
image
image
image

Here's a separate attempt where I unselected "Web Search" in Cherry Studio and then asked it explicitly to do a web search:
image
image
image


I'll investigate a bit more and see why it's getting in a loop. Seems like it fails a Cherry Studio web search tool, silently completes it using CC web search, then commenting re-attempting and re-reporting

BenIsLegit:
Realized I had Thinking and Thinking Passthrough disabled. Turned those on, but it seems to really think it doesn't have web search tools available now
image

rynfar:
Yes this is an issue another user reported im hoping to find some time to
day to tackle this.

rynfar:
Superseded by #608 (merged). Same idea, but #608 fixes two gaps this branch never got past its "needs real-UI verification" hold:

1. **`getAllowedMcpTools()` returned `[]`** here — but the SDK's `allowedTools` is an *allowlist*, so unblocking WebSearch from the disallowed list isn't enough; it also has to be explicitly allowed. #608 returns `["WebSearch","WebFetch"]`, and I confirmed live that the search actually runs on the Max/OAuth path (grounded answer + source URLs).
2. **Internal `tool_use`/`thinking` leaked to the client** — with `supportsThinking: true` and no internal-tool hiding, the response carried `stop_reason: tool_use` plus a WebSearch block the chat client can't execute (it would loop). #608 hides these in both stream and non-stream, verified end-to-end through the real proxy.

Closing in favor of #608. Thanks past-me for the groundwork.

Issue #489, opened 2026-05-06
# Passthrough sends SDK built-in tool catalog (~21k tokens) to upstream Claude in v 1.42.0


albe-jj:
**Follow-up to #408.**

## Summary

PR #469 (commit `2d0130c`) fixed the config plumbing so `codeSystemPrompt: false` is now respected for the passthrough adapter. However, the **upstream request to Claude still includes the SDK's built-in tool catalog** (Read/Write/Edit/Bash/Glob/Grep/etc.), which is ~21k tokens of pure overhead per request.

The intent stated in #408 was to strip both the Claude Code system prompt **and** the built-in tool catalog. Only the system-prompt half landed on `main`.

## Reproduction (meridian 1.42.0)

```bash
UNIQUE=$(uuidgen)
curl -s http://localhost:3456/v1/messages \
  -H "x-api-key: dummy_key" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -H "x-meridian-agent: passthrough" \
  -d "{
    \"model\": \"claude-3-5-sonnet-20241022\",
    \"max_tokens\": 30,
    \"system\": \"tag-$UNIQUE\",
    \"messages\": [{\"role\": \"user\", \"content\": \"hi $UNIQUE\"}]
  }" | jq '.usage'
```

Observed:

```json
{
  "input_tokens": 3,
  "cache_read_input_tokens": 21436,
  "cache_creation_input_tokens": 1266
}
```

Total tokens going to Claude: **~22,700**. Actual user content: **<50 tokens**. The ~21k cached prefix is the SDK tool catalog being re-sent on every request.

## Root cause

`src/proxy/query.ts` passthrough branch sets `disallowedTools` but never strips the catalog:

```ts
...(passthrough
  ? {
      disallowedTools: [...allBlockedTools],   // runtime guard only
      // no `tools: []` — SDK still sends its built-in catalog upstream
    }
  : { ... })
```

`disallowedTools` blocks invocation at runtime; it does **not** remove tool definitions from the upstream request body.

## Proposed fix

Add `tools: []` to the passthrough branch in `query.ts`:

```ts
...(passthrough
  ? {
      tools: [],                              // strip SDK's built-in tool catalog
      disallowedTools: [...allBlockedTools],
      ...(passthroughMcp ? {
        allowedTools: [...passthroughMcp.toolNames],
        mcpServers: { [PASSTHROUGH_MCP_NAME]: passthroughMcp.server },
[...]

Issue #190, opened 2026-03-30
perf: skip claude_code preset system prompt in passthrough mode


xpolb01:
## Problem

In passthrough mode, every SDK call uses:

```typescript
systemPrompt: { type: "preset", preset: "claude_code", append: systemContext }
```

This loads the full Claude Code system prompt (~15-25K tokens) containing coding guidelines, built-in tool descriptions, and behavioral instructions — none of which are used in passthrough mode since the calling agent provides its own system prompt and handles all tool execution.

### Measured overhead

Captured from a real OpenCode session:

| Payload | Size | ~Tokens |
|---------|------|---------|
| Agent system prompt (from client) | 52,262 chars | ~13K |
| Tool schemas (46 tools) | 65,738 chars | ~16K |
| **Claude Code preset (unnecessary)** | **~60,000 chars** | **~15K** |

The preset adds **~15K tokens of dead weight per turn**. In passthrough mode with `maxTurns: 1`, every tool execution round-trip pays this overhead.

## Fix

In passthrough mode, pass the agent's system prompt directly instead of wrapping it in the `claude_code` preset:

```typescript
systemPrompt: passthrough
  ? systemContext  // raw — agent provides its own instructions
  : { type: "preset", preset: "claude_code", append: systemContext }
```

This preserves the full Claude Code preset for internal mode (where the SDK executes tools and needs those instructions) while eliminating the overhead in passthrough mode.

## Impact

- ~15K fewer input tokens per turn in passthrough mode
- No quality loss — the calling agent's system prompt already covers all necessary instructions
- Session resume continues to work (keyed by session ID, not system prompt content)
- All 507 tests pass with this change (including a new test for the behavior)

rynfar:
reviewing this

rynfar:
Addressed by #191 — merged in 98e8f9b. Thanks @xpolb01!

PR #526, opened 2026-06-05
fix(query): honour clientContext when codeSystemPrompt=false

ilie-neacsu:
## Problem

When `codeSystemPrompt` is explicitly set to `false` via
`PATCH /settings/api/features/:adapter`, the function `resolveSystemPrompt()`
returns `{ systemPrompt: "" }` unconditionally at line 208, discarding the
client's own system prompt before it is used.

Any non-coding API client (Open WebUI, curl, any OpenAI-compatible tool) that
disables the Claude Code preset via the settings API gets an empty system prompt —
the model ignores their instructions and falls back to the Claude Code persona
from Anthropic's prompt cache.

## Root Cause

The defensive empty-string return added in #489 fires before `clientContext`
is evaluated:

```typescript
// Before — clientContext is computed but never used in this branch
if (codeSystemPrompt === false) return { systemPrompt: "" }
```

## Fix

```typescript
// After — passes client system prompt through when present
if (codeSystemPrompt === false) return { systemPrompt: clientContext ?? "" }
```

The `?? ""` preserves the original defensive behaviour: when no system prompt
is provided, an explicit empty string prevents the SDK from reintroducing the
`claude_code` preset downstream.

## Test Results

All 55 existing assertions in `query.test.ts` pass with no modifications:

```
55 pass / 0 fail / 86 expect() calls
```

The existing test `forces systemPrompt='' when codeSystemPrompt false and no
systemContext` continues to pass because `clientContext ?? ""` evaluates to
`""` when `clientContext` is `undefined`.

## Reproduction

1. `PATCH /settings/api/features/opencode` with `{"codeSystemPrompt": false}`
2. Send any `/v1/chat/completions` request with a `system` message
3. Observe: `prompt_tokens: 3` (system prompt not counted), response reflects
   Claude Code persona instead of your system prompt

### DIFF
--- a/src/proxy/query.ts
+++ b/src/proxy/query.ts
@@ -205,7 +205,7 @@ function resolveSystemPrompt(
   // `systemPrompt` undefined and let downstream defaults reintroduce the
   // preset. (#489 follow-up — low impact in practice since most callers
   // send a `system` field; belt-and-suspenders for the empty case.)
-  if (codeSystemPrompt === false) return { systemPrompt: "" }
+  if (codeSystemPrompt === false) return { systemPrompt: clientContext ?? "" }
   return {}
 }
 


rynfar:
Thanks for digging into this @ilie-neacsu — the symptom you hit is **real and worth fixing** (generic OpenAI/Open WebUI clients getting the Claude Code persona instead of their own system prompt). But after tracing it end-to-end, this particular change is a **no-op**, and the client's system prompt is never actually discarded. Sharing the full evidence so the fix lands in the right place.

### Why the diff can't change behavior

`clientContext` is already folded into `append` a few lines above the line you edited:

```ts
const clientContext = includeClient ? systemContext : undefined
const append = [clientContext, cwdNote].filter(Boolean).join("") || undefined
...
if (append) return { systemPrompt: append }          // ← client prompt returns HERE when present
if (codeSystemPrompt === false) return { systemPrompt: "" }   // ← only reached when append is falsy
```

Line 208 is reachable **only when `append` is falsy**, which means `clientContext` is already empty/undefined. So `clientContext ?? ""` evaluates to `""` in every reachable case — identical to the current code. (The "55 pass / 0 fail, no test changes" result is the signature of a no-op: a real behavioral fix moves at least one assertion.)

### What's actually happening (deterministic trace, no live calls)

For an Open WebUI request `{system: "...", user: "..."}` through `/v1/chat/completions`:

| `codeSystemPrompt` | resulting `systemPrompt` sent to the SDK |
|---|---|
| `undefined` / `true` (what OpenAI traffic gets today) | `{preset: "claude_code", append: "<your prompt>"}` — **28KB Claude Code persona injected** |
| `false` | `"<your prompt>"` — clean, already correct |

The system message survives translation intact (`openai.ts` → `body.system` → `systemContext`); it is **never dropped**. The persona you saw is the `claude_code` *preset* being applied, with your prompt demoted to an `append` underneath it. (`prompt_tokens: 3` is that 28KB preset being served from cache, not a missing prompt.)

### Real root cause

`/v1/chat/completions` re-routes internally to `/v1/messages` with the User-Agent stripped, so the inner hop's `detectAdapter` falls to the default **`opencode`** adapter — whose `codeSystemPrompt` defaults to **`true`**. So generic chat clients get the coding preset by default, and there's no discoverable knob to turn it off (you'd have to know to configure "opencode").

### The fix we're shipping

A dedicated adapter identity for the OpenAI-compatible endpoint that defaults the preset **off** — same rationale/precedent as the `passthrough` adapter (#190: non-coding clients shouldn't be fed the 28KB coding preset), plus a configurable row in the settings UI for anyone who *does* want it on.

Closing this in favor of that fix — but genuinely, thank you for surfacing the behavior. Crediting your report in the follow-up. 🙏

PR #527, opened 2026-06-09
feat: add Fable 5 (claude-fable-5) model support

colto:
## Summary

Claude Fable 5 (`claude-fable-5`) launched after Meridian's last release, so `mapModelToClaudeModel` buckets fable requests into the default `sonnet` branch — requests for Fable silently answer as Sonnet, the same failure class as #419. This PR adds first-class fable routing, mirroring the existing opus handling.

## Changes

- **Model mapping** (`models.ts`): requests containing `fable` map to `fable[1m]` for primary agents and base `fable` for subagents, honoring the existing Extra-Usage cooldown — a structural mirror of the opus branch. `ClaudeModel` type and `stripExtendedContext` extended accordingly. Fable 5 supports the 1M context window (verified on a Max subscription — no Extra Usage error).
- **Model discovery** (`openai.ts`): `GET /v1/models` now advertises `claude-fable-5` (1M context for Max, 200k otherwise).
- **SDK default pin**: `ANTHROPIC_DEFAULT_FABLE_MODEL` is pinned to `claude-fable-5` at the SDK subprocess boundary alongside the existing opus/sonnet/haiku pins (claude-code ≥ 2.1.170 reads this env var — confirmed by inspecting the binary). Overridable via `MERIDIAN_DEFAULT_FABLE_MODEL` (proxy-side) or shell `ANTHROPIC_DEFAULT_FABLE_MODEL` (wins), and explicit `claude-fable-*` request IDs pin per-request, mirroring the `claude-opus-*` handling. Startup `Model pins:` log includes fable.
- **Dependency floor**: `@anthropic-ai/claude-code` raised `^2.1.117` → `^2.1.170`. The fable alias is rejected by earlier CLIs (2.1.119 returns "There's an issue with the selected model (fable[1m])"); 2.1.170 resolves it. `bun.nix` regenerated via `bun run nix:lock` for the bun-nix-sync workflow.
- **Docs**: README architecture map updated for `models.ts`.

## Verification

- `npm test`: 1750 pass / 0 fail, including new fable coverage mirroring the opus tests (mapping, subagent mode, cooldown, strip/has extended context, model list, pin injection, per-request pin, override precedence)
- `npm run typecheck`: clean
[...]

PR #530, opened 2026-06-10
feat(models): add Claude Fable 5 to model list and routing

StanChmielewski:
## What

Adds **Claude Fable 5** (`claude-fable-5`) to the proxy so OpenWebUI (and any OpenAI-compatible client) can list and use it.

## Why

Fable is the tier above Opus and has no `sonnet`/`opus`/`haiku` SDK alias, so the existing `mapModelToClaudeModel` collapse routes a `claude-fable-5` request to `sonnet` and it never appears on `/v1/models`. Max users on Fable currently can't reach it through Meridian.

## Changes

- `src/proxy/models.ts` — extend `ClaudeModel` with `claude-fable-5`; `mapModelToClaudeModel` passes the concrete id straight through (no alias to collapse to, so it's handed to the SDK directly — the Claude CLI resolves it).
- `src/proxy/openai.ts` — `buildModelList` advertises `claude-fable-5` (1M context on Max, 200k otherwise), matching the opus entries.
- Tests updated: model count 5→6, fable mapping + list assertions.

All `[1m]` rate-limit / extra-usage strip paths are no-ops for fable (`hasExtendedContext` is false), and `envOverrides` stays undefined (it only pins opus). No exhaustive `ClaudeModel` switches needed updating.

## Verification

Deployed on a Claude Max account and verified end-to-end: `claude-fable-5` lists on `/v1/models` and round-trips a chat completion (PONG), confirming Max serves Fable through the proxy.

### DIFF
--- a/src/__tests__/models.test.ts
+++ b/src/__tests__/models.test.ts
@@ -55,6 +55,12 @@ describe("mapModelToClaudeModel", () => {
     expect(mapModelToClaudeModel("", undefined)).toBe("sonnet")
   })
 
+  it("passes the concrete fable model straight through (no alias collapse)", () => {
+    expect(mapModelToClaudeModel("claude-fable-5")).toBe("claude-fable-5")
+    expect(mapModelToClaudeModel("claude-fable-5", "max")).toBe("claude-fable-5")
+    expect(mapModelToClaudeModel("fable")).toBe("claude-fable-5")
+  })
+
   it("respects explicit sonnet[1m] override when opted in", () => {
     process.env.CLAUDE_PROXY_SONNET_MODEL = "sonnet[1m]"
     expect(mapModelToClaudeModel("sonnet", "team")).toBe("sonnet[1m]")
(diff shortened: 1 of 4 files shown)


rs-prasad:
@StanChmielewski, could you take a look at the PR below? It addresses a similar issue, and I suspect we'll need to upgrade the claude-code version to enable Fable.

https://github.com/rynfar/meridian/pull/527/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

rynfar:
when fable support returns i will look into this thanks for submitting

rynfar:
Superseded by #561. We went with the `fable[1m]` alias approach (#527's design) now that the CLI floor is handled by #560, which gives Fable full opus-parity (1M on Max, subagent downgrade, Extra-Usage cooldown). Thanks for the quick concrete-passthrough fix — it was the right call before the CLI bump. Closing in favor of #561.

PR #540, opened 2026-06-17
refactor(nix): modernize flake and home-manager module

connor-grady:
Nix-only: fixes and cleanup to the flake distribution, with no change to the TypeScript or the proxy's runtime behavior.

## Fixes

- Replace the deprecated `pkgs.system` with `pkgs.stdenv.hostPlatform.system`.
- Stop the home-manager module from forcing the package onto `$PATH` via `home.packages`; it now only manages the systemd user service, and users who want the `meridian` CLI can add `config.services.meridian.package` to `home.packages` themselves.
- Use `Type = "exec"` so a failed startup is detected accurately.

## Cleanup

- Adopt `flake-parts`. It is already in the dependency closure (pulled in by `bun2nix`), so this adds no new dependency; it replaces the hand-rolled `eachSystem`/`pkgsFor` plumbing.
- Extract the package derivation into `nix/package.nix`, consumed via `callPackage`.
- The overlay and home-manager module use `withSystem` / `moduleWithSystem`, the flake-parts-native way to express the per-system package resolution they already did.
- Rebuild the `Environment` construction so adding a setting no longer requires editing the env block, plus general tidying (sorted inputs/outputs, aligned `follows`).
## home-manager input, for type checking

`home-manager` is added solely to expose the module through `home-manager.flakeModules.home-manager`, which types the output. Concretely it:

- validates the exported value is an actual module (`deferredModule`) at flake-eval time,
- tags it `_class = "homeManager"`, so importing it into a NixOS or nix-darwin config fails with a clear class-mismatch error instead of confusing "option does not exist" errors,
- attaches `_file` provenance so module errors point back at this flake.

## Notes

- `homeManagerModules` → `homeModules`: the README documented `homeManagerModules.default`, so I renamed to the modern `homeModules`, kept `homeManagerModules` as a deprecated alias, and updated the README. Happy to drop the alias for a clean break if you'd prefer.
- Env booleans now render as `true`/`false` instead of `1`/`0`, and `MERIDIAN_TELEMETRY_PERSIST` is always emitted; both are parsed identically by `src/env.ts`, so runtime behavior is unchanged.

## Testing

`nix flake check` passes; verified the package evaluates and the module's `Environment` resolves to the expected `KEY=value` list, including a user `environment` entry overriding a settings-derived variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Then in `~/.config/opencode/opencode.json`:
 
 # home-manager config
 {
-  imports = [ meridian.homeManagerModules.default ];
+  imports = [ meridian.homeModules.default ];
 
   services.meridian = {
     enable = true;
--- a/flake.lock
+++ b/flake.lock
@@ -49,6 +49,26 @@
         "type": "github"
       }
     },
+    "home-manager": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
+      "locked": {
+        "lastModified": 1781642113,
+        "narHash": "sha256-mAR7KTS9rjreTcXCNqfCbN96mnhJO8lDQq1vl7GviBQ=",
+        "owner": "nix-community",
+        "repo": "home-manager",
+        "rev": "df4e0465717a2d34f05b8ccd967275aaf3ceaa01",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "home-manager",
+        "type": "github"
+      }
+    },
     "import-tree": {
       "locked": {
         "lastModified": 1763762820,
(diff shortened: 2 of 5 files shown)


archer-65:
Nice cleanup 😄

A few improvements for follow up PRs that come to mind:
- The first login has to be done through `claude login`, what if a convenient way to use the bundled claude executable is provided to users?
- A `launchd` service created by the home-manager module would be cool, so macOS systems can make use of it.
- Packaging plugins like [opencode-scrub](https://github.com/rynfar/meridian-plugin-opencode-scrub) and configure them through the home-manager module.

connor-grady:
Thanks! Those are all good followups; I particularly like the `launchd` suggestion as a next step. But I'd rather not address them in this PR since it's already pretty big. Maybe you can open some issues so that they don't get lost?

rynfar:
i'll get to this shortly i am sorry for the delay

PR #544, opened 2026-06-21
fix: implement passthrough for the pi adapter

joematthews:
## Description

Noticed pi wasn't making proper tool calls through Meridian. Digging in, the pi adapter's passthrough was never actually hooked up: the transform never set the flag, so pi silently ran in internal mode and only the six built-in tools (`read`, `write`, `edit`, `bash`, `glob`, `grep`) reached the model.

This PR implements passthrough for the pi adapter (defaulting on, mirroring `opencode`/`claudecode`; opt out with `MERIDIAN_PASSTHROUGH=0`) and fixes the tests. With it, pi's full tool set reaches the model.

### 1. Pull this branch, build, and run the proxy

```shell
git clone -b fix/pi-passthrough-default https://github.com/joematthews/meridian.git
cd meridian
bun install
bun run build
bun dist/cli.js
```

### 2. Install pi, some web tools, and the Meridian extension

```shell
npm install -g @earendil-works/pi-coding-agent
pi install npm:pi-meridian-extension
pi install npm:pi-smart-fetch
pi install npm:pi-smart-web-search
```

The `pi-meridian-extension` registers a `meridian` provider, sends the `x-meridian-agent: pi` header automatically, and rewrites pi's system prompt so Anthropic's backend does not misidentify it as automated abuse. No manual `models.json` configuration is needed.

### 3. Run pi against Meridian and try a prompt that needs the web

```shell
pi --model meridian/claude-sonnet-4-6
```

Then ask for something current, e.g.:

> What are the three most interesting open-source TUI coding agents right now? Search the web and give me a one-line take on each.

**Expected:** pi actually runs `web_search` (and likely `web_fetch`/`batch_web_fetch`), then answers from real results instead of from memory. Before this PR, those tools were silently dropped because the pi adapter ran in internal mode.

### 4. Verify passthrough opt-out (optional)

Stop Meridian, restart with passthrough disabled, and confirm pi falls back to internal mode (only `read/write/edit/bash/glob/grep` available):

```shell
MERIDIAN_PASSTHROUGH=0 bun dist/cli.js
```

---

> **Note:** The raw `models.json` approach (overriding pi's `anthropic` provider to point directly at Meridian) currently triggers 400 errors because pi's default system prompt is incorrectly flagged by Anthropic's backend as abuse. The `pi-meridian-extension` works around this by rewriting the prompt. The `models.json` path should work in the future if Anthropic relaxes prompt-based filtering for human-driven agentic tools.

## Docs

This PR also includes a few small README fixes the change touches:

- The passthrough section now lists Pi alongside OpenCode as passthrough-by-default.
- The Pi section notes it runs in passthrough by default, with the opt-out.
- Corrected the Tested Agents row for Pi — it said `requires MERIDIAN_DEFAULT_AGENT=pi`, but detection is via the `x-meridian-agent: pi` header (the env var is just an alternative).

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ The system prompt controls are independent — any combination works:
 
 The core question is **who executes the tools** — the SDK or the client?
 
-- **Passthrough mode** (default for OpenCode) — Claude generates tool calls, but Meridian captures them and sends them back to the client for execution. The client runs the tool using its own implementation, with its own sandboxing, file tracking, and UI, then sends the result in the next request. This is how OpenCode, oh-my-opencagent (OMO), and most coding agents work — they have their own read/write/bash tools and need to stay in control of what runs on the user's machine.
+- **Passthrough mode** (default for OpenCode and Pi) — Claude generates tool calls, but Meridian captures them and sends them back to the client for execution. The client runs the tool using its own implementation, with its own sandboxing, file tracking, and UI, then sends the result in the next request. This is how OpenCode, oh-my-opencagent (OMO), and most coding agents work — they have their own read/write/bash tools and need to stay in control of what runs on the user's machine.
 - **Internal mode** — Claude Code handles everything. The SDK executes tools directly on the host, runs its full agent loop, and returns the final result. This is for clients that are purely chat interfaces (Open WebUI, simple API consumers) with no tool execution of their own.
 
 Most users don't need to configure anything — the adapter sets the right mode automatically. To override:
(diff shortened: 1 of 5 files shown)


joematthews:
Did a rebase to sign the commits. Hope that explains the force push.

joematthews:
@rynfar Do I need to do anything else to this PR? Any questions?

rynfar:
sorry for the delay i will look at this today - i've been travelling

PR #545, opened 2026-06-21
fix: strip Hermes Agent system-prompt boilerplate on the openai adapter

qinwang-ai:
## Problem

When a client forwards a system prompt that contains **Hermes Agent**'s
operational boilerplate (its self-description, tool/permission directives) over
`/v1/chat/completions` with `clientSystemPrompt: true`, the Claude Agent SDK
**silently returns an empty stream** — `data: [DONE]` with no content blocks —
which surfaces to the client as `empty stream with no finish_reason`.

The trigger is the **content**, not the size: generic filler text of the same
length (24 KB+) streams fine, while the real Hermes system prompt reliably
empty-streams. Bisection isolated it to a section of Hermes' boilerplate (the
SDK appears to try to parse directives embedded in it). The error is swallowed —
even with `sdkDebug` enabled, nothing is logged.

This forced operators onto `clientSystemPrompt: false` (dropping the persona
entirely), so the agent ran with no identity / language / behavior rules.

## Fix

Add an `openai`-adapter `onRequest` transform that truncates the forwarded
system prompt at the Hermes boilerplate anchor (`"You run on Hermes Agent"`),
keeping only the leading profile persona — which the SDK accepts and streams
normally. The persona we want lives in that leading block; everything after the
anchor is operational text the SDK-backed model doesn't need (it has its own
tool handling).

- Scoped to the `openai` adapter — `opencode` and other clients are untouched.
- No-op when the anchor is absent.

## Tests

`src/__tests__/openai-strip-hermes.test.ts` — 10 tests (truncation, persona
preservation, no-op cases, adapter scoping, registry wiring). Full transform/
openai suites pass.

### DIFF
--- a/src/__tests__/openai-strip-hermes.test.ts
+++ b/src/__tests__/openai-strip-hermes.test.ts
@@ -1,0 +1,86 @@
+import { describe, it, expect } from "bun:test"
+import {
+  runTransformHook,
+  createRequestContext,
+  type RequestContext,
+} from "../proxy/transform"
+import {
+  stripHermesBoilerplateTransform,
+  HERMES_BOILERPLATE_ANCHOR,
+  openAiTransforms,
+} from "../proxy/transforms/openai"
+import { getAdapterTransforms } from "../proxy/transforms/registry"
+
+function makeCtx(systemContext: string | undefined, adapter = "openai"): RequestContext {
+  return createRequestContext({
+    adapter,
+    body: {},
+    headers: new Headers(),
+    model: "opus",
+    messages: [],
+    systemContext,
+    stream: false,
+    workingDirectory: "/tmp",
+  })
+}
+
+const PERSONA = "# Bob — PublicAI CPO Agent\n\nYou are Bob, PublicAI's CPO.\n\n"
+const BOILERPLATE = `${HERMES_BOILERPLATE_ANCHOR} (by Nous Research). Load the hermes skill...\nYou are a CLI AI Agent.`
+
+describe("stripHermesBoilerplateTransform", () => {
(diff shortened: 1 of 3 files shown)


rynfar:
Thanks @qinwang-ai — this is a well-built PR for a real problem, and your bisection was right that the *content* is the trigger. We dug further and can now name the mechanism: the Hermes harness boilerplate matches Anthropic's coding-harness fingerprint and gets metered as **Extra Usage** — on accounts without extra-usage headroom the API rejects the request (`400 You're out of extra usage`), which surfaces through the OpenAI-compat stream as exactly the empty-stream-no-finish_reason you saw. Your truncate-at-anchor fix works because it removes the fingerprint.

We're not merging it into core, though: Meridian's rule is that content-mutating transforms live in external opt-in plugins, not `src/` — and there's an existing plugin solving this exact problem more surgically: [meridian-plugin-hermes-scrub](https://github.com/rynfar/meridian-plugin-hermes-scrub) strips just the `# Finishing the job` harness block while preserving everything else (including persona content *after* the boilerplate, which truncate-at-anchor would drop). We verified it live today against the current SDK: raw Hermes prompt → extra-usage 400; scrubbed → normal response.

The plugin wasn't published or documented until now — that's on us. It's listed in Meridian's README as of today, and the npm publish is landing shortly as `@rynfar/meridian-plugin-hermes-scrub`. Could you try it against your Hermes deployment and confirm it resolves your case? If your Hermes version trips on a different block, a PR against the plugin repo would be very welcome.

Closing this one — genuinely appreciate the quality of the investigation and tests.

PR #548, opened 2026-06-23
fix(thinking): honor explicit "disabled" setting over client requests

mwd102:
## Problem

Setting **Thinking → `disabled`** in the settings UI (`sdk-features.json`) is silently ignored whenever the client requests thinking itself.

In `server.ts`, the per-adapter `thinking` setting was only ever consulted as a *fallback default*:

```ts
if (!thinking) {
  if (sdkFeatures.thinking === "adaptive") thinking = { type: "adaptive" }
  else if (sdkFeatures.thinking === "enabled") thinking = { type: "enabled" }
  // no branch for "disabled"
}
```

Two consequences:

- **Anthropic `/v1/messages`:** a client `body.thinking` or `x-opencode-thinking` header makes `thinking` truthy *before* this block, so the setting is never consulted — "disabled" is overridden by the client (e.g. OpenCode's own thinking toggle).
- **`effort` is independent:** `effort` only tunes thinking depth but is forwarded to the SDK regardless of the thinking setting, so reasoning can still be triggered (notably via `reasoning_effort` on the OpenAI endpoint) even with thinking "disabled".

So the only thing that actually force-disabled thinking was the beta-stripped path; the UI setting did not.

## Fix

Make an **explicitly configured** `"disabled"` authoritative — it now hard-disables thinking and drops `effort` at the SDK boundary, mirroring the existing beta-stripped hard-disable:

```ts
if (getExplicitThinking(adapter.name) === "disabled") {
  thinking = { type: "disabled" }
  effort = undefined
} else if (!thinking) {
  ...adaptive/enabled defaults (unchanged)...
}
```

### Why a new `getExplicitThinking()` helper?

`DEFAULT_FEATURES.thinking` is itself `"disabled"`, so the *merged* value from `getFeaturesForAdapter()` can't distinguish "user explicitly chose disabled" from "default disabled". Those must behave differently:

- **explicit `"disabled"`** → authoritative, overrides the client.
- **default `"disabled"`** (user never touched it) → stays a no-op, so clients can still request thinking per-request (preserves current out-of-box behavior).

`getExplicitThinking()` reads the *raw* user config (key present vs absent) to tell them apart. `"adaptive"`/`"enabled"` remain fallback-only and are unchanged.

## Tests

New `src/__tests__/proxy-thinking-setting.test.ts`:

- explicit `"disabled"` overrides `body.thinking` and the `x-opencode-thinking` header
- explicit `"disabled"` drops `effort`
- explicit `"disabled"` disables thinking when the client sends nothing
- **regression guards:** default `"disabled"` (unset) leaves client `body.thinking` and `effort` untouched
- `"adaptive"`/`"enabled"` still defer to client-supplied thinking

Full suite (`npm test`) passes; `tsc --noEmit` clean.

## Notes

No existing issue tracked this. It's the same class as #408 (`passthrough` ignored `codeSystemPrompt: false`) — a settings toggle dropped at the resolution boundary.

One reviewer decision worth flagging: I made explicit `"disabled"` also drop `effort`, on the basis that effort's only documented role is thinking depth. If you'd rather keep `effort` independent of the thinking setting, that line is easy to remove.

### DIFF
--- a/src/__tests__/proxy-thinking-setting.test.ts
+++ b/src/__tests__/proxy-thinking-setting.test.ts
@@ -1,0 +1,170 @@
+/**
+ * The per-adapter `thinking` setting (sdk-features.json / settings UI) must be
+ * respected at the SDK boundary.
+ *
+ * Bug: an explicit "disabled" choice was ignored. The setting was only ever
+ * consulted as a fallback default (inside `if (!thinking)`), and "disabled" had
+ * no branch — so a client sending its own `thinking` (body or x-opencode-thinking
+ * header) overrode it, and `effort` (which only tunes thinking depth) still flowed
+ * through. An explicit "disabled" must now hard-disable thinking and drop effort,
+ * while the *default* "disabled" stays a no-op so clients can request thinking
+ * per-request.
+ */
+
+import { describe, it, expect, mock, beforeEach } from "bun:test"
+import { assistantMessage } from "./helpers"
+
+// ─── controllable sdk-features state ──────────────────────────────────────────
+// `explicitThinking` = the raw, user-configured value (undefined = unset/default).
+// `defaultThinking`  = the merged value getFeaturesForAdapter() returns.
+let explicitThinking: "adaptive" | "enabled" | "disabled" | undefined = undefined
+let defaultThinking: "adaptive" | "enabled" | "disabled" = "disabled"
+
+// Self-contained feature object (no delegation to the real module — once mocked,
+// the namespace IS the mock, so delegating would recurse infinitely).
+const FEATURES = {
+  codeSystemPrompt: true,
+  clientSystemPrompt: true,
+  claudeMd: "off" as const,
+  memory: false,
+  dreaming: false,
+  thinking: "disabled" as const,
+  thinkingPassthrough: false,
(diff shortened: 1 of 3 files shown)


rynfar (approved):
Solid, well-diagnosed fix. The raw-vs-merged distinction via `getExplicitThinking()` is the right mechanism, the OpenAI `reasoning_effort` path is covered (it re-dispatches through /v1/messages), and the test coverage — including regression guards for default-"disabled" — is thorough.

One natural follow-up (not a blocker): the beta-stripped hard-disable at server.ts:634 sets `thinking: disabled` but leaves `effort` intact, so it retains the same latent issue this PR fixes for the explicit path (effort forwards to the SDK independently at query.ts:311). I'll open a small follow-up to drop `effort` there too, building on the `let effort` you introduced here.

rynfar:
Recreated on current `main` as #562 (your fork branch had drifted — its `mock.module("../proxy/sdkFeatures")` leaked into sibling test files, failing the OpenAI system-prompt test on the merge-with-main). #562 reapplies your exact fix and tests, isolates the new test file, and credits you as co-author. Thanks @mwd102! Closing in favor of #562.

PR #554, opened 2026-07-01
fix: forward token usage on OpenAI-format streaming responses

shanemmattner:
## Problem

When a client requests `stream_options: { include_usage: true }` on `/v1/chat/completions` (standard OpenAI streaming spec), Meridian never emits the trailing usage-only chunk. The token counter in clients like Hermes Agent stays at 0.

The data is already available — Anthropic's `message_delta` SSE event includes a `usage` object with `input_tokens` and `output_tokens`, and Meridian's own `/v1/messages` endpoint emits it correctly. But `translateAnthropicSseEvent` never reads `event.usage`, and the route handler never parses `stream_options` from the request body. The `OpenAiChatRequest` type doesn't even declare the field.

Non-streaming works correctly because `translateAnthropicToOpenAi` already reads `response.usage` — this PR brings streaming to parity.

## Changes

- `src/proxy/openai.ts`: Add `stream_options` to `OpenAiChatRequest`, `usage` to `AnthropicSseEvent` and `OpenAiStreamChunk`, `includeUsage` to `SseTranslatorContext`. Modify `createSseTranslator` to capture `event.usage` from `message_delta` events and expose `buildUsageChunk()` (attached as a property on the returned function — fully backward-compatible, zero existing call sites change).
- `src/proxy/server.ts`: Parse `stream_options.include_usage` from the request, pass it to `createSseTranslator`, emit the trailing usage chunk before `[DONE]`.
- `src/__tests__/openai.test.ts`: 4 new tests covering include_usage not requested, no usage observed, populated chunk, and multiple message_delta events.

## Spec compliance

Per the [OpenAI streaming spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options):
- Usage chunk has `choices: []` (empty array)
- Fields: `prompt_tokens`, `completion_tokens`, `total_tokens`
- Emitted as the last data chunk before `data: [DONE]`
- Only emitted when `stream_options.include_usage` is explicitly `true`

### DIFF
--- a/src/__tests__/openai.test.ts
+++ b/src/__tests__/openai.test.ts
@@ -1149,6 +1149,34 @@ describe("createSseTranslator", () => {
     expect(chunks[4]!.choices[0]!.delta.tool_calls![0]!.index).toBe(0)
     expect(chunks[5]!.choices[0]!.finish_reason).toBe("tool_calls")
   })
+
+  it("buildUsageChunk returns null when includeUsage was not requested", () => {
+    const translate = createSseTranslator(CTX)
+    translate({ type: "message_delta", delta: { stop_reason: "end_turn" }, usage: { input_tokens: 10, output_tokens: 5 } })
+    expect(translate.buildUsageChunk()).toBeNull()
+  })
+
+  it("buildUsageChunk returns null when no message_delta.usage was observed", () => {
+    const translate = createSseTranslator({ ...CTX, includeUsage: true })
+    translate({ type: "message_delta", delta: { stop_reason: "end_turn" } })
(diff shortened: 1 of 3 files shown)


rynfar:
Thanks @shanemmattner — verified and merged. We confirmed the gap on our side (the streaming path never read `message_delta.usage` while non-streaming forwarded usage fine), and the tap point is exactly right given the internal `/v1/chat/completions` → `/v1/messages` hop. Clean spec compliance on the trailing chunk too (`choices: []`, gated on `include_usage === true`, before `[DONE]`). Merged with your authorship preserved — this will ship in the next release. Two solid contributions today, much appreciated!

Issue #558, opened 2026-07-01
Claude Sonnet 5 Falls through to Sonnet 4.6


thatMacAdmin:
When attempting to use Claude Sonnet 5 - it seems to fall back to Sonnet 4.6.

I tried adding the following to my Crush providers:
```json
{ "id": "claude-sonnet-5",   "name": "Claude Sonnet 5 (1M)",    "context_window": 1000000, "default_max_tokens": 32768, "can_reason": true, "supports_attachments": true }
```

But this does not work. I assume there is something you need to add on the meridian side?

tom-groves:
I've raised https://github.com/rynfar/meridian/pull/563

rynfar:
Fixed in v1.45.1 by #563 (thanks @tom-groves): explicit `claude-sonnet-*` / `claude-haiku-*` requests now pin `ANTHROPIC_DEFAULT_SONNET_MODEL`/`ANTHROPIC_DEFAULT_HAIKU_MODEL` for the SDK subprocess, so `claude-sonnet-5` is actually served by sonnet-5 instead of falling back to the canonical sonnet-4-6 pin. Verified live: pinning the alias env var provably controls which model serves the request (confirmed via the SDK result's modelUsage). Update to 1.45.1 and your Crush config will work as-is.

PR #563, opened 2026-07-02
fix: forward sonnet/haiku model requests via ANTHROPIC_DEFAULT_*_MODEL overrides

tom-groves:
## Bug

Requesting an explicit sonnet or haiku model (e.g. `claude-sonnet-5`) via the proxy does not actually route to that model. Meridian logs success with the intended model (`adapter=... model=sonnet`), but Claude Code reports itself as a different, older version at runtime. Using `claude` directly with the same model ID works correctly.

## Root cause

In `src/proxy/server.ts`, the `envOverrides` block that pins `ANTHROPIC_DEFAULT_{TYPE}_MODEL` for the SDK subprocess only special-cases `claude-opus-` and `claude-fable-` prefixes:

```js
const envOverrides = requestedModel.startsWith("claude-opus-")
  ? { ANTHROPIC_DEFAULT_OPUS_MODEL: requestedModel }
  : requestedModel.startsWith("claude-fable-")
    ? { ANTHROPIC_DEFAULT_FABLE_MODEL: requestedModel }
    : undefined
```

`mapModelToClaudeModel` (`src/proxy/models.ts`) collapses any `claude-sonnet-*`/`claude-haiku-*` request down to the generic `"sonnet"`/`"haiku"` SDK alias, and `resolveSdkModelDefaults()` fills in `ANTHROPIC_DEFAULT_SONNET_MODEL`/`ANTHROPIC_DEFAULT_HAIKU_MODEL` with Meridian's hardcoded canonical pin whenever no override wins. Since sonnet/haiku never produced an override, an explicit request (e.g. `claude-sonnet-5`) was silently downgraded to the canonical pin (`claude-sonnet-4-6`) at the SDK subprocess boundary.

This is the same class of bug fixed for opus/fable — just not extended to sonnet/haiku at the time.

## Fix

Extends the existing `envOverrides` branch to `claude-sonnet-` and `claude-haiku-` prefixes, mirroring the opus/fable pattern exactly. Opus/fable behavior is unchanged.

## Tests

- Added: explicit `claude-sonnet-5`/`claude-haiku-4-5` requests set their respective `ANTHROPIC_DEFAULT_*_MODEL` override, including beating an inherited shell env pin (mirrors existing opus coverage).
- Added: bare `"sonnet"` alias still falls through to the canonical/shell pin (no change to alias behavior).
- Adjusted two existing tests that asserted canonical-pin/shell-pin fallback using a request model (`claude-sonnet-4-5`) that, with this fix, now legitimately triggers an override — switched those specifically to the bare `sonnet` alias, which is what they intended to exercise.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `npm test` — full suite green, including new/adjusted cases in `proxy-env-stripping.test.ts`
- [x] `bun run build` — clean

## Closes

https://github.com/rynfar/meridian/issues/558

### DIFF
--- a/src/__tests__/proxy-env-stripping.test.ts
+++ b/src/__tests__/proxy-env-stripping.test.ts
@@ -216,7 +216,8 @@ describe("SDK model pin injection (fixes #419)", () => {
 
   it("injects Meridian's canonical model pins when no shell env is set", async () => {
     const app = createTestApp()
-    await post(app, BASIC_REQUEST)
+    // Bare alias: no envOverride kicks in, so the canonical pin is exercised.
+    await post(app, { ...BASIC_REQUEST, model: "sonnet" })
     expect(capturedQueryOptions.env.ANTHROPIC_DEFAULT_FABLE_MODEL).toBe("claude-fable-5")
     expect(capturedQueryOptions.env.ANTHROPIC_DEFAULT_OPUS_MODEL).toBe("claude-opus-4-8")
     expect(capturedQueryOptions.env.ANTHROPIC_DEFAULT_SONNET_MODEL).toBe("claude-sonnet-4-6")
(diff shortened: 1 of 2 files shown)


Femumme:
Please merge this - really annoying that we're unable to use sonnet 5 atm

rynfar:
I am sorry for the delay.  I am working on this now.

rynfar:
Thanks @tom-groves — excellent diagnosis and a textbook fix. We verified the full chain on our side: `mapModelToClaudeModel` does collapse explicit sonnet/haiku versions to bare aliases, and a live SDK test confirmed `ANTHROPIC_DEFAULT_SONNET_MODEL` fully controls what the alias resolves to (we pinned `sonnet` to a haiku model and the result's `modelUsage` proved the pin wins). Fun detail from that test: the CLI's own unpinned `sonnet` default is already `claude-sonnet-5`, so the canonical pin was downgrading explicit sonnet-5 requests exactly as #558 reported. Test adjustments were spot-on too. Merged with your authorship preserved — shipping in v1.45.1 shortly. Appreciate the contribution!

PR #565, opened 2026-07-06
feat: md-codebase context optimization with progressive disclosure

davifernan:
## Summary

- Adds `filterSystemContext` as an optional hook on `AgentAdapter` — applied in `server.ts` after the full system context is assembled
- Implements Progressive Disclosure for the OpenCode adapter: strips Git workflow, release, and troubleshooting sections from CLAUDE.md before forwarding (~500 tokens saved per request for coding tasks)
- Upgrades `stripWorkflowGuidance` to a proper section-aware stripper that removes entire section bodies (not just header lines) by tracking heading level
- Trims CLAUDE.md to ~80 lines following md-codebase best practice — full docs moved to `docs/ARCHITECTURE.md` and `docs/AGENT-GUIDELINES.md`

## Test plan

- [ ] `bun test src/__tests__/opencode-progressive-disclosure.test.ts` — 6 tests covering section-body stripping, boundary detection, and no-op passthrough
- [ ] Full test suite: `npm test` — 1644 pass, 0 fail
- [ ] Manual smoke: start proxy, send a coding request via OpenCode, verify Git/Release sections absent from forwarded system context in logs

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,9 +2,18 @@
 
 Project guidelines for AI agents working in this codebase.
 
+## Context Strategy
+
+This proxy implements md-codebase patterns (2026 best practice):
+- **CLAUDE.md is kept lean (~80 lines)** for stateless LLM sessions
+- **Task-specific guidance** lives in `docs/ARCHITECTURE.md` and `docs/AGENT-GUIDELINES.md`
+- **OpenCode adapter** applies Progressive Disclosure — strips workflow/release docs for coding tasks (~15-20% token reduction)
+
+See `docs/ARCHITECTURE.md` for the full module map, dependency rules, and design patterns.
+
 ## What This Is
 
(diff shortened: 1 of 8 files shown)


rynfar:
Thanks for the effort here, Davi — but this one changes Meridian's behavior in a way we can't ship. `filterSystemContext`/`stripWorkflowGuidance` runs on the *client's* system prompt, so it would silently delete "Git", "Releasing" etc. sections from **users' own project CLAUDE.md files** on every request — Meridian's own CLAUDE.md never flows through proxied requests, so the intended savings don't exist. Meridian's design principle is to be a clean proxy: content-mutating transforms live in external opt-in plugins (see meridian-plugin-hermes-scrub for the pattern), never unconditionally in core. If you'd like to pursue context slimming as an opt-in plugin with user-configured section filters, I'd be happy to look at that.

The docs reorganization also introduces some factual errors (OpenCode isn't Anthropic's built-in agent, passthrough doesn't involve OpenAI, `droid.ts` is Factory AI's Droid) and personal vault paths in the ADR, so I'd want those redone regardless.

Closing this one — appreciate you experimenting with the codebase.

PR #566, opened 2026-07-07
fix: remove MultiEdit from BLOCKED_BUILTIN_TOOLS

rarest:
## Problem

`MultiEdit` was merged into `Edit` in Claude Code 2.1.x and no longer exists as a distinct tool. Passing `--disallowedTools MultiEdit` to the Claude Code CLI now triggers:

```
Permission deny rule "MultiEdit" matches no known tool — check for typos.
```

This warning is harmless (does not block execution) but shows up on every proxied request that hits the tool-blocking path, adding log noise and occasionally surfacing in downstream error messages surfaced to users (e.g. mixed in with quota-exceeded errors, which is how I found it).

## Fix

Remove `MultiEdit` from `BLOCKED_BUILTIN_TOOLS` in `src/proxy/tools.ts` (and the mirrored list in the test file). `Edit` already covers multi-file/multi-edit operations in current Claude Code versions, so no functional blocking is lost.

## Testing

Verified against a live Meridian instance (v1.45.0) proxying Claude Code 2.1.202 — warning no longer appears after the fix.

### DIFF
--- a/src/__tests__/proxy-tool-blocking.test.ts
+++ b/src/__tests__/proxy-tool-blocking.test.ts
@@ -62,7 +62,7 @@ afterAll(() => {
 
 // The complete set of tools that must ALWAYS be blocked
 const BLOCKED_BUILTIN_TOOLS = [
-  "Read", "Write", "Edit", "MultiEdit",
+  "Read", "Write", "Edit",
   "Bash", "Glob", "Grep", "NotebookEdit",
   "WebFetch", "WebSearch", "TodoWrite"
 ]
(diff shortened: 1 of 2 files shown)


rynfar:
Thanks @rarest — confirmed on our side too: the bundled Claude Code (2.1.198) no longer registers MultiEdit as a tool, and the deny-rule warning was showing up in the stderr tail of every diagnostic log line (and occasionally inside user-facing error messages, exactly as you described). Clean fix, both lists updated. Merged with your authorship preserved — appreciate the contribution!

PR #570, opened 2026-07-08
fix: interrupt nested SDK session immediately on passthrough tool deny

shanemmattner:
## Bug

In passthrough mode, the `PreToolUse` hook's deny response for a
forwarded tool call is `{ decision: "block", reason: "..." }` with no
interrupt signal. The model is free to keep retrying or talking until the
(much larger) `maxTurns` cap eventually kills the session — recoverable
via the max_turns/aborted recovery paths, but slow, burning turns the
request didn't actually need.

## Fix

Add `interrupt: true` to the deny response in the `PreToolUse` hook
(`src/proxy/server.ts`) so the SDK ends the nested session's turn
immediately after the deny, letting recovery (streaming or
non-streaming) kick in on turn 1 instead of waiting for the turn cap.

## Caveat / composition note

This can change `extractSdkTermination()`'s inferred termination reason
from `"max_turns"` to an abort/interrupt shape (e.g. `"aborted"`). Both
recovery paths need to accept either reason for this to compose safely —
see the companion PR that adds `"max_turns" || "aborted"` acceptance to
the non-streaming recovery path (the streaming path already accepts
both). If a deployment's recovery path only checks for `"max_turns"`,
reverting this one file is safe — the max_turns-cap recovery path still
works, just slower.

## Validation

This fix has been running as a runtime monkey-patch against the built
bundle in a personal-assistant production deployment, alongside the
accepting recovery-path changes, with no regressions observed.

### DIFF
--- a/src/proxy/server.ts
+++ b/src/proxy/server.ts
@@ -958,8 +958,18 @@ export function createProxyServer(config: Partial<ProxyConfig> = {}): ProxyServe
                 // tool, burning the maxTurns budget. Be explicit that the call
                 // succeeded externally and that the model should stop here —
                 // see telemetry for the failure mode this addresses.
+                //
+                // `interrupt: true` asks the SDK to end the nested session's
+                // turn immediately after this deny, instead of leaving the
+                // model free to keep retrying/talking until the (much
+                // larger) maxTurns cap kills the session. Without it, a
+                // passthrough call that's already forwarded to the client
+                // can burn many turns before the max_turns/aborted recovery
+                // path (see the non-stream/stream recovery code) even gets
+                // a chance to kick in.
                 return {
                   decision: "block" as const,
+                  interrupt: true,
                   reason:
                     "This tool call has been forwarded to the client for execution. " +
                     "The result will be delivered in a future turn. " +


rynfar:
Thanks @shanemmattner — your diagnosis of the turn-burning deny behavior was right, and this PR prompted us to dig all the way to the bottom of it. 🙏

We couldn't take the `interrupt: true` approach as-is: the CLI's hook-output Zod schema has no `interrupt` key on the flat `decision`/`reason` shape, so it's silently stripped before the deny is processed (`interrupt` is only honored on `canUseTool`/PermissionRequest deny shapes — which is likely the path your bundle monkey-patch hit, and why it worked in your deployment). We A/B'd baseline vs `interrupt: true` vs `continue: false` through the real SDK and all three produced turn-for-turn identical traces, including in a chained-tool scenario that reproduces the burn.

The superseding fix landed in #575: since the proxy holds the SDK query's `AbortController` (#574), the deny hook now aborts it the moment loop-detection fires — the subprocess is killed mid-generation instead of draining the turn budget, and both recovery paths convert the abort into a clean `stop_reason: "tool_use"` response. You're credited in the commit (`Reported-by`). Closing this one in favor of #575 — thanks again for the contribution!

PR #573, opened 2026-07-10
fix: support native structured output

rynfar:
## What changed

- validates Anthropic `output_config.format` requests at the HTTP boundary
- forwards JSON Schema contracts through the Claude Agent SDK native `outputFormat` option
- returns the authoritative `structured_output` value for non-streaming calls
- buffers streaming output until schema validation and then emits a valid Anthropic SSE message
- fails clearly when the SDK omits a requested structured result

## Why

AI SDK structured generation can send `output_config.format`, but Meridian only forwarded `output_config.effort`. The ignored format forced clients onto the synthetic JSON-tool path, which could loop, exceed client timeouts, and produce unparseable output for complex schemas.

## Impact

Clients can use native JSON-schema output without custom submit-result tools or application-side parsing workarounds. Existing requests without `output_config.format` are unchanged.

## Validation

- `bun run test`
- `bun run typecheck`
- `bun run build`
- focused structured-output, SDK-parameter, and query-option suites: 50 passed

### DIFF
--- a/src/__tests__/proxy-structured-output.test.ts
+++ b/src/__tests__/proxy-structured-output.test.ts
@@ -1,0 +1,146 @@
+import { afterEach, beforeEach, describe, expect, it, mock } from "bun:test"
+
+let capturedOptions: Record<string, unknown> = {}
+let mockMessages: unknown[] = []
+let queryCalls = 0
+
+mock.module("@anthropic-ai/claude-agent-sdk", () => ({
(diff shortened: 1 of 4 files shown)


rynfar:
Added in the rebased push: requests combining `tools` with `output_config.format` are now rejected with a 400 `invalid_request_error` at the HTTP boundary — structured-output mode buffers wire events, so a tool_use turn would otherwise be swallowed and surface as an opaque 500.

PR #593, opened 2026-07-11
fix: resume Claude Code sessions after tool results

Ca134:
## Summary

- extract Claude Code's conversation ID from `metadata.user_id`
- resume the backing Claude SDK session after client-executed tool results
- preserve fingerprint resume when older Claude Code clients omit metadata
- keep the headerless client-loop behavior introduced by #571 for other adapters

## Root cause

Claude Code does not send a session-affinity header. Its session ID is embedded
as JSON in `metadata.user_id`, but the Claude Code adapter always returned
`undefined` from `getSessionId()`.

Since #571, requests whose final message contains `tool_result` and have no
recognized session ID are treated as independent. This caused normal Claude
Code tool-result turns to skip session lookup and start a new SDK session.

## Fix

The Claude Code adapter now extracts `session_id` from both supported
`metadata.user_id` shapes:

- JSON-encoded string
- object form

Claude Code tool-result requests are also excluded from the generic headerless
independent-session rule, preserving fingerprint resume for clients that omit
metadata.

## Validation

A real two-request Claude Code flow was tested against Meridian 1.45.3:

1. the first request returned `stop_reason=tool_use`
2. the second request returned the tool result with the same metadata session ID
3. Meridian logged `lineage=continuation`
4. the original Claude SDK session was resumed
5. the resumed request achieved a 99% cache hit rate

Tests:

- TypeScript typecheck passed
- production build passed
- 1765 main-suite tests passed
- metadata session resume regression passed
- fingerprint fallback regression passed
- existing passthrough single-turn tests passed

### DIFF
--- a/src/__tests__/claude-code-adapter.test.ts
+++ b/src/__tests__/claude-code-adapter.test.ts
@@ -19,21 +19,47 @@ describe("claudeCodeAdapter — identity", () => {
 })
 
 describe("claudeCodeAdapter.getSessionId", () => {
-  it("always returns undefined — Claude Code sends no session header", () => {
+  it("extracts a session ID from Claude Code's JSON-string metadata", () => {
     const ctx = {
       req: { header: () => "any-value" },
     }
-    expect(claudeCodeAdapter.getSessionId(ctx as any)).toBeUndefined()
+    const body = {
+      metadata: {
+        user_id: JSON.stringify({
+          device_id: "device-1",
+          account_uuid: "",
+          session_id: "session-from-metadata",
+        }),
+      },
+    }
+    expect(claudeCodeAdapter.getSessionId(ctx as any, body)).toBe("session-from-metadata")
   })
 
-  it("returns undefined even when x-opencode-session is present", () => {
+  it("accepts object-form metadata for compatible gateways", () => {
+    const ctx = { req: { header: () => undefined } }
+    const body = { metadata: { user_id: { session_id: "object-session" } } }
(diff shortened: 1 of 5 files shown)


rynfar (approved):
Approving — thanks for this, and for the thorough write-up. I verified the load-bearing parts independently since this touches the session-resume machinery:

- **Confirmed the premise empirically**: pointed the real `claude.exe` at a capture server; it sends `metadata.user_id` as a JSON string containing `session_id`, and the same `session_id` on every request in a conversation — a valid, stable session key. `extractClaudeCodeSessionId` handles both the JSON-string and object forms.
- **#580 protection intact**: the `storeSession` gate (`!isIndependentSession && !sawDuplicateToolUse`) is untouched, so single-step-aborted sessions still can't be resumed.
- **Divergence net intact**: resume still flows through `lookupSession(..., body.messages, ...)`, so history coherence is verified even when keyed by the metadata session ID — this makes Claude Code behave like OpenCode-with-a-session-header, a well-tested path.
- Full suite green locally (1870 pass), typecheck clean.

Two non-blocking follow-ups for later:
1. The `adapter.name !== "claude-code"` check is agent-specific logic in `server.ts`; an adapter capability flag (e.g. `wantsResumeAfterToolResult()`) would be cleaner long-term.
2. The line-745 "undo → diverged" safety downgrade is opencode-only; for the (defensive) Case-B path where Claude Code omits metadata, making that downgrade adapter-agnostic for the no-session case would close a very narrow history-leak edge. In practice `claude-cli` always sends metadata, so it's belt-and-suspenders.

Merging.

PR #595, opened 2026-07-11
feat: add workspace authority control primitive

nickembrey:
## Summary

Add a disabled-by-default workspace authority primitive for trusted local orchestrators.

- add a bounded, expiring in-memory lease registry that mints opaque capabilities bound to owner, agent, and absolute cwd
- add an authenticated newline-framed Unix control socket for issue, renew, revoke, and agent cleanup operations
- use a Meridian-generated runtime credential, exact `0700` directory validation, and `0600` socket/credential permissions
- make startup and shutdown transactional across the HTTP and control listeners, including duplicate/live socket protection
- expose opt-in CLI configuration through `MERIDIAN_WORKSPACE_AUTHORITY_SOCKET`

This PR intentionally does **not** consult capabilities during request cwd resolution. With the feature unset, no registry or control socket exists; with it enabled, existing HTTP/cwd behavior is still unchanged. Capability enforcement is a separate follow-up.

## Security and lifecycle properties

- live sockets are probed and never unlinked by a duplicate starter
- stale runtime artifacts are replaced only when owned by the current user
- capabilities expire, cannot be renewed or revoked across owner/agent bindings, and cannot be replayed after revoke/expiry
- registry storage is bounded and expired entries are pruned before admission
- control frames are size-limited and concurrent clients remain isolated
- child sockets, leases, credential files, Unix sockets, and the HTTP listener are cleaned up on failure/shutdown paths
- credentials and capabilities are not logged or placed in process arguments/environment values

## Test plan

- `bun run typecheck`
- `bun test` (1888 pass, 1 pre-existing skip, 0 fail)
- `bun run build`
- independent Codex review loop completed with no remaining findings


nickembrey:
Closing at the author's request; this work was opened on the wrong publication surface.

PR #596, opened 2026-07-13
feat(models): route Claude Mythos 5 (claude-mythos-*) through the fable tier

davidj4tech:
## Problem

`mapModelToClaudeModel` recognizes `haiku`/`fable`/`opus` substrings and defaults everything else to the sonnet tier. Since #561 added Fable 5, the one remaining current-generation id that still falls through is **Claude Mythos 5** (`claude-mythos-5`, Project Glasswing — same underlying model, context window, and API surface as `claude-fable-5`, different id). A client configured with a mythos id today is silently answered by sonnet — the same failure mode as #419 (opus-* requests silently answering as sonnet-4), just for a different tier.

## Fix

Mythos rides the fable tier, since the Claude Agent SDK has no separate `mythos` alias:

- **`src/proxy/models.ts`** — `mapModelToClaudeModel` now treats `mythos` like `fable`: primary agents get `fable[1m]`, subagents get base `fable`, and the shared Extra Usage cooldown / `MERIDIAN_1M_CONTEXT_SUPPORT` downgrade logic applies unchanged.
- **`src/proxy/server.ts`** — explicit `claude-mythos-*` ids pin `ANTHROPIC_DEFAULT_FABLE_MODEL` to the requested id, mirroring the existing opus/fable/sonnet/haiku verbatim-passthrough pattern, so the concrete model id reaches the API instead of collapsing to the canonical fable pin.

Users whose Claude auth can't serve `claude-mythos-5` (it's gated to Project Glasswing orgs) get a clear upstream error rather than a silent sonnet response, and can use `fable` / `claude-fable-5` instead. I deliberately did **not** add `claude-mythos-5` to `buildModelList` in `openai.ts` — advertising a gated model in UI pickers seemed misleading; happy to add it if you'd prefer parity.

## Tests

- `models.test.ts`: new `mythos 5` describe block mirroring the fable coverage (primary → `fable[1m]`, subagent → `fable`, Extra Usage cooldown and `MERIDIAN_1M_CONTEXT_SUPPORT=0` downgrades).
- `proxy-env-stripping.test.ts`: explicit `claude-mythos-5` requests pin `ANTHROPIC_DEFAULT_FABLE_MODEL=claude-mythos-5` through the HTTP layer.

Full suite (`npm test`, all 11 bun invocations): 1876 pass, 0 fail.

### DIFF
--- a/src/__tests__/models.test.ts
+++ b/src/__tests__/models.test.ts
@@ -143,6 +143,33 @@ describe("mapModelToClaudeModel", () => {
     })
   })
 
+  describe("mythos 5 (rides the fable tier)", () => {
+    afterEach(() => {
+      resetExtendedContextUnavailable()
+      delete process.env.MERIDIAN_1M_CONTEXT_SUPPORT
+    })
+
+    it("maps mythos to fable[1m] for primary agents instead of falling through to sonnet", () => {
+      expect(mapModelToClaudeModel("claude-mythos-5")).toBe("fable[1m]")
+      expect(mapModelToClaudeModel("mythos")).toBe("fable[1m]")
+      expect(mapModelToClaudeModel("claude-mythos-5", "max", "primary")).toBe("fable[1m]")
+    })
+
+    it("gives subagents base fable", () => {
+      expect(mapModelToClaudeModel("claude-mythos-5", "max", "subagent")).toBe("fable")
+    })
+
+    it("downgrades to base fable during the Extra Usage cooldown", () => {
+      recordExtendedContextUnavailable()
+      expect(mapModelToClaudeModel("claude-mythos-5", "max")).toBe("fable")
+    })
+
+    it("downgrades to base fable when MERIDIAN_1M_CONTEXT_SUPPORT=0", () => {
+      process.env.MERIDIAN_1M_CONTEXT_SUPPORT = "0"
+      expect(mapModelToClaudeModel("claude-mythos-5", "max")).toBe("fable")
+    })
+  })
+
   describe("subagent mode", () => {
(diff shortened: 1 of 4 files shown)


rynfar (approved):
Approving — thanks for this, and for the clear write-up and the #419 framing. I verified it end to end:

- **Premise confirmed**: Mythos 5 (`claude-mythos-5`, Project Glasswing) shares Fable 5's underlying model and API surface, and the Agent SDK has no separate `mythos` alias — so routing through the fable alias + pinning `ANTHROPIC_DEFAULT_FABLE_MODEL` to the concrete id is exactly right.
- **Logic checks out**: check order is haiku → fable(‖mythos) → opus → sonnet; `mythos` is a unique substring with no collision, `supports1mContext` correctly returns true for `claude-mythos-5`, and it reuses the fable `use1m`/subagent/Extra-Usage-cooldown/`MERIDIAN_1M_CONTEXT_SUPPORT` logic verbatim. The `server.ts` pin slots into the existing passthrough chain, ordered so mythos can't shadow fable.
- **Fails safe**: users without Glasswing access get a clear upstream error instead of a silent sonnet answer — strictly better than today.
- Merges cleanly with current main; full suite green (1875 pass, 0 fail); tsc clean.

Agree with leaving `claude-mythos-5` out of `buildModelList` — advertising a gated model in every user's picker would mislead the majority without access.

The only thing that's inherently untestable is the live SDK resolution of `fable[1m]` + a mythos pin (needs Glasswing access), but it's structurally identical to the proven fable path and the failure mode is a clear error, not a silent wrong-model. Safe to merge. Thanks again!

PR #623, opened 2026-07-15
feat: env-configurable plugin loading and home-manager plugin settings

connor-grady:
`ProxyConfig.pluginDir` and `pluginConfigPath` were only settable through the embedded `createProxyServer` API — the CLI never read them, so a standalone `meridian` process always used the fixed `~/.config/meridian/plugins/` and `~/.config/meridian/plugins.json`. `bin/cli.ts` now reads `MERIDIAN_PLUGIN_DIR` and `MERIDIAN_PLUGIN_CONFIG` and forwards them into `startProxyServer` (documented in `--help`), so both paths are configurable for the standalone proxy.

The home-manager module gains two settings. `services.meridian.settings.pluginConfig` takes a list of `{ path, enabled }` submodules and renders them, in order, to a `plugins.json` manifest in the Nix store (via `pkgs.formats.json`), passed as `MERIDIAN_PLUGIN_CONFIG`. `services.meridian.settings.pluginDir` maps to `MERIDIAN_PLUGIN_DIR`. The service `Environment` now derives each `MERIDIAN_*` variable name automatically from its `settings` attribute path, so new settings are exported without a hand-maintained key mapping.

### DIFF
--- a/bin/cli.ts
+++ b/bin/cli.ts
@@ -38,6 +38,8 @@ Environment variables:
   MERIDIAN_HOST                     Host to bind to (default: 127.0.0.1)
   MERIDIAN_PASSTHROUGH              Enable passthrough mode (tools forwarded to client)
   MERIDIAN_IDLE_TIMEOUT_SECONDS     Idle timeout in seconds (default: 120)
+  MERIDIAN_PLUGIN_DIR               Plugin auto-discovery directory (default: ~/.config/meridian/plugins)
+  MERIDIAN_PLUGIN_CONFIG            Plugin manifest path (default: ~/.config/meridian/plugins.json)
 
 See https://github.com/rynfar/meridian for full documentation.`)
   process.exit(0)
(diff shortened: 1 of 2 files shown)


rynfar:
Landed in #641 with your three commits cherry-picked verbatim (authorship preserved) — thank you @connor-grady, this closes a real gap for standalone/Nix deployments.

Two fix-ups went in on top, worth knowing about for your own config:

1. **`pluginConfig`'s empty default no longer renders a manifest.** As submitted, `apply = mkPluginFile` ran for the default `[]` too, so `MERIDIAN_PLUGIN_CONFIG` was always exported and pointed at an empty store manifest — silently overriding `~/.config/meridian/plugins.json` for every existing home-manager user. Empty now maps to `null` (env var not set).
2. **The env-name derivation now uses only long-stable `lib` functions.** `mapAttrsToListRecursive` / `splitStringBy` / `join` are 2025+ nixpkgs additions, and nothing in our CI evaluates the HM module — consumers on older nixpkgs would have hit eval errors we'd never see. Same behavior, no version floor (verified by evaluating the module via `lib.evalModules` in a container: defaults export no plugin config; every derived `MERIDIAN_*` name checks out).

Also documented both env vars in the README config table and added a `pluginConfig` example to the home-manager section.

PR #624, opened 2026-07-15
fix: expose model-scoped usage limits

thalysguimaraes:
## Summary

- parse Anthropic's generic `limits` entries for model-scoped weekly quotas
- expose those entries as the existing normalized window shape (for example, `seven_day_fable`)
- show Fable's quota with the short `7d Fable` label on the profile page

## Why

Anthropic now returns Fable-only usage in a `weekly_scoped` entry under `limits[]`. Meridian currently reads only the older top-level window fields, so `/v1/usage/quota` and `/v1/usage/quota/all` omit that quota even though the upstream response contains it.

The parser ignores session, aggregate weekly, and non-model scoped entries because Meridian already handles the aggregate windows through the existing top-level fields. It also de-duplicates normalized window types to preserve current behavior if Anthropic overlaps response shapes.

## Checks

- `npm test`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/src/__tests__/oauth-usage.test.ts
+++ b/src/__tests__/oauth-usage.test.ts
@@ -85,6 +85,53 @@ describe("oauthUsage", () => {
     expect(sevenDay!.utilization).toBeCloseTo(0.05, 5)
   })
 
+  test("parses model-scoped weekly limits without duplicating aggregate windows", async () => {
+    const resetsAt = "2026-07-20T12:00:00Z"
+    const fetchImpl = fixedFetch(() => new Response(JSON.stringify({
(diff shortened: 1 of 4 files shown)


rynfar:
Merged via #636 — your commit landed as-authored (546d86cb), thanks! Verified live against a real Max account before merging: Anthropic is indeed returning the Fable-scoped weekly limit in `limits[]`, and with your parser it now surfaces as `seven_day_fable` (9% used on the test account — previously invisible). Landed together with #629's cost estimation as one usage-visibility batch. Ships in the next release.

Issue #477, opened 2026-05-03
Security: /settings/api/* bypasses MERIDIAN_API_KEY.


SilverResort:

rynfar:
thanks for reporting this, let me investigate @SilverResort

PR #629, opened 2026-07-15
feat(telemetry): cost estimation cards with configurable model pricing

kilabyte:
## Summary

Adds cost estimation across Meridian's UI: what the window's token usage would have cost at Anthropic API list prices. Since Claude Max covers actual usage, the figure is framed everywhere as an equivalent-cost estimate, not a charge; it answers "how much value is the subscription delivering?"

### Dashboard (Overview tab)
- New **Estimated Cost** section: headline card with the window total, per-model cards, and a breakdown table (requests, input / output / cache-read / cache-write tokens, est. USD per model)
- Respects the existing window selector and auto-refresh

### Landing page
- **Est. Cost (24h)** card in the stat row next to Median TTFB and Error Rate

### Settings page
- New **Model Pricing** section: every built-in model's rates are editable (an edit becomes an override with a Reset action), and models missing from the built-in table can be added with custom rates
- Overrides persist to `~/.config/meridian/model-pricing.json` (relocatable via `MERIDIAN_PRICING_CONFIG`), are read with a short cache so changes apply without a restart, and win over the built-in table

## Design

- **Server-side computation.** Cost is a new `costEstimate` field on `TelemetrySummary`, computed in the shared `computeSummary`, so the memory and SQLite stores stay identical and any window size works via `GET /telemetry/summary`. No extra client fetches.
- **`src/telemetry/pricing.ts` (pure).** Static rate table verified against the official pricing docs (July 2026), with family fallback so versioned, dated, and `[1m]` model IDs resolve (e.g. `claude-haiku-4-5-20251001`, `opus[1m]`). Cache reads price at 0.1x input; cache writes at the 5-minute-TTL rate of 1.25x input (the TTL the SDK uses per MONITORING.md). `claude-sonnet-5` carries its introductory rate ($2/$10 through 2026-08-31, noted in code).
- **Unknown models are excluded from totals** and flagged "no pricing" in the UI rather than silently counted as $0; the settings page is the fix for those.
- **`src/telemetry/pricingStore.ts` (stateful).** Persistence mirrors `sdkFeatures.ts` (atomic tmp+rename writes, 5s read cache, read-at-request-time).
- **New endpoints** `GET /settings/api/pricing`, `PUT/DELETE /settings/api/pricing/:model` sit under the existing `/settings/*` auth boundary, so the #477 `MERIDIAN_API_KEY` protection applies automatically. All changes are additive; nothing in the Stable API Contract is modified.

## Testing

- 33 new unit tests (`pricing-unit.test.ts`, `pricing-store-unit.test.ts`): resolution and family fallback, override precedence, cache-rate derivation, aggregation math, validation, and a persistence roundtrip via a temp config path
- Full suite passes (2028 tests) and typecheck is clean
- Verified in-browser against a locally served landing page, dashboard, and settings page: editing a built-in rate produced an override reflected in the summary; Reset reverted it; adding pricing for an unknown model priced it immediately and cleared the unpriced-request note

## Docs

- `MONITORING.md`: new Estimated Cost section, including the override workflow and `MERIDIAN_PRICING_CONFIG`
- `ARCHITECTURE.md`: `pricing.ts` and `pricingStore.ts` added to the telemetry module map

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -69,6 +69,8 @@ src/
 │   ├── routes.ts              ← Telemetry API endpoints
 │   ├── logStore.ts            ← Diagnostic log ring buffer
 │   ├── dashboard.ts           ← HTML dashboard
+│   ├── pricing.ts             ← Static API list prices + cost estimation (pure)
+│   ├── pricingStore.ts        ← User pricing overrides (persisted JSON)
 │   ├── profileBar.ts          ← Shared profile switcher bar (injected into HTML pages)
 │   ├── profilePage.ts         ← Profile management page HTML
 │   └── types.ts               ← Telemetry types
--- a/MONITORING.md
+++ b/MONITORING.md
@@ -34,6 +34,27 @@ The SDK uses 5-minute TTL prompt caching. On each turn, the system prompt + tool
 
 Meridian sorts tools alphabetically before registration to keep the prefix stable. If you see cache misses on continuations, something is changing the prefix between turns.
 
+## Estimated Cost
+
+The Overview tab shows an **Estimated Cost** section: a headline card with the window's total, per-model cards, and a table breaking down input/output/cache-read/cache-write tokens with the estimated USD per model. The same data is available on the API as `costEstimate` in `GET /telemetry/summary`.
+
+What the number means:
+
+- It is the **equivalent Anthropic API list price** for the window's token usage. Claude Max requests are covered by the subscription and are never billed per token. Use it to gauge how much value the subscription is delivering, not as a bill.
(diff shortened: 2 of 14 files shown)


rynfar:
Thanks for providing this.  I will review this!

kilabyte:
im currently using it myself and has been great! its obvs just an estimate not meant to give actual spend but more to show dollar usage against my prepaid Max plan

rynfar:
Merged via #636 — your commit landed as-authored (afd3eda8), thank you for an exceptionally well-built contribution. The pricing table survived a rate-by-rate audit against current list prices (including the Sonnet 5 intro pricing catch — nice), and the architecture followed the house patterns exactly. Three small hardening changes were applied on top in #636: malformed-JSON PUT now 400s instead of 500ing, model names are HTML-escaped in the dashboard (requestModel is client-supplied — also fixed the same pre-existing gap in the byModel cards), and route-level tests were added for the three endpoints. Ships in the next release.

PR #635, opened 2026-07-16
feat(nix): package plugins as a meridianPlugins set

connor-grady:
Packages the scrub plugins centrally in the meridian flake so consumers pull them from the store instead of vendoring each one. Adds `meridian-plugin-hermes-scrub`, `meridian-plugin-opencode-scrub`, and `meridian-plugin-pi-scrub` as `flake = false` inputs (pinned to each repo's default branch) and builds them into a `legacyPackages.meridianPlugins.<name>` set.

## Packaging

A shared `nix/plugin.nix` builder compiles each plugin with `tsc` and exposes `passthru.path`, the absolute path to the loadable `index.js` a `plugins.json` entry points at. The set is derived from the `meridian-plugin-*` inputs by prefix and exposed through the `meridian` overlay alongside `meridian`, so a consumer references `pkgs.meridianPlugins.opencode-scrub.path` and feeds it to a `plugins.json` entry (or the `pluginConfig` option from #623).

## Keeping inputs fresh

`.github/workflows/update-plugins.yml` runs daily and on `workflow_dispatch`: it bumps every `meridian-plugin-*` input, rebuilds each plugin to gate the commit, and auto-commits `flake.lock` only when all builds pass. It installs upstream Nix via `cachix/install-nix-action` and enumerates inputs from the lock, so it needs no edits when a plugin is added.

## Follow-up for @rynfar: event-driven updates

Daily cron is a stopgap, and only you can replace it. The correct trigger is event-driven: each plugin repo fires a `repository_dispatch` at this repo on push to its default branch, so a bump happens immediately instead of up to 24h later. That wiring lives in the plugin repos you own and needs a cross-repo token in each plugin repo's secrets, since the default `GITHUB_TOKEN` can't dispatch across repos. Once it's in place it fully replaces the cron in this PR. Happy to draft the trigger and the dispatch step for you.

### DIFF
--- a/.github/workflows/update-plugins.yml
+++ b/.github/workflows/update-plugins.yml
@@ -1,0 +1,27 @@
+name: Update plugin inputs
+
+on:
+  schedule:
+    - cron: "0 6 * * *"
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v31
+
+      - name: Bump plugin inputs and verify each still builds
+        run: |
+          for input in $(nix flake metadata --json | jq -r '.locks.nodes.root.inputs | keys[] | select(startswith("meridian-plugin-"))'); do
+            nix flake update "$input"
+            nix build ".#meridianPlugins.${input#meridian-plugin-}" --no-link --print-out-paths
+          done
+
+      - uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: "chore: update plugin flake inputs"
+          file_pattern: flake.lock
--- a/flake.lock
+++ b/flake.lock
@@ -84,6 +84,54 @@
         "type": "github"
       }
     },
+    "meridian-plugin-hermes-scrub": {
+      "flake": false,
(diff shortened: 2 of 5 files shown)


rynfar:
will get this reviewed today thank you!

rynfar:
Landed in #649 with your three commits cherry-picked verbatim (authorship preserved) — thank you @connor-grady. This completes a really nice arc with #623: inputs → builder → `meridianPlugins` set → `pluginConfig`, all composing.

Verification before landing: confirmed all three plugin repos are dependency-free ESM with `outDir: dist` (so the tsc-only builder's assumptions hold), and built `.#meridianPlugins.pi-scrub` from the branch in a nix container — clean build, `passthru.path` resolves to the store `lib/index.js`.

One fix-up went in on top: the update workflow pushed `flake.lock` straight to main via `git-auto-commit-action` — this repo is PRs-only and branch protection rejects direct pushes, so the auto-commit would have failed every run. It now opens an idempotent PR via `peter-evans/create-pull-request`, with the per-plugin rebuilds still gating the bump before the PR opens.

Your `repository_dispatch` follow-up is tracked in a dedicated issue so it doesn't get lost.

connor-grady:
Awesome! And makes sense for the workflow fixup; noted for any future potential PRs.

Issue #501, opened 2026-05-08
Running embedded claude executable fails on NixOS


IwailerI:
Issue specific to NixOS. Verified with both user systemd service and a manually started system package.
Using latest (v1.42.1 currently)

All queries result in the following error:

```
e11d56c8-022f-4f84-a169-54deab771463 sdk_termination reason=process_exit exit=127 model=sonnet resume=false deferred=false stderr="Could not start dynamically linked executable: /nix/store/pxvw9hzrbh6x309nbxz0j37d39az4n1s-meridian-1.42.1/lib/meridian/node_modules/@anthropic-ai/claude-code/bin/claude.exe\nNixOS cannot run dynamically linked executables intended for generic\nlinux environments out of the box. For more information, see:\nhttps://nix.dev/permalink/stub-ld"
```

Unescaped stderr:

```
Could not start dynamically linked executable: /nix/store/pxvw9hzrbh6x309nbxz0j37d39az4n1s-meridian-1.42.1/lib/meridian/node_modules/@anthropic-ai/claude-code/bin/claude.exe
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
```

Any help greatly appreciated.

dzorko179:
This is a NixOS issue, not a Meridian/Claude bug.

The bundled claude.exe is actually a Linux binary, but it’s dynamically linked and expects a normal Linux filesystem + dynamic loader setup. On NixOS this isn’t available by default, which is why you get the stub-ld error.

Fix is to enable nix-ld:

sudo nano /etc/nixos/configuration.nix

Add:

programs.nix-ld.enable = true;

Then apply:

sudo nixos-rebuild switch

Restart Meridian after that — it should run fine.

rynfar:
really sorry for the delays on this, i will addresst his

rynfar:
Fixed properly in #612 (merged) — no more nix-ld required.

**Root cause:** the Nix package skipped fixup entirely (`dontFixup = true`), so the vendored `claude.exe` — a dynamically linked Linux ELF the SDK spawns for every query — kept its generic `/lib64/ld-linux-*` interpreter, which doesn't exist on NixOS → `exit=127` on everything.

[...]

PR #656, opened 2026-07-18
refactor(nix): use nixpkgs claude-code instead of vendoring

connor-grady:
Switches the Nix package from vendoring the bundled Claude Code binary to using nixpkgs `claude-code`, pointed at via `MERIDIAN_CLAUDE_PATH`.

## Why

With the binary vendored, meridian had to patch a proprietary binary itself. `cp -r node_modules` shipped three ~235 MB copies of the Claude CLI, and `autoPatchelfHook` rewrote the interpreter of Anthropic's prebuilt binary so it would run on NixOS at all; every query died with `exit=127` until that patching landed (#501). It duplicated, less completely, work nixpkgs already does: nixpkgs' `claude-code` wraps the same binary with ripgrep on `PATH`, `procps`, and the bubblewrap sandbox, and it is maintained and reviewed as a first-class package. Relying on it removes both the vendored binaries (the closure was over 1 GB) and the fiddly, breakage-prone job of hand-patching an upstream binary.

## What changes

The wrapper sets `MERIDIAN_CLAUDE_PATH` to `${lib.getExe claude-code}` (the resolver's `env` source wins unconditionally), the vendored binaries are pruned in `installPhase`, and `--ignore-scripts` skips the postinstall that materialized the 248 MB `bin/claude.exe`. That drops `autoPatchelfHook`, `buildInputs`, `dontStrip`, and `autoPatchelfIgnoreMissingDeps`; the only runtime-loaded native binary, libsql's `.node`, is a shared object that resolves its libraries through node's own closure. Closure goes from ~1 GB to ~87 MB.

## The package is now correctly licensed

The most important improvement is licensing. With the binary vendored, the package declared `license = lib.licenses.mit` while its build output redistributed Anthropic's proprietary `claude.exe`, mislabeling the license and shipping an unfree binary that was never declared as such. The package is now correct: it carries no proprietary binary, its MIT license accurately describes meridian's own code, and it depends on the declared-unfree `pkgs.claude-code` like any other unfree dependency.

The unfree component now lives where it belongs, in the dependency. One caveat worth stating plainly: `overlays.default` and the home-manager module re-export the derivation the flake already evaluated with `allowUnfree`, so consuming Meridian through them pulls in `claude-code` without the consumer's own `allowUnfree` opt-in. That is a known property of flakes: a consumer's `nixpkgs.config` does not re-run against an already-evaluated flake output (NixOS/nixpkgs#118115). It applies to any flake exporting an unfree-dependent package and is not introduced by this PR.

## Version coupling

nixpkgs `claude-code` tracks its own cadence, not the SDK's `@anthropic-ai/claude-code` pin, so `meta.broken = versionOlder claude-code.version <floor>` (read from `package.json`) fails the build when the pinned nixpkgs is older than the SDK requires, with `allowBroken` as the escape hatch. `.github/workflows/check-claude-code.yml` runs on `package.json`/`bun.lock` PRs: if a raised floor outpaces the pinned nixpkgs it reports whether `nix flake update nixpkgs` fixes it (bump the lock in the same PR) or nixpkgs hasn't packaged it yet (hold the bump).

Verified: builds, closure ~87 MB, libsql loads unpatched, the SDK's JS imports without its pruned platform binary, and `/health` reports `claudeExecutable.source: "env"` resolving to nixpkgs `claude-code`.

### DIFF
--- a/.github/workflows/check-claude-code.yml
+++ b/.github/workflows/check-claude-code.yml
@@ -1,0 +1,29 @@
+name: Check claude-code availability
+
+on:
+  pull_request:
+    paths:
+      - package.json
+      - bun.lock
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: cachix/install-nix-action@v31
+
+      - name: nixpkgs must provide a claude-code satisfying the floor
+        run: |
+          satisfies() { nix eval .#packages.x86_64-linux.meridian.drvPath >/dev/null 2>&1; }
+          if satisfies; then
+            echo "current nixpkgs claude-code satisfies the @anthropic-ai/claude-code floor"
+            exit 0
+          fi
+          nix flake update nixpkgs
+          if satisfies; then
+            echo "::error::the @anthropic-ai/claude-code floor outpaces the pinned nixpkgs; a bump satisfies it, so run 'nix flake update nixpkgs' and include the flake.lock change in this PR"
+            exit 1
+          fi
+          echo "::error::no released nixpkgs claude-code satisfies the @anthropic-ai/claude-code floor yet; hold this bump until nixpkgs packages it"
+          exit 1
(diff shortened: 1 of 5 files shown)


rynfar:
Landed via #660 with your commits as-authored — thank you for this one, @connor-grady. The licensing correction and the 1 GB → 87 MB closure drop are both significant, and the `meta.broken` floor + CI check is exactly the right guard for the version-coupling problem.

One fix-up on top: the wrapper now uses `--set-default` instead of `--set` for `MERIDIAN_CLAUDE_PATH`, so a user's own override (e.g. via the home-manager module's `environment` option) still wins over the nixpkgs default.

PR #668, opened 2026-07-21
feat: install plugins in Docker via MERIDIAN_PLUGINS

### LINKED ISSUE #667: Install plugins in Docker
I'm currently playing with Meridian + pi and am currently getting the 400 Extra Usage error when I run it with the default configuration. However, it works as expected when I disable the passthrough of pi's system prompt in the settings. I'm assuming what I really need is to install the scrubber plugin for pi into Meridian. Because I'm running through Docker, there isn't an obvious, blessed way to install plugins on Docker.

First, is there something I'm missing here? Is the expectation that we install it somewhere on the host and bind mount it or is there another preferred workflow?

Second, if there isn't a preferred workflow, I would be willing to contribute this. My first thought was to add something to the entrypoint command to use an env var (`MERIDIAN_PLUGINS`) to install them on startup into the correct location before bootstrapping Meridian. If this would be an acceptable approach, I'd be happy to contribute it. Alternatively, I'm open to workshopping an alternative approach if you have a preference or other ideas.

Thanks for developing this!

mAAdhaTTah:
Create a script to install `MERIDIAN_PLUGINS` on startup.

Fixes #667.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,8 +49,8 @@ RUN node /app/node_modules/@anthropic-ai/claude-code/install.cjs
 RUN mkdir -p /app/bin/shims \
     && ln -sf /app/node_modules/@anthropic-ai/claude-code/bin/claude.exe /app/bin/shims/claude
 ENV PATH="/app/bin/shims:$PATH"
-COPY --chown=claude:claude bin/docker-entrypoint.sh bin/claude-proxy-supervisor.sh ./bin/
-RUN chmod +x ./bin/docker-entrypoint.sh ./bin/claude-proxy-supervisor.sh
+COPY --chown=claude:claude bin/docker-entrypoint.sh bin/claude-proxy-supervisor.sh bin/docker-install-plugins.mjs ./bin/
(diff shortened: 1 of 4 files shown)


rynfar:
Landed via #670 with your commit as-authored — thank you @mAAdhaTTah, this was a well-built contribution (the injection-safe exec, require.resolve instead of a hardcoded entry path, and the boundary-aware dedup were all exactly right).

One fix-up on top, caught by live-testing against the real registry: npm needs an anchor `package.json` in `~/.config/meridian`, or it walks up the directory tree and installs into the nearest ancestor project — the manifest step then throws and the container starts without plugins. The landed version writes the anchor first, warns loudly in the entrypoint on install failure, adds `--prefer-offline` for restarts, and swaps the docs example to packages that are published today (`hermes-scrub` isn't on npm yet — tracked in rynfar/meridian-plugin-hermes-scrub#5).

Ships in the next release. For your #667 use case in the meantime: `MERIDIAN_PLUGINS="@rynfar/meridian-plugin-pi-scrub"` is exactly the right call.

PR #683, opened 2026-07-24
feat: image input and incomplete status for Responses API (/v1/responses)

diwakar-s-maurya:
## Summary

Two small, additive improvements to the Codex Responses API (POST /v1/responses): image input and truncation reporting. Both were gaps in the #654 implementation — image parts were silently dropped and a token-limited response always reported completed.

Without the fix in this pull request, codex-cli fails to view images.
Screenshot_20260724_112021

## What's included

- Image input — `input_image` content parts carrying a base64 data URL now translate to Anthropic image blocks. Text-only turns keep the existing fast text path untouched; remote (non-data-URL) image URLs remain unsupported and are replaced with an omission note, matching the /v1/chat/completions adapter.
- Incomplete status — an Anthropic `max_tokens` stop now maps to a Responses incomplete status with `incomplete_details: { reason: "max_output_tokens" }`, in both the non-streaming response object and the streaming path (emits response.incomplete instead of response.completed). A truncated answer is now distinguishable from a finished one.

## Verification

- Unit (`openai-responses.test.ts`): data-URL → image block, omission note for non-data-URL images, and the incomplete status for both non-stream and stream.
- End-to-end (`proxy-responses-image.test.ts`): a real 1×1 PNG posted through /v1/responses (non-stream and stream) reaches the mocked SDK as an actual image content block with the exact base64 payload — not the [Image attached] placeholder the text-flatten path would produce. Confirms the image survives translation, hasMultimodalContent detection, and prompt construction.
- Tested with codex-cli 0.145.0 and claude-code-cli 2.1.218

### DIFF
--- a/src/__tests__/openai-responses.test.ts
+++ b/src/__tests__/openai-responses.test.ts
@@ -11,6 +11,7 @@ import {
   translateAnthropicToResponses,
   createResponsesSseTranslator,
 } from "../proxy/openaiResponses"
+import type { AnthropicContentBlock } from "../proxy/openai"
 
 describe("translateResponsesToAnthropic", () => {
   it("returns null without input", () => {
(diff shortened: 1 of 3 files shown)


rynfar:
thanks i will review this shortly

rynfar:
Thanks @diwakar-s-maurya — this is a well-built patch, and the screenshot made the problem concrete immediately.

I've carried your commit into #687 rather than merging here, because I wanted to add a follow-up on top and would rather that be a separate commit than a review round-trip on your branch. Your commit is cherry-picked unchanged and still authored by you (`92745e0c`), and #687 merges rather than squashes so it stays its own entry in the history.

The thing I want to single out is the e2e test. Asserting the exact base64 payload arrives at the SDK — and explicitly asserting the *absence* of the `[Image attached]` placeholder that the text-flatten path in `server.ts` would produce — is testing the actual regression rather than the happy path. That caught the real failure mode at the HTTP boundary, and it's the kind of test that will still be doing work a year from now.

Two notes on the review:

**One thing I checked and cleared.** `partsToBlocks` pushes `{type:"text", text: part.text}` for any string including `""`, where the pre-image path guarded with `if (text)`. Empty text blocks are normally a 400 from the REST API, and an uncaptioned image seemed like a plausible Codex shape, so I expected a bug. It isn't one — against the real SDK, `[{text:""}, image, question]` and `[image, question]` return identical answers. No change needed; noting it so it doesn't get re-raised later.

**What the follow-up commit does.** `parseDataUrlImage` was copied byte-for-byte from `openai.ts` into `openaiResponses.ts`. Entirely reasonable given it wasn't exported, but two identical copies drift, so I exported the original and imported it.

With one home for the parser, I closed a gap that's **pre-existing and not something your PR introduced** — the copy you inherited had it too. The regex accepts any `image/*`, so `image/svg+xml` or `image/bmp` was forwarded and rejected upstream with an opaque API error. It's now an allowlist of jpeg/png/gif/webp, with unsupported types omitted via the same note you used for remote URLs. I also normalized the `image/jpg` misspelling to `image/jpeg`, since it's common in real data URLs and the allowlist would otherwise have turned a broken-but-forwarded image into a silently dropped one.

Both behaviors are now asserted from both adapters, which is what should stop them diverging again.

Closing in favor of #687 — the credit stays with you. Thanks for the clean patch and for testing it against real codex-cli before sending.

PR #684, opened 2026-07-24
feat: add Claude Opus 5 to the model list, make it the canonical opus

robertn702:
## Summary

Two changes, mirroring the Sonnet 5 rollout (#644):

1. **`/v1/models` never listed `claude-opus-5`** — explicit requests already routed correctly via `explicitModelPin` (the `claude-(...)-\d` regex matches it), but tools that build their picker from the endpoint couldn't offer it. Added as the first Opus entry.
2. **The bare `opus` alias resolved to `claude-opus-4-8`.** Canonical pin bumped to `claude-opus-5` — an alias means "the current model of that tier". `MERIDIAN_DEFAULT_OPUS_MODEL=claude-opus-4-8` restores the old default; explicit ids are honored exactly.

## Verification
- Updated `explicit-model-pins`, `openai`, and `proxy-env-stripping` assertions; added a canonical-opus unit test and a bare-opus integration test.
- Full suite **2204 pass, 1 skip, 0 fail**; `tsc --noEmit` clean.
- Live through a real proxy on Max: `POST /v1/messages` with `claude-opus-5` answered from `claude-opus-5`.

### DIFF
--- a/src/__tests__/explicit-model-pins.test.ts
+++ b/src/__tests__/explicit-model-pins.test.ts
@@ -13,7 +13,7 @@ import { describe, it, expect, mock, beforeEach } from "bun:test"
 
 // ---------- unit: explicitModelPin + canonical bump ----------
 
-const { explicitModelPin, CANONICAL_SONNET_MODEL } = await import("../proxy/models")
+const { explicitModelPin, CANONICAL_SONNET_MODEL, CANONICAL_OPUS_MODEL } = await import("../proxy/models")
 
 describe("explicitModelPin (#631)", () => {
   it("pins fully-versioned sonnet ids", () => {
@@ -22,6 +22,7 @@ describe("explicitModelPin (#631)", () => {
   })
 
   it("pins fully-versioned opus and date-suffixed haiku ids", () => {
+    expect(explicitModelPin("claude-opus-5")).toEqual({ ANTHROPIC_DEFAULT_OPUS_MODEL: "claude-opus-5" })
     expect(explicitModelPin("claude-opus-4-7")).toEqual({ ANTHROPIC_DEFAULT_OPUS_MODEL: "claude-opus-4-7" })
     expect(explicitModelPin("claude-haiku-4-5-20251001")).toEqual({ ANTHROPIC_DEFAULT_HAIKU_MODEL: "claude-haiku-4-5-20251001" })
   })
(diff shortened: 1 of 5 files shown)


rynfar:
thank you i will get this in asap.

rynfar:
Thanks @robertn702 — both halves of this are right, and the reasoning in the description matched what I found reviewing it. The bare-alias argument in particular is the correct call: an alias should mean "the current model of that tier," and you kept the `MERIDIAN_DEFAULT_OPUS_MODEL` escape hatch intact for anyone who wants to pin back.

I've carried your commit into #685 rather than merging here, because I wanted to add a few things on top and would rather they be separate commits than review round-trips on your branch. Your commit is cherry-picked unchanged and still authored by you (`87cfcdad`), and #685 will be merged rather than squashed so it stays its own entry in the history.

What the follow-up commit adds:

- **Test coverage for the new entry.** The two `buildModelList` context-window tests are named "all opus variants" but asserted only 4-6/4-7/4-8, so `claude-opus-5` was the one entry in that list whose `isMaxSubscription ? 1_000_000 : 200_000` branch nothing covered. Both tests now include it. Easy to miss — the test names read as though they already covered everything.
- **`BUILTIN_MODEL_PRICING` registration.** Not a bug in your change: `resolveModelPricing` has an `includes("opus")` family fallback that was already returning the correct $5/$25 for `claude-opus-5`. The gap is that the same table is what `GET /pricing` serves as the settings page's list of user-overridable models, so opus-5 was routable but wouldn't appear there — unlike `claude-sonnet-5`, which got an explicit entry when Sonnet 5 landed. One line, plus an assertion in `pricing-routes.test.ts`.
- **Docs.** Added Opus 5 to the crush and factory example model lists in `docs/agents.md`.

One thing worth passing along, since it was the only real risk I saw in the canonical bump: `package.json` pins `@anthropic-ai/claude-code` at `^2.1.198`, and that floor was set by #560 specifically so Fable 5 would resolve, immediately before the Fable 5 routing PR #561. The 2.1.198 binary contains no `claude-opus-5` string, so I wanted to know whether someone sitting on the floor would regress on every bare `opus` request.

They don't — the SDK forwards `ANTHROPIC_DEFAULT_OPUS_MODEL` to the API verbatim instead of validating it against its own model registry. Checked against the pinned binary:

| requested | pin | `modelUsage` |
|---|---|---|
| `opus` | `claude-opus-5` | `claude-opus-5` |
| `opus[1m]` | `claude-opus-5` | `claude-opus-5[1m]` |
| `opus` | `claude-opus-4-8` | `claude-opus-4-8` (control) |

So the #560 → #561 precedent doesn't transfer to model-id changes, and no floor bump was needed. Useful to know for the next one of these.

Closing in favor of #685 — the credit stays with you. Appreciate the clean patch and the live verification in the description.

PR #688, opened 2026-07-26
fix: make postinstall script Windows-portable

fhalim:
The postinstall script used POSIX shell syntax (2>/dev/null, || true) which cmd.exe cannot parse on Windows, causing npm install to fail entirely instead of tolerating a missing/failed claude-code install step. Replace it with a single node -e invocation that spawns the same install.cjs via spawnSync (never throws on non-zero exit) and uses require.resolve so it also finds claude-code when npm hoists it above meridian's own node_modules.


rynfar:
Thanks for this — the diagnosis is exactly right, and it's a real breakage. I verified the fix on my end: `require.resolve` correctly walks up to hoisted `node_modules` (the old hardcoded path couldn't), and `spawnSync` never throws on non-zero exit or ENOENT, so install failures stay non-fatal.

I've landed it in #693 with your commit cherry-picked, so the fix is authored by you. I added one follow-up commit on top: `stdio:'ignore'` also silenced stdout, where the old shell form only redirected stderr. Since that step downloads a ~200 MB platform binary, users would lose all progress feedback and a failed download would be invisible — which matters because `models.ts` has a fallback resolver chain built around "the upstream postinstall failed" (#445). Switched to `stdio:'inherit'` and made the `catch` log instead of swallow.

Closing this in favor of #693. Not a reflection on the patch — I just wanted it in before the next release rather than leaving it pending on a round-trip. Appreciate you catching this; the Windows install path doesn't get enough eyes.

Worth noting: our `windows-smoke` CI job is supposed to catch this exact bug class, but it runs `bun install`, and bun's bundled shell parses `2>/dev/null || true` fine on Windows — so it never could have caught it. Fixing that separately.

PR #690, opened 2026-07-26
fix(session): bound modified-continuation resume so stale lineage replays fresh

### LINKED ISSUE #689: Modified-continuation resume drops client tool results when the stored lineage is behind (client-driven tool loops)
## Summary

In `mode: internal` with a client-driven tool loop (opencode adapter), the "Modified continuation" branch in `src/proxy/session/lineage.ts` resumes a stored SDK session that is **missing every message from intervening tool rounds**, and the resume slice silently drops those tool results from the model's context. The agent then truthfully reports "the tool returned nothing" / "results never came back to me" on follow-up turns, while the client's transcript shows the results arrived fine.

## Environment

- Meridian 1.57.0 (also observed on 1.56.1; the branch is unchanged back through 1.5x)
- `mode: internal`, adapter `opencode` (OpenClaw gateway traffic)
- Claude model: `claude-opus-5` (not model-specific)
- Client drives its own tool loop: mid-loop requests end in `user[tool_result]`

## Mechanism

Two interacting behaviors:

1. **Tool-loop requests are forced onto fresh sessions** (`isClientDrivenLoop` → `lineage=new session=new`) and those sessions are **never persisted back** to the lineage store when an entry already exists for the conversation key. Only clean conversational turns advance the stored session.
2. `verifyLineage()`'s modified-continuation branch allows resume on **any** `prefixOverlap > 0` when the incoming conversation is longer, with no bound on how far behind the stored session is:

```ts
// src/proxy/session/lineage.ts @ meridian-v1.57.0, ~L296
if (prefixOverlap > 0 && messages.length > cached.messageCount) {
  // "Modified continuation … Allowing resume"
  return { type: "continuation", session: cached }
}
```

So after any tool round, the stored lineage is N messages behind (we observed gaps of 5 to 14). The next conversational turn resumes that stale session, and the SDK-side slice (computed from the stored `messageCount`) feeds the SDK a suffix that starts past the tool round, so the tool_use/tool_result pairs never reach the model's context.

## Live log capture (fresh store, reproduced in ~90 seconds)

```
18:09:29 lineage=new                                   # user turn 1 (tool round 1)
18:10:01 lineage=new msgCount=6                        # round 1 final, becomes stored lineage
18:10:51 Modified continuation: overlap 5/6,  incoming 8   gap 2  ✓ resume ok ("thanks")
18:10:57 lineage=new msgCount=10                       # tool round 2, throwaway, never persisted
18:11:19 Modified continuation: overlap 7/8,  incoming 13  gap 5  ✗ round 2 missing ("nice")
18:11:43 Modified continuation: overlap 12/13, incoming 18 gap 5  ✗ recall turn
```

At the recall turn the model answered (verbatim): *"The GREEN-33 and GOLD-44 calls were dispatched twice, but both times the results were forwarded to the client and never came back to me."*, while the client transcript contains both tool_result blocks with the correct output.

## Reproduction

1. `mode: internal`, opencode-style client that runs its own tool loop.
2. Turn 1: prompt that triggers ≥2 parallel tool calls (e.g. two `exec echo` calls); client returns tool_results; model summarizes. ✓
3. Turn 2: plain "thanks" (conversational; this creates/advances the stored lineage).
4. Turn 3: another tool round (results returned fine). ✓
5. Turn 4: plain filler ("nice").
6. Turn 5: "Without running any tools, list the exact strings those commands printed."

Expected: model recalls all outputs (they're in the conversation history the client sends).
Actual: model has no knowledge of turn-3's tool results and typically asserts the tools failed. Clearing the session store only helps until the next tool round; the gap re-forms immediately (this is not stale-store accumulation).

## Suggested fix

Bound the modified-continuation resume: only resume when the stored lineage is actually current, e.g. `prefixOverlap >= stored - 1 && incoming - stored <= 2` (last-slot hash churn plus exactly one new exchange). Anything else should be treated as `diverged` → fresh full-history replay, which is always correct; resume is only a latency/cache optimization. We are running exactly this guard as a local source patch on our 1.57.0 build and it eliminates the failure (the fresh replay re-adopts the full history, so the store self-heals after every tool round).

Alternative: persist the tool-loop sessions back to the lineage store so the stored session is never behind, but the gap-bound is the smaller, safer change.

Happy to provide the full diagnostic logs or test against a branch. Will get to a PR shortly

kilabyte:
Fixes #689

## Problem

In `mode: internal` with a client-driven tool loop (opencode adapter), the modified-continuation branch in `src/proxy/session/lineage.ts` allowed resume on any `prefixOverlap > 0` whenever the incoming conversation was longer, with no bound on how far behind the stored session was.

Client-driven tool loop requests are forced onto throwaway sessions that never persist back to the lineage store, so after any tool round the stored lineage is several messages behind. The next conversational turn resumed that stale session, and because the branch advances `messageCount` to the incoming length before the resume slice is computed, the proxy sent only the last user message. The intervening tool_use/tool_result pairs never reached the model's context, and the agent reported that the tools returned nothing while the client transcript showed the results arrived fine.

## Fix

Bound the modified-continuation resume exactly as suggested in the issue. Resume is allowed only when the stored lineage is actually current:

- `prefixOverlap >= stored - 1`: at most the last stored slot may differ (benign churn)
- `incoming - stored <= MAX_CONTINUATION_GAP` (2): at most one new exchange appended (the previous assistant reply plus the new user message)

Within that bound, sending only the last user message is provably safe: everything else in the incoming history is already inside the stored SDK session. Anything further behind is treated as `diverged`, and the stale cache entry is evicted. The caller then replays the full history on a fresh session, which is always correct (resume is only a latency/cache optimization), and the store re-adopts the full history at end of turn, so it self-heals after every tool round.

A `Stale modified continuation` diagnostic line is logged on the new path, matching the existing lineage log format.

## Tests

Four new unit tests in `src/__tests__/lineage-unit.test.ts` cover the bound through the public `verifyLineage` API:

- last-slot churn plus one appended exchange still resumes, and the store adopts the incoming history (mirrors the healthy `overlap 5/6, incoming 8` capture from the issue)
- a stored lineage behind by a tool round diverges and evicts the cache entry (mirrors the failing `overlap 7/8, incoming 13, gap 5` capture)
- a gap of `MAX_CONTINUATION_GAP + 1` diverges
- churn deeper than the last stored slot diverges even with a small gap

The pre-existing modified-continuation test (grow 7 to 9 with the last slot modified) sits exactly inside the bound and still passes unchanged.

Full suite: 2218 pass, 0 fail. Build clean.

### DIFF
--- a/src/__tests__/lineage-unit.test.ts
+++ b/src/__tests__/lineage-unit.test.ts
@@ -12,6 +12,7 @@ import {
   verifyLineage,
   normalizeContextUsage,
   MIN_SUFFIX_FOR_COMPACTION,
+  MAX_CONTINUATION_GAP,
   type SessionState,
 } from "../proxy/session/lineage"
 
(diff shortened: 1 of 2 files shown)


rynfar:
Thanks for this — the diagnosis is exact, and the write-up made it straightforward to verify. I ran your change against `main` across 264 synthetic scenarios (stored length × churn position × gap): 60 classifications change, and **every one moves `continuation → diverged`, none the other way**. `compaction` and `undo` are untouched. That's the property that made this safe to land quickly — it can't loosen a resume, only tighten one.

Landed in #700. One logistical note so this doesn't read as a rejection: `main` requires signed commits, so the fork commit left the PR permanently `BLOCKED` despite all three checks passing — nothing you did wrong, and not something you could have fixed from your side. Cherry-picking re-signs it while preserving you as commit author, so the change on `main` is byte-identical to yours and credited to you.

Two notes from validating it:

- I couldn't reproduce "2218 pass, 0 fail" — I get 2 failures on your branch *and* on its base commit (both profile-routing tests, unrelated, green in CI). Pre-existing and not your problem, but the "0 fail" claim didn't hold in my environment.
- The bound trades the wrong answers for full replays on affected clients, so those sessions rebuild their prompt cache. Tracked in #701 along with the root fix you suggested in the issue — persisting the tool-loop sessions — which recovers the performance.

Appreciate the quality of the report; the log captures with the gap arithmetic made this reviewable in one pass.

PR #692, opened 2026-07-27
fix: replay stale session histories safely

Ca134:
## Summary

- make lineage verification side-effect free and return an explicit resume boundary
- fresh-replay the complete client history when a growing request changes the cached prefix
- resume compaction after the preserved suffix so every newly accumulated turn is forwarded
- document the safety rule and add streaming/non-streaming regression coverage

## Root cause

In a distributed Meridian deployment, a request can return to a proxy node with a stale local/shared session snapshot. In the observed shape, the node had 515 cached messages while the incoming request had 727, with only the first 514 cached messages matching.

The previous `modified continuation` path updated the cached count/hash to the incoming 727-message state before the server selected the request delta. The server then resumed the older SDK session and sent only the final user message, silently omitting the intervening client history.

This change treats a growing history with a changed cached prefix as unverifiable. It starts a fresh SDK session, replays the complete client history, and only stores the new lineage after the upstream request succeeds. A following request resumes the repaired session normally.

Strict prefix continuations still resume, compaction resumes from the matched suffix boundary, undo still forks at its rollback UUID, and busy/stale-UUID retry paths remain unchanged.

## Tests

- `bun test src/__tests__/lineage-unit.test.ts src/__tests__/proxy-session-resume.test.ts src/__tests__/session-lineage.test.ts` (`82 pass`)
- `bun run test` (`2218 pass`, `1 skip`, `0 fail`)
- `bun run typecheck`
- `bun run build`

The HTTP regression tests cover both streaming and non-streaming 515-to-727 stale-session requests, verify that middle-history sentinels reach the fresh prompt, and verify that the next turn resumes the repaired session instead of replaying again.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -172,10 +172,16 @@ Every request verifies that incoming messages are a valid continuation of the ca
 
 | Classification | Condition | Action |
 |---------------|-----------|--------|
-| **Continuation** | Prefix hash matches stored | Resume normally |
-| **Compaction** | Suffix preserved, beginning changed | Resume (agent summarized old messages) |
+| **Continuation** | Full stored prefix hash matches | Resume from the stored message boundary |
+| **Compaction** | Suffix preserved, beginning changed | Resume after the matched suffix |
 | **Undo** | Prefix preserved, suffix changed | Fork at rollback point |
-| **Diverged** | No meaningful overlap | Start fresh session |
+| **Diverged** | No overlap, unverifiable state, or a changed cached prefix followed by new messages | Start fresh and replay the complete client history |
+
+Lineage verification is side-effect free. A classification may select an SDK
+session and a replay boundary, but updated message counts and hashes are only
+stored after the upstream request succeeds. When Meridian cannot prove that an
+SDK session contains a section of client history, it starts fresh rather than
+silently skipping that section.
 
 ## Testing Strategy
 
--- a/src/__tests__/lineage-unit.test.ts
+++ b/src/__tests__/lineage-unit.test.ts
@@ -11,7 +11,6 @@ import {
   measureSuffixOverlap,
   verifyLineage,
   normalizeContextUsage,
-  MIN_SUFFIX_FOR_COMPACTION,
   type SessionState,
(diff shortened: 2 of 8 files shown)


rynfar:
Landed in #705 with your commit cherry-picked, so the fix is authored by you.

Your diagnosis was right and the approach was the right one. Making verification side-effect free and returning an explicit `resumeFrom` doesn't just fix the stale-node case — it removes the whole class. I verified that while reviewing: there was a **third, unreported instance** of the same defect still live on `main` after #700, on the compaction path. An early churned message plus an appended tool round classifies as compaction, and the same `messageCount` mutation collapsed the slice:

```
before:  messages reaching the model: 1   dropped: tool_use RED-77, tool_result RED-77, summary, BLUE-99
after:   messages reaching the model: 5   dropped: none   (resumeFrom=8)
```

Your `resumeFrom: suffixStartInIncoming + suffixOverlap` fixes it. Neither #689 nor this PR mentioned that path, so it would still be dropping tool results today.

Two notes on the rebase:

- Removed `MAX_CONTINUATION_GAP`, added by #700 to bound the modified-continuation resume. Dead under your change — a changed prefix slot means the SDK session holds content the client no longer claims, so the shape diverges regardless of gap size.
- Reworked #700's regression block rather than dropping it. Three of its four scenarios already asserted `diverged`; the fourth asserted last-slot churn *resumes*, which your change deliberately reverses, so it now asserts divergence.

I also added a safety harness on top (`src/__tests__/lineage-safety.test.ts`): property checks over 131 generated scenarios, the key one being that everything before `resumeFrom` must genuinely already be in the SDK session. All three bugs violate it. I verified the harness actually fails by reintroducing the mutation — a test never seen failing isn't evidence of much.

Closing this in favour of #705. To be clear about why it wasn't merged directly: `main` requires signed commits, so fork PRs sit at `BLOCKED` with every check green and no visible reason. Not something you could have fixed from your side — #688 and #690 hit the same wall. I've documented it in #706 so it stops surprising people.

Thanks for this one. The 515/727 capture and the root-cause write-up made it reviewable in a single pass, and it turned out to fix more than it claimed.

PR #711, opened 2026-07-28
feat(health): report how long the login has left

CrazyCoder:
## Problem

The background refresh scheduler keeps the access token alive indefinitely, so the proxy looks healthy right up until it isn't. The **refresh token** behind it has its own expiry, and when that lapses nothing automated recovers — every request 401s until someone runs `claude login` interactively.

That deadline was invisible from outside the process. `/health` reported only `loggedIn: true`, and `refreshTokenExpiresAt` — which the CLI already writes into `.credentials.json` — was referenced nowhere in the codebase. The first sign of trouble is the proxy already being down, which for an unattended deployment can mean hours.

## Change

Three fields on the `/health` auth block:

```jsonc
"auth": {
  "loggedIn": true,
  "email": "...",
  "subscriptionType": "max",
  "refreshTokenExpiresAt": 1786905597848,
  "daysUntilRenewal": 19,
  "renewalRequiredSoon": false
}
```

`renewalRequiredSoon` flips inside a warning window defaulting to **3 days** (`MERIDIAN_AUTH_RENEWAL_WARN_DAYS` overrides), so a monitor can alert on a single boolean rather than doing date arithmetic in a shell script. Mine is a monit stanza:

```
if failed port 3456 protocol http request "/health"
    status = 200
    content != "renewalRequiredSoon.:true"
then alert
```

`daysUntilRenewal` deliberately uses the same `Math.ceil(remaining / 86400000)` the CLI uses for its own notice, so `/health` and the terminal never disagree by a day.

### Also: `doRefresh` was discarding the value

It spread the previous credentials and overwrote only `accessToken`/`refreshToken`/`expiresAt`, and the token-response type didn't parse a refresh-token expiry at all. The stored value could therefore only ever be as old as the last interactive login.

The CLI derives it from `refresh_token_expires_in` on every refresh and merges with `??` — keeping the previous value when the response omits it. This now does the same.

## Notes on the edge cases

- **A returned expiry that isn't in the future is rejected.** A refresh that just succeeded proves the token is still valid, so a seconds-rather-than-milliseconds value would otherwise be persisted as a 1970 timestamp and read downstream as "already expired", pinning a deployment to a permanent alert.
- **Absent `refreshTokenExpiresAt` never raises the flag.** Absence isn't evidence of imminent expiry; treating it as such would fire on every deployment whose credentials predate the field.
- **The lookup reads the resolved profile's credential store**, not the default one, since profiles are separate auth contexts keyed by `CLAUDE_CONFIG_DIR`. Reading the default store would report an unrelated account's expiry.
- **`0` is a valid window** (warn only once lapsed), so the env var is parsed with an explicit finite check rather than `Number(...) || DEFAULT`, which would swallow it as falsy.

## Testing

19 new cases in `src/__tests__/token-refresh.test.ts` covering persistence (returned / derived from `_in` / omitted / non-future / seconds-precision), the status helper (inside and outside the window, expired, absent field, no credentials, store throwing), and CLI parity on the rounding. `bun run test` and `tsc --noEmit` are clean; the failures on my machine are pre-existing Windows-only ones (POSIX file modes, path separators, sqlite) present on `main` without this change.

Verified against a live deployment: the countdown tracks a real 30-day login window and matches what the CLI prints.

### DIFF
--- a/src/__tests__/token-refresh.test.ts
+++ b/src/__tests__/token-refresh.test.ts
@@ -821,3 +821,172 @@ describe("startBackgroundRefresh timer clamping (#515)", () => {
     expect(d).toBeLessThan(MAX_32BIT)
   })
 })
+
+// ---------------------------------------------------------------------------
+// Login (refresh-token) expiry — the "your login expires in N days" signal
+// ---------------------------------------------------------------------------
+
+describe("refreshTokenExpiresAt persistence", () => {
+  let originalFetch: typeof globalThis.fetch
+  beforeEach(() => { originalFetch = globalThis.fetch })
+  afterEach(() => { globalThis.fetch = originalFetch })
+
+  it("persists refresh_token_expires_at when Anthropic returns it", async () => {
+    const { refreshOAuthToken } = await import("../proxy/tokenRefresh")
+    const rolled = Date.now() + 30 * 86_400_000
+    const { store, getStored } = makeStore()
+    mockFetch(async () => makeSuccessResponse({ ...MOCK_TOKEN_RESPONSE, refresh_token_expires_at: rolled }))
+
+    expect(await refreshOAuthToken(store)).toBe(true)
+    expect(getStored().claudeAiOauth.refreshTokenExpiresAt).toBe(rolled)
+  })
+
+  it("derives it from refresh_token_expires_in", async () => {
+    const { refreshOAuthToken } = await import("../proxy/tokenRefresh")
+    const { store, getStored } = makeStore()
+    const before = Date.now()
+    mockFetch(async () => makeSuccessResponse({ ...MOCK_TOKEN_RESPONSE, refresh_token_expires_in: 60 }))
+
+    expect(await refreshOAuthToken(store)).toBe(true)
+    const got = getStored().claudeAiOauth.refreshTokenExpiresAt
+    expect(got).toBeGreaterThanOrEqual(before + 60_000)
+    expect(got).toBeLessThanOrEqual(Date.now() + 60_000)
+  })
+
+  it("keeps the existing value when the response omits it", async () => {
+    const { refreshOAuthToken } = await import("../proxy/tokenRefresh")
+    const pinned = Date.now() + 5 * 86_400_000
+    const seeded = JSON.parse(JSON.stringify(MOCK_CREDENTIALS))
+    seeded.claudeAiOauth.refreshTokenExpiresAt = pinned
+    const { store, getStored } = makeStore(seeded)
+    mockFetch(async () => makeSuccessResponse(MOCK_TOKEN_RESPONSE))
+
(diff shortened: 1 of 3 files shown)


rynfar:
excellent i love this - ill review and get this commited thank you again for all your contributions @CrazyCoder

rynfar:
Thanks @CrazyCoder — merged as 22d0c818, cherry-picked so the commit keeps your authorship.

This was a genuinely good catch. The refresh-token deadline being invisible is exactly the kind of failure that only shows up as "the proxy has been down for hours," and the write-up made the reasoning easy to follow.

Two of your edge-case decisions earned their keep on the very first live run here, which is worth reporting back:

The active profile on this machine is a `work` profile whose keychain entry predates the CLI writing `refreshTokenExpiresAt`, while the default store has it:

| store | has the field |
|---|---|
| default | yes |
| `work` profile | no |

So `/health` returned `renewalRequiredSoon: false` with the other two fields absent. Had absence been treated as evidence of expiry, this deployment would have alerted permanently from the moment it shipped. And had the lookup not been profile-scoped, it would have reported the default account's expiry under the work account's identity — a number that looks plausible and is simply wrong. Both calls were right.

Verified working where the field does exist:

```
{"refreshTokenExpiresAt":1786080046027,"daysUntilRenewal":9,"renewalRequiredSoon":false}
expires: 2026-08-07 05:20
```

I added two commits on top:

**Caching (b33f2beb).** `getAuthRenewalStatus` read the credential store on every call, and on macOS that read spawns `/usr/bin/security find-generic-password`. `/health` turns out to be hot — the dashboard polls it every 10s per open page (`profileBar.ts`, `landing.ts`), plugin health checks hit it, and monitors like your monit stanza poll it too — so that was a subprocess per poll for a value that moves on a 30-day cadence. Now cached for 5 minutes keyed on the store's `refreshKey`, with in-flight sharing, mirroring what `getClaudeAuthStatusAsync` and `fetchOAuthUsage` already do. A store without a `refreshKey` isn't cached (unknown identity), and failed reads aren't cached so a keychain blip retries rather than suppressing the warning for the whole TTL. Only the read is cached — the day count and flag stay live.

**Extracted the warn-window parse (632e35c5).** Your reasoning about `0` being legitimate and `|| DEFAULT` swallowing it was right, but it lived inline in the handler where it couldn't be tested without process-global module mocking. It's now `resolveRenewalWarnDays()` with direct coverage, including that `0` case.

One small correction: the inline comment said the previous code "dropped it unconditionally" — the spread actually preserved any on-disk value, it just never parsed a new one from the response. Your PR description had this right; only the comment was off. Fixed.

Appreciate the care in this one, particularly the seconds-vs-milliseconds guard and matching the CLI's `Math.ceil`. Both are the sort of thing that's invisible when right and maddening when wrong.

PR #713, opened 2026-07-29
fix(session): reject compaction that resumes past the last message

### LINKED ISSUE #712: Session resume drops the user's message for stateless clients that append a trailing injected message
## Summary

For a **stateless client that re-sends the full conversation every turn and appends one or more injected messages after the user's own message** (SillyTavern and most roleplay frontends do exactly this — assistant prefill / post-history instructions), session-resume `continuation` silently drops the user's actual input. The model receives only the trailing injected message and continues its own previous output, which looks like "the model ignores me / is always one turn behind".

Observed on **1.57.1**, adapter `opencode` (reached via the default fallback — the client sends no identifying UA or header).

## Symptom

16 consecutive turns where the user's message never reached the model. The upstream SDK transcript (`~/.claude/projects/<slug>/<sdk-session>.jsonl`) shows the *same 6-character user turn* every time:

```
[2]  user  len=20605   <- first request, lineage=new, full replay: correct
[10] user  len=6       <- lineage=continuation
[16] user  len=6
[21] user  len=6
...                    <- 16x identical
[83] user  len=6
```

Those 6 characters are byte-identical to the client's configured **assistant prefill** string. The user's actual text appears nowhere.

Meanwhile the proxy log confirms the client *did* send everything — `msgCount` grows monotonically by 2 each turn (`…→45→47→49→51`), and the `msgs=` role summary shows the full history. So the loss is entirely inside the proxy.

## Mechanism

1. `detectAdapter()` — client sends no UA/`x-meridian-agent`/session header, so it falls through to `defaultAdapter` (`opencode`), i.e. it is treated as a **stateful** coding agent.
2. `lookupSession()` — no session header, so it falls back to `getConversationFingerprint()`, which hashes the **first user message + cwd**. For a chat frontend this is stable across the whole conversation, so every turn matches the cached session.
3. `verifyLineage()` returns `continuation` with `resumeFrom = cached.messageCount`.
4. `messagesToConvert = allMessages.slice(resumeFrom)` — only the delta is forwarded to the resumed SDK session.
5. The delta is then narrowed twice more:
   - on resume, assistant turns are blanked (`isResume → { role: "assistant", text: "" }`) and dropped by `turns.filter(t => t.text)`;
   - user text goes through `sanitizeTextContent()`, which strips paired `ORCHESTRATION_TAGS`.

Net effect: whatever survives is just the client's trailing injected message. `frameReplayTurns()` then sees `nonEmpty.length < 2` and returns that single string as the entire prompt.

The failure is silent — no warning, no `diverged` fallback, HTTP 200, and the model produces fluent output, so it reads as a model quality problem rather than a transport bug.

## Secondary issue: `ORCHESTRATION_TAGS` includes `thinking`, and is applied to user-authored content

`sanitizeTextContent()` is applied to **user** message text via `flattenUserContent()`, and `ORCHESTRATION_TAGS` contains `thinking`, so any paired `<thinking>…</thinking>` in a user message is deleted wholesale.

`<thinking>` is not a proxy-internal orchestration tag — it is an extremely common CoT convention in user-authored prompts and preset libraries. A preset built around `<thinking>` has its instructions partially eaten before the model ever sees them, on full replay too. The other entries in that list (`env`, `system_information`, `tool_exec`, `skill_content`, …) are genuinely internal; `thinking` is not in the same category.

## Suggested fixes

Any one of these would have prevented the data loss:

1. **Sanity-check the delta.** If, after assistant-blanking and sanitizing, the prompt built from `messagesToConvert` contains no non-empty user turn originating from a message that wasn't already in the stored prefix, fall back to full replay instead of sending it. Cheap, and turns a silent wrong answer into a slightly more expensive correct one.
2. **Don't resume when the client looks stateless.** If there is no session header and the match came only from a content fingerprint, full history was sent anyway — resuming buys little and risks exactly this. Consider making fingerprint-only resume opt-in.
3. **Require the stored prefix to end where the new turn begins.** The current check accepts any exact prefix match; it does not verify that the messages after `resumeFrom` actually constitute one new user turn.
4. **Scope tag stripping.** Restrict `ORCHESTRATION_TAGS` stripping to assistant/tool-generated content, drop `thinking` from the list, or make the list configurable.

## Reproduction sketch

1. Point a stateless client at meridian with no identifying UA (e.g. SillyTavern's Claude reverse-proxy field → `http://127.0.0.1:3456`).
2. Configure any non-empty assistant prefill, or any preset that injects a message *after* the user's message.
3. Send 3+ turns.
4. Watch `[PROXY] … lineage=` go `new` → `continuation`, then read the SDK transcript: from the second turn on, the recorded user turns contain only the injected trailing text.

Happy to test a patch against this setup.

1A7432:
Fixes #712.

## The bug

`verifyLineage` can return `compaction` with `resumeFrom === messages.length`. The slice is then empty, so `server.ts` takes its `resumeFrom < allMessages.length` fallback and calls `getLastUserMessage()`. For a client whose last message is not the user's turn, that sends the wrong message — and the user's actual input never reaches the model. 200, fluent output, nothing in the logs.

Stateless chat frontends hit this routinely. They re-send the whole history every turn and append a constant trailing block after the user's own message — an injected assistant line, plus a prefill sent as a user message. Repeat a turn verbatim and that block matches the stored tail for several slots, so `measureSuffixOverlap` anchors on the **final** message and `suffixStart + suffixOverlap` lands exactly on `messages.length`.

Captured from a real SillyTavern session (47 messages, stored count 39):

```
suffixOverlap=4  resumeFrom=41  messages.length=41   → slice empty
  → getLastUserMessage() → "思考已结束。"   (the client's prefill, 6 chars)
```

The upstream SDK transcript for that conversation shows the same 6-character user turn 16 times in a row. The model was continuing its own previous output every turn, because that is all it had.

## The fix

One condition in the compaction branch: the preserved suffix must end *before* the last message, so there is at least one new message to send.

The fast path already refuses this shape — `messages.length <= cached.messageCount` → `replayed-request`. The compaction branch simply lacked the equivalent guard. Nothing new after the suffix means it is a replay, not a compaction, so it now falls through to the existing divergence handling and replays in full.

Genuine compaction is unaffected: a client that compacts also appends the new turn, which keeps the run away from the end. The pruning-survival scenario in `session-lineage.test.ts` (middle tool outputs pruned, two turns appended) computes `resumeFrom=9` against `length=11` and still resumes.

## Verification

Replaying the captured 28-turn conversation through the lineage + prompt-building path:

| | user's turn preserved | lost |
|---|---|---|
| before | 25 | 3 |
| after | 28 | 0 |

End-to-end against a local build, same two requests, instruction placed where only a full replay delivers it:

```
stock 1.57.1  → lineage=compaction → "…if 思考已结束。 is meant as a new instruction, could you clarify"
patched       → BANANA
```

## Tests

- `lineage-unit.test.ts` — new block reproducing the shape: asserts no resume with an empty delta, asserts the full replay, and asserts a genuine compaction that appends a new turn still resumes. The first two fail without the fix.
- `lineage-safety.test.ts` — the resume-soundness invariant tightened from `resumeFrom <= incoming.length` to `< incoming.length`. Resuming at the end is never useful: it sends nothing and silently defers to `getLastUserMessage()`.

## Golden matrix

12 verdicts change, all `C -> D` — the conservative direction the header calls for.

- `stored=6|8 churn=@k gap=0` (10 rows) — same-length re-send with one slot edited. Every one had `resumeFrom == length`, so the old verdict resumed with an empty delta and re-sent the last user message while ignoring the edit. Replay is both safer and cheaper to reason about.
- `stored=6|8 compaction` (2 rows) — the synthetic scenario is `[summary, ...stored.slice(-3)]` with nothing appended, so there is genuinely nothing to send. Real clients append the new turn; the new unit test covers that shape and it still resolves to `compaction`.

`npm test` (2164 tests) and `npm run typecheck` pass.

### DIFF
--- a/src/__tests__/lineage-safety.test.ts
+++ b/src/__tests__/lineage-safety.test.ts
@@ -167,7 +167,10 @@ describe("lineage safety: invariants", () => {
       const incomingHashes = computeMessageHashes(s.incoming)
       const { resumeFrom } = result
 
-      expect(resumeFrom, `${s.label} resumeFrom past end`).toBeLessThanOrEqual(s.incoming.length)
+      // Strictly inside the array: resumeFrom is a slice start, so resuming at
+      // the end sends nothing and the caller silently falls back to the last
+      // user message, which is not necessarily the turn the client just added.
+      expect(resumeFrom, `${s.label} resumeFrom leaves nothing to send`).toBeLessThan(s.incoming.length)
 
       if (result.type === "continuation") {
         // Continuation resumes from the stored count, so the stored history
@@ -255,10 +258,10 @@ stored=4 churn=@3 gap=5 => D
 stored=4 shrink-to=1 => U
 stored=4 shrink-to=2 => U
 stored=6 churn=none gap=0 => D
-stored=6 churn=@0 gap=0 => C
-stored=6 churn=@1 gap=0 => C
-stored=6 churn=@2 gap=0 => C
-stored=6 churn=@3 gap=0 => C
+stored=6 churn=@0 gap=0 => D
+stored=6 churn=@1 gap=0 => D
+stored=6 churn=@2 gap=0 => D
+stored=6 churn=@3 gap=0 => D
 stored=6 churn=@4 gap=0 => D
 stored=6 churn=@5 gap=0 => U
 stored=6 churn=none gap=1 => R
@@ -291,14 +294,14 @@ stored=6 churn=@4 gap=5 => D
 stored=6 churn=@5 gap=5 => D
 stored=6 shrink-to=1 => U
 stored=6 shrink-to=3 => U
-stored=6 compaction => C
+stored=6 compaction => D
 stored=8 churn=none gap=0 => D
-stored=8 churn=@0 gap=0 => C
-stored=8 churn=@1 gap=0 => C
-stored=8 churn=@2 gap=0 => C
-stored=8 churn=@3 gap=0 => C
-stored=8 churn=@4 gap=0 => C
-stored=8 churn=@5 gap=0 => C
+stored=8 churn=@0 gap=0 => D
+stored=8 churn=@1 gap=0 => D
+stored=8 churn=@2 gap=0 => D
+stored=8 churn=@3 gap=0 => D
+stored=8 churn=@4 gap=0 => D
+stored=8 churn=@5 gap=0 => D
 stored=8 churn=@6 gap=0 => D
 stored=8 churn=@7 gap=0 => U
 stored=8 churn=none gap=1 => R
@@ -339,7 +342,7 @@ stored=8 churn=@6 gap=5 => D
 stored=8 churn=@7 gap=5 => D
 stored=8 shrink-to=1 => U
 stored=8 shrink-to=4 => U
-stored=8 compaction => C
+stored=8 compaction => D
 unrelated history => D`
 
     expect(actual).toBe(expected)
--- a/src/__tests__/lineage-unit.test.ts
+++ b/src/__tests__/lineage-unit.test.ts
@@ -573,3 +573,73 @@ describe("normalizeContextUsage", () => {
     expect(result.output_tokens).toBe(50)
   })
 })
+
+describe("verifyLineage compaction that reaches the end of the incoming array", () => {
+  // Stateless chat frontends (SillyTavern and most roleplay clients) re-send the
+  // whole history every turn and append a constant trailing block after the
+  // user's own message — an injected assistant line plus a prefill sent as a
+  // user message. When the user repeats a turn verbatim, that trailing block
+  // matches the stored tail for several slots, so the suffix anchor lands on the
+  // final message and resumeFrom comes out equal to messages.length.
+  //
+  // The slice is then empty and the caller falls back to getLastUserMessage(),
+  // which returns the constant prefill rather than the turn the user just typed,
+  // so the request reaches the model with the user's input missing entirely.
+  const FAKE_ASSISTANT = "[injected] let's look at what the user wrote"
+  const LATEST = "<latest_human_message>\ngo on\n</latest_human_message>"
+  const PREFILL = "done thinking."
+
+  const head = [
+    msg("user", "turn one"),
+    msg("assistant", "reply one"),
+    msg("user", "turn two"),
+    msg("assistant", "reply two"),
+  ]
+  const trailingBlock = [
+    msg("user", "go on"),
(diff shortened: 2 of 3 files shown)


rynfar:
thanks for this, i will review this and get it repaired today

rynfar:
Thanks @1A7432 — merged as 8c308995, cherry-picked so the commit keeps your authorship. #712 is closed too.

This was a genuinely hard bug to find and an unusually clean write-up. The captured numbers made it reproducible immediately:

```
lineage        : compaction  suffixOverlap=2
resumeFrom     : 10   messages.length = 10
slice empty?   : true
MODEL RECEIVES : "思考已结束。"
USER ACTUALLY SAID: "MY ACTUAL QUESTION"
```

Your reasoning about the fix is what sold it: the fast path already refuses this shape via `messages.length <= cached.messageCount`, so the compaction branch was simply missing the equivalent guard. That framing made it obvious this is a missing symmetry rather than a new special case, and it's why the change is one condition instead of a workaround.

Also worth recording — your PR's analysis corrects the issue's own. #712 attributes the loss to `continuation`; it's actually `compaction`. Post-#705, a changed prefix on a growing history already returns `diverged`, so `continuation` couldn't produce this shape. Anyone reading the issue later should trust the PR.

I added one commit on top (d029f94e): an end-to-end regression test. Your three tests pin the classification, which is right, but the data loss only *happens* via `server.ts`'s `resumeFrom`-out-of-range fallback to `getLastUserMessage()` — and no lineage test can see that. The new test drives your reported shape through the real HTTP path with a mocked SDK and asserts the prompt contains the user's message. I checked it's load-bearing: with the guard removed it fails with

```
Expected to contain: "WHAT THE USER ACTUALLY ASKED"
Received: "思考已结束。"
```

which is your symptom exactly. Without it, someone could change that fallback later and restore the data loss while every lineage test stayed green.

Your second finding — `ORCHESTRATION_TAGS` including `thinking` and being applied to user-authored content via `flattenUserContent` — is confirmed and filed separately as #720, so this change stays one fix. You're right that it's not in the same category as `env` or `tool_exec`: those are harness-internal and a user never writes them, whereas `<thinking>` is a common CoT convention in presets. The likely fix is scoping the stripping to assistant/tool-generated content, which preserves the original leak protection since replayed thinking is assistant-originated. Input welcome there if you have a view.

Thanks again — the SDK transcript evidence and the before/after turn counts turned what would normally be a long investigation into a short one.

rynfar:
Landed via #719 — closing this one since the commit is already on `main` with your authorship (8c308995).

PR #741, opened 2026-08-01
feat: Hermes Agent integration example plugin + docs

devsuitup:
## What

- **`examples/hermes-plugin/`** — a ready-to-install [Hermes Agent](https://hermes-agent.nousresearch.com) user plugin (`meridian-affinity`) that injects on every LLM request the headers Meridian understands:
  - `x-session-affinity` — session/lineage continuity → prompt-cache reuse
  - `x-request-id` — kanban task id, for `/telemetry/requests` correlation
  - `x-meridian-source` — `hermes-<profile>`, for per-profile log filtering
  - `x-opencode-effort` — per-profile reasoning-effort defaults with env override
- **`docs/agents.md`** — a "Hermes Agent" section (base_url setup, per-profile caveat, pointer to the plugin).

## Why

Hermes drives Meridian through the default `opencode` adapter without any session header. Every agentic turn ending in a `tool_result` — i.e. virtually every turn — is then classified as an independent session: client `cache_control` blocks are stripped and the whole history is replayed into a fresh SDK session, so cache creation is re-billed on every turn.

Measured on a real Hermes kanban workload (46-min implementer run, Claude Max subscription): average cache hit **47%**, 5.35M cache-creation tokens for that single run. After installing this plugin: **`cache=100% · cache_read=63k · cache_write=79`** per turn — cache writes divided by roughly the number of turns.

Nothing changes on the Meridian side — this is documentation plus a client-side example, in the spirit of `examples/opencode-plugin`.

## Testing

- Plugin deployed on a 10-profile Hermes fleet routing through Meridian 1.57.1 (passthrough mode), headers verified present in `/telemetry/requests` and session continuity confirmed in proxy logs (`lineage=continuation`).
- `python -m py_compile` clean; no Meridian code touched.


rynfar:
Landed in #762 (8b789e4e) with your commits squashed into one that keeps you as `Author`, plus `Co-authored-by: Jean-Baptiste <jean-baptiste@konvex-it.fr>` on the squash.

Closed by the merge trailer rather than as a rejection — `main` requires signed commits, which a fork PR can't satisfy from your side. Nothing to do with the change.

The plugin is genuinely good, and the comments are the reason. The string-vs-dict return at the tool-call boundary, the native-toolset requirement for platform sessions, and "a description that summarises the result invites the model to produce it from context instead of calling" — those are all things you only learn by shipping it and watching it fail. I verified all four injected headers are real and read by Meridian (`opencode.ts:22`, `server.ts:3604`, `server.ts:822`, `server.ts:933`).

Four things fixed on top:

1. **A compiled `.pyc` was committed** (`__pycache__/__init__.cpython-39.pyc`). The root cause was `.gitignore` having no Python rules at all — so removing the file alone would have let it return. Both fixed.

2. **The adapter claim wasn't universally true.** The docs said requests "are handled by the default `opencode` adapter", which holds only when `MERIDIAN_DEFAULT_AGENT` is unset. Hermes sends `User-Agent: python-httpx`, matching no heuristic, so it falls through to whatever that variable names. On my own instance that's `pi` — and installing your plugin *switches* it to `opencode`, because `x-session-affinity` resolves there in `detect.ts:174`. Usually what you want, since those are the headers that adapter reads, but it's a real switch and it was silent. Now documented in both the docs and the plugin README.

3. **`usage_status` stayed visible but failed on every call when the proxy sets `MERIDIAN_API_KEY`.** Your `check_fn` probed `/health`, which is deliberately exempt from auth, while the tool calls `/v1/usage/quota`, which is behind `app.use("/v1/*", requireAuth)`. So the availability check passed and every invocation 401'd — precisely the case that check exists to prevent. It now probes the endpoint it depends on and sends `x-api-key` when configured. Worth knowing for your own fleet if any of those proxies are keyed.

4. **Squashed the six commits** so the `Co-Authored-By: Claude` trailers don't reach `main` — this repo's CLAUDE.md disallows AI attribution lines. Your authorship is unchanged.

One gap left open deliberately: these docs cover prompt-cache continuity but not Extra-Usage metering. Hermes' `# Finishing the job` block reads as an autonomous coding-agent harness and gets metered as Extra Usage rather than under Max — a separate problem with a separate fix that isn't published yet, so there was nothing to link. Being tracked independently; if you've seen metering surprises on that fleet, that's the cause rather than anything in this plugin.

Thanks — the cache-hit measurement (47% → ~100% on a real 46-minute run) is exactly the kind of evidence that makes a change easy to accept.

Issue #634, opened 2026-07-16
memory: false in sdk-features.json is silently ignored when claudeMd: "off" (settings/settingSources coupling)


michaelkrupp:
## Summary
In `buildQueryOptions` (`src/proxy/query.ts`), the SDK `settings` object — which carries
`autoMemoryEnabled` — is bundled into the *same* conditional spread as `settingSources`:

```ts
...(settingSources && settingSources.length > 0 ? {
  settingSources,
  settings: {
    autoMemoryEnabled: ctx.memory ?? true,
    autoDreamEnabled:  ctx.dreaming ?? false,
  },
} : {}),
```

When `sdkFeatures.claudeMd === "off"` (the default), `settingSources` resolves to `[]`, so the
`length > 0` gate drops the **entire** object — including `settings`. As a result
`autoMemoryEnabled` is never sent to the Claude Agent SDK, which then falls back to its built-in
default (auto-memory **on**) and injects the auto-memory index (`MEMORY.md`) into every session.

Net effect: **`memory: false` in `~/.config/meridian/sdk-features.json` has no effect unless
`claudeMd` is `"project"` or `"full"`.** Users cannot disable auto-memory without also enabling
CLAUDE.md loading — the two are unintentionally coupled.

Notably, `DEFAULT_FEATURES.memory` is already `false`, so even the intended default (memory off)
is defeated: out of the box, auto-memory is on.

## Environment
- `@rynfar/meridian` 1.49.1 (also reproduces on 1.50.0 — latest at time of writing)
- `opencode-with-claude` 1.6.17 (adapter: `opencode`)
- Claude Agent SDK 0.2.x, `CLAUDE_CODE_ENTRYPOINT=sdk-ts`

## Reproduction
1. `~/.config/meridian/sdk-features.json`:
   ```json
   { "opencode": { "memory": false, "claudeMd": "off" } }
   ```
2. Start opencode via the meridian-backed anthropic provider; begin a session.
3. Observe the system context still contains the `# Memory` / "persistent file-based memory"
   section and an injected `MEMORY.md` listing.
4. Set `"claudeMd": "project"` → memory is correctly disabled. This is the only workaround,
   and it forces project CLAUDE.md/settings loading.

## Expected
`memory: false` disables auto-memory regardless of `claudeMd`, since the SDK accepts `settings`
(a `--settings` flag domain) independently of `settingSources` (file domains).

## Actual
[...]

PR #748, opened 2026-08-03
feat: add webFetchPreflight toggle for the WebFetch domain check

wayniacal:
Before WebFetch retrieves a URL, the subprocess sends the target hostname to `api.anthropic.com/api/web/domain_info` to test it against a blocklist. Upstream exempts that check from `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`, and `skipWebFetchPreflight` is a settings-file key — but Meridian passes `settingSources: []`, so no settings file is ever loaded. There was no way to turn it off through Meridian.

The SDK already accepts `skipWebFetchPreflight` in the `--settings` domain, which Meridian populates for the memory controls, so this reuses that path.

## Change

`webFetchPreflight` becomes a per-adapter feature, defaulting to **on** — behaviour is unchanged unless you flip it.

It is emitted on every request rather than only when disabled, for the same reason the memory keys are: an omitted key falls back to the subprocess default, which is what #634 hit.

## Verification

Client CLI pointed at Meridian on loopback, so the only reason that process opens a public connection is the preflight:

| | public peers opened by the client | answer |
|---|---|---|
| preflight on | Cloudflare (example.com) **+ api.anthropic.com** | `Example Domain` |
| preflight off | Cloudflare only | `Example Domain` |

WebFetch still works with the check off.

End to end through the proxy, reading the spawned subprocess's argv:

```
--settings {"autoMemoryEnabled":false,"autoDreamEnabled":false,"skipWebFetchPreflight":true}
```

Per-adapter routing holds — with the flag set only on `passthrough`, an `opencode` request still emitted `skipWebFetchPreflight:false`.

`npm test`: 11 suites, 2472 passing, 0 failing. `tsc --noEmit` clean.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -139,6 +139,20 @@ Configure per-adapter at **`/settings`** in the Meridian web UI. Changes take ef
 | **Thinking** | disabled / adaptive / enabled | Extended thinking mode for complex reasoning |
 | **Thinking Passthrough** | on / off | Forward thinking blocks to the client for display |
 | **Shared Memory** | on / off | Share memory directory with Claude Code (`~/.claude`) instead of isolated storage |
+| **WebFetch Preflight** | on / off | Check each WebFetch hostname against the Anthropic blocklist before fetching (default on) |
+
+### WebFetch preflight
+
+Before the WebFetch tool retrieves a URL, the subprocess sends the target
+hostname to `api.anthropic.com/api/web/domain_info` to check it against a
+safety blocklist. Only the hostname goes — not the path or the page — and the
+result is cached per host for five minutes.
+
+The check runs on every provider and is not covered by
+`CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`, so turning it off here is the only
+way to keep fetch targets local. With it off, WebFetch retrieves any URL
+without consulting the blocklist; if that matters for your deployment, pair it
+with WebFetch tool permissions on the calling agent.
 
 ### System prompts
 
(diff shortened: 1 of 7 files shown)


rynfar:
Thanks for this — the feature is yours and it's landed in #752 (e2aa19ff), with your commit cherry-picked so you stay as `Author` and the squash carries `Co-authored-by: Wayne <wayne@petml.com>`.

This closed automatically from the merge commit rather than as a rejection. The reason it couldn't merge from your fork is that `main` requires signed commits, which a fork PR can't satisfy from your side — a repo constraint, nothing about the change. Your `--settings` argv capture was the right check, and it's what confirmed the plumbing works.

One thing the follow-up commit adds. The preflight lives inside the SDK's **built-in** `WebFetch`, so it only fires when the Meridian-spawned subprocess runs that tool itself. Passthrough adapters send `tools: []` (the SDK's "disable all built-ins") and internal ones put `WebFetch` in `disallowedTools` — `cherry` is the only adapter that unblocks the built-in web tools (#481), and it had no settings UI at all. So the toggle was rendering on seven adapters where it does nothing and none of the one where it does. #752 exposes `cherry` (and `codex`, missing since #654) and asserts the scope in an E2E so it can't rot.

On the network table: I think that measured your client process rather than Meridian's subprocess — the subprocess contacts `api.anthropic.com` for the model call regardless, so that peer can't appear and disappear with this flag. #752 uses argv plus tool-flag capture instead, which is the only signal that separates the two. `scripts/e2e-webfetch-preflight.mjs` reproduces it with a stubbed `claude` and an isolated `HOME` — no tokens, no Max subscription.

Appreciate you digging into the `settingSources: []` interaction; that was the non-obvious part.

PR #749, opened 2026-08-03
feat: quiet the subprocess's non-essential outbound traffic

wayniacal:
The Claude Code CLI Meridian spawns is infrastructure rather than somebody's editor: it runs headless, nobody reads its usage metrics, its crash reports describe a process the operator never launched by hand, and `/feedback` and the session-quality survey have no interactive session to report on. Auto-updating underneath a running proxy is version skew, not a feature.

## Change

Seven documented opt-outs set on the subprocess env:

| Variable | Effect |
|---|---|
| `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` | Metrics, error reports, feedback uploads, surveys |
| `DISABLE_TELEMETRY` | Usage metrics |
| `DISABLE_ERROR_REPORTING` | Crash reports to the third-party tracker |
| `DISABLE_FEEDBACK_COMMAND` | `/feedback`, `/bug`, `/share` |
| `CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY` | Session-quality survey |
| `DISABLE_AUTOUPDATER` | CLI self-update mid-run |
| `CLAUDE_CODE_DISABLE_OFFICIAL_MARKETPLACE_AUTOINSTALL` | Plugin marketplace auto-install |

Spread **before** the inherited env, so any value the operator sets wins — including setting one to `0` to opt back in. Tests cover the defaults, both opt-back-in paths (`cleanEnv` and `envOverrides`), and passthrough.

Deliberately not included: `CLAUDE_CODE_SKIP_PROMPT_HISTORY`, which would stop the subprocess writing `~/.claude/projects/*.jsonl` — the files session resume and recovery read.

This does not cover the WebFetch domain check, which upstream exempts from `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`. That is #748.

`npm test`: 11 suites, 2471 passing, 0 failing. `tsc --noEmit` clean.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -47,6 +47,34 @@ Environment variables, endpoints, authentication, SDK feature toggles, passthrou
 
 †Sonnet 1M requires Extra Usage on all plans including Max ([docs](https://code.claude.com/docs/en/model-config#extended-context)). Opus 1M is included with Max/Team/Enterprise at no extra cost. Fable 1M is also included at no Extra Usage cost, verified live on both Max and Team.
 
+### Subprocess traffic
+
+Meridian runs the Claude Code CLI as a headless subprocess, so it sets these
+on that process by default:
+
+| Variable | Effect |
+|---|---|
+| `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC` | Usage metrics, error reports, feedback uploads, session-quality surveys |
+| `DISABLE_TELEMETRY` | Usage metrics |
+| `DISABLE_ERROR_REPORTING` | Crash reports to the third-party error tracker |
+| `DISABLE_FEEDBACK_COMMAND` | `/feedback`, `/bug`, `/share` uploads |
+| `CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY` | Session-quality survey |
+| `DISABLE_AUTOUPDATER` | CLI self-update underneath a running proxy |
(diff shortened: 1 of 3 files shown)


rynfar:
Thanks — this landed in #757 with your commit cherry-picked, so you're preserved as `Author` and the squash carries `Co-authored-by: Wayne <wayne@petml.com>`.

It closed automatically from the merge commit rather than as a rejection. The reason it couldn't merge from your fork is that `main` requires signed commits, which a fork PR can't satisfy from your side — a repo constraint, nothing to do with the change.

Two things I checked before merging, both held:

- All seven variable names exist in the bundled CLI (`strings` over `claude.exe`) — 7/7 present, so none of them are silently inert.
- The env precedence actually delivers the documented opt-back-in. `QUIET_SUBPROCESS_ENV` spreads at L374, inherited env at L384, `envOverrides` at L406 — so setting one to `0` genuinely wins. Spread the other way it would have read as supported while doing nothing, and your tests cover both paths.

Excluding `CLAUDE_CODE_SKIP_PROMPT_HISTORY` was the right call and clearly deliberate — those `~/.claude/projects/*.jsonl` files are exactly what session resume and recovery read.

One thing added on top: a docs cross-link. Your exemption paragraph about the WebFetch preflight was already correct, but #752 landed separately and neither section pointed at the other, so a reader hitting one had no way to know the other existed. They link both ways now, and the preflight section notes the exemption only reaches the wire on `cherry` — the sole adapter that lets the subprocess run the built-in WebFetch at all.

Two solid contributions in a day; both are in 1.59.0. Appreciate it.

Issue #441, opened 2026-04-27
OpenCode provider can inherit Claude Code PowerShell tool and call unavailable PowerShell


BenIsLegit:
## Summary

When using Meridian as the Claude provider for OpenCode on Windows, Claude can attempt to call the Claude Code SDK's `PowerShell` tool even though OpenCode only exposes its own lowercase tool set (`bash`, `read`, `glob`, etc.). OpenCode then reports:

```text
⚙ invalid [tool=PowerShell, error=Model tried to call unavailable tool 'PowerShell'. Available tools: invalid, question, bash, read, glob, grep, edit, write, task, webfetch, todowrite, skill, quota_status.]
```

I believe this is triggered when the user's Claude Code settings/environment enables:

```json
"CLAUDE_CODE_USE_POWERSHELL_TOOL": "1"
```

## Environment

- OS: Windows
- Client: OpenCode using Meridian as the Claude provider
- Meridian repo: `rynfar/meridian`
- OpenCode tool set includes `bash`, but not `PowerShell`/`powershell`

## What I Verified

Meridian currently passes most environment variables through to the Claude Agent SDK subprocess:

- `src/proxy/server.ts` strips `ANTHROPIC_API_KEY`, `ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, and `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS`, then keeps the remaining env as `cleanEnv`.
- `src/proxy/query.ts` passes `cleanEnv` into the SDK `query()` options as `env`.
- `CLAUDE_CODE_USE_POWERSHELL_TOOL` is not stripped or overridden.

OpenCode does not advertise a `PowerShell` tool:

- OpenCode's built-in shell tool is `bash` (`packages/opencode/src/tool/bash.ts`).
- OpenCode's repair logic lowercases mismatched tool names only if the lowercase version exists (`packages/opencode/src/session/llm.ts`). `PowerShell` lowercases to `powershell`, but OpenCode has no `powershell` tool, so it routes to the `invalid` tool and reports the unavailable-tool error.

Meridian's OpenCode passthrough mode registers OpenCode's request tools as MCP tools, but the inherited SDK setting can still make Claude aware of/use the SDK-side `PowerShell` tool name, which OpenCode cannot execute.

## Expected Behavior

[...]

PR #750, opened 2026-08-03
feat: gate claude.ai connectors behind an opt-in flag

wayniacal:
Outside passthrough mode, the subprocess fetches the MCP connectors attached to the account's claude.ai profile from `/v1/mcp_servers` and connects each one through `mcp-proxy.anthropic.com`. Their tools then join the session alongside the ones Meridian registered — so a caller that negotiated a tool surface gets third-party tools it never asked for, plus an outbound fetch describing the account.

`ENABLE_CLAUDEAI_MCP_SERVERS: "false"` was already set, but only in passthrough mode. Whether your Gmail shows up in Droid currently depends on a passthrough default that has nothing to do with connectors.

## Change

`claudeAiConnectors`, defaulting to **off**, applied to every adapter.

Passthrough continues to force it off whatever the flag says: there the client executes tools and has no way to run one that exists only inside the subprocess — the same mismatch behind #441.

## Behaviour change

Adapters that do not use passthrough by default — `cherry` (`usesPassthrough(): false`), `droid` (`resolvePassthrough(false)`), and anything running `MERIDIAN_PASSTHROUGH=0` — previously loaded connectors and now need the flag. Adapters that default to passthrough, including `opencode`, are unaffected: they were already covered by the old line.

The existing `does not disable Claude.ai MCP servers in normal mode` test asserted the old default and is updated, with new cases for the opt-in and for passthrough overriding it.

## Verification

Same proxy, `MERIDIAN_PASSTHROUGH=0` (the configuration that previously loaded connectors), subprocess debug via `CLAUDE_CODE_DEBUG_LOGS_DIR`:

**Default (`claudeAiConnectors: false`)**
```
[claudeai-mcp] Disabled via env var
```

**Opted in (`claudeAiConnectors: true`)**
```
[claudeai-mcp] Fetching from https://api.anthropic.com/v1/mcp_servers?limit=1000
[claudeai-mcp] Fetched 3 servers
MCP server "claude.ai Google Drive":    Using claude.ai proxy at https://mcp-proxy.anthropic.com/v1/mcp/…
MCP server "claude.ai Gmail":           Using claude.ai proxy at https://mcp-proxy.anthropic.com/v1/mcp/…
MCP server "claude.ai Google Calendar": Using claude.ai proxy at https://mcp-proxy.anthropic.com/v1/mcp/…
```

`npm test`: 11 suites, 2470 passing, 0 failing. `tsc --noEmit` clean.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -139,6 +139,22 @@ Configure per-adapter at **`/settings`** in the Meridian web UI. Changes take ef
 | **Thinking** | disabled / adaptive / enabled | Extended thinking mode for complex reasoning |
 | **Thinking Passthrough** | on / off | Forward thinking blocks to the client for display |
 | **Shared Memory** | on / off | Share memory directory with Claude Code (`~/.claude`) instead of isolated storage |
+| **claude.ai Connectors** | on / off | Load the MCP connectors attached to your claude.ai account (default off) |
+
+### claude.ai connectors
+
+If your claude.ai account has connectors attached — Drive, Gmail, Calendar —
+the subprocess can fetch them from `/v1/mcp_servers` and connect each one
+through `mcp-proxy.anthropic.com`. Their tools then appear in the session
+alongside the ones Meridian registered.
+
+That is rarely what you want from a proxy: the calling agent negotiated a tool
+surface, and this adds third-party tools it never asked for, plus an outbound
+fetch describing the account. So it is **off by default**.
+
+Passthrough mode keeps it off regardless of this setting. There the client
+executes tools, and it has no way to run one that exists only inside the
+subprocess.
 
 ### System prompts
 
--- a/src/__tests__/query.test.ts
+++ b/src/__tests__/query.test.ts
@@ -297,10 +297,24 @@ describe("buildQueryOptions", () => {
     expect(env.ENABLE_CLAUDEAI_MCP_SERVERS).toBe("false")
   })
 
-  it("does not disable Claude.ai MCP servers in normal mode", () => {
+  it("disables Claude.ai MCP servers in normal mode too, by default", () => {
     const result = buildQueryOptions(makeContext({ passthrough: false }))
     const env = (result.options as any).env
+    expect(env.ENABLE_CLAUDEAI_MCP_SERVERS).toBe("false")
+  })
+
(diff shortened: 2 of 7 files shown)


rynfar:
Landed in #759 (370c286c), cherry-picked so you stay as `Author` with `Co-authored-by: Wayne <wayne@petml.com>` on the squash. Closed by the merge trailer, not as a rejection — the fork + signed-commits constraint again.

**First, a correction I owe you.** I initially doubted your verification, because `"Disabled via env var"` doesn't appear in the bundled CLI's string table and I assumed you'd paraphrased. It appears verbatim at runtime — the string is constructed or compressed in the binary, so `strings` was the unreliable instrument, not your transcript. My reproduction matched yours exactly.

The diagnosis is the valuable part here, more than the flag: `ENABLE_CLAUDEAI_MCP_SERVERS` was set only in passthrough, so whether an account's Gmail reached a Droid session depended on a passthrough default that has nothing to do with connectors. The coupling was the defect; the default was the smaller question.

Kept your uniform default-off rather than carving out `cherry`. A per-adapter exception would make "which adapter am I on?" load-bearing for whether your email is exposed — the same invisible-behaviour problem #752 just fixed. `ADAPTER_DEFAULTS` is there if that call changes later.

**One change on top.** Your opt-in expressed "on" by omitting the env var, which makes it mean "whatever the subprocess defaults to" rather than "on" — the #634 failure mode, and `query.ts` carries that lesson a few lines above where your change lands. An upstream default flip would silently re-enable connectors for everyone opted in, and disable them for everyone else, with nothing failing either way. It's now always emitted, `"true"` or `"false"`.

That fix had its own trap, so I verified rather than assumed: an `ENABLE_*` flag read by *presence* would mean explicit `"true"` disables, turning the opt-in into a silent no-op. Reproduced your setup:

```
claudeAiConnectors=true   → [claudeai-mcp] Fetching from …/v1/mcp_servers?limit=…
                            [claudeai-mcp] Fetched 0 servers
claudeAiConnectors=false  → [claudeai-mcp] Disabled via env var
```

Reads by value; `"true"` enables. Also resolved six mechanical conflicts against #752/#754, which had landed since you opened this.

Four solid contributions in two days, and the two diagnoses — the passthrough coupling here, and the billing header in #751 — were both things nobody would have found by reading the code looking for bugs. Thanks.

PR #751, opened 2026-08-03
fix: don't treat transport metadata as a system instruction

wayniacal:
Claude Code passes a header through the request body. Its `system` arrives as an array of three blocks:

```
[0] "x-anthropic-billing-header: cc_version=2.1.220.8b8; cc_entrypoint=sdk-cli;"
[1] "You are a Claude agent, built on Anthropic's Claude Agent SDK."
[2] <the actual system prompt>
```

`server.ts` joined every text block verbatim, so block [0] landed at the top of the SDK subprocess's system prompt. It is a header the client smuggles through the body, not something written for a model to read — and it is meaningless downstream anyway, since the subprocess authenticates as its own subscription and sends its own billing header. The client's copy describes a request that is never made.

## Change

Extracts the flattening from `server.ts` into a pure `extractSystemText()` in the `messages.ts` leaf module, and drops header-shaped blocks.

## Scope

Deliberately narrow. The pattern is `/^\s*x-anthropic-[a-z0-9-]*header\s*:/i`, anchored at block start:

| input | result |
|---|---|
| `x-anthropic-billing-header: …` | filtered |
| `X-Anthropic-Billing-Header : …` | filtered |
| `x-anthropic-version: 2023-06-01` | **kept** — name doesn't end in "header" |
| `Always send the x-anthropic-billing-header when relaying.` | **kept** — prose |
| header not at the start of the block | **kept** |

## Safety

`query.ts` decides `usePreset` via `codeSystemPrompt ?? (hasSettings || (!passthrough && !!systemContext))`, so a shorter `systemContext` could in principle change preset selection. It cannot in practice: all six call sites pass `codeSystemPrompt: sdkFeatures.codeSystemPrompt`, which is typed `boolean` and always merged over `DEFAULT_FEATURES`, so the `??` never falls through.

This is hygiene, not a behaviour fix — I have no measurement showing the header changes model output.

`npm test`: 11 suites, 2473 passing, 0 failing. `tsc --noEmit` clean.

### DIFF
--- a/src/__tests__/messages.test.ts
+++ b/src/__tests__/messages.test.ts
@@ -2,7 +2,7 @@
  * Unit tests for message parsing utilities.
  */
 import { describe, it, expect } from "bun:test"
-import { frameReplayTurns, normalizeContent, getLastUserMessage, extractAdvisorModel, stripAdvisorTools, stripNonStandardStreamFields, consolidateMultimodalOntoLastUser, buildToolUseIndex, describeToolCall } from "../proxy/messages"
+import { frameReplayTurns, normalizeContent, getLastUserMessage, extractAdvisorModel, stripAdvisorTools, stripNonStandardStreamFields, consolidateMultimodalOntoLastUser, buildToolUseIndex, describeToolCall, extractSystemText } from "../proxy/messages"
 
 const img = (id: string) => ({ type: "image", source: { type: "base64", media_type: "image/png", data: id } })
 function userMsg(content: unknown) {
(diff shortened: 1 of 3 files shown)


rynfar:
Landed in #758 (59931d0e) with your commit cherry-picked, so you stay as `Author` and the squash carries `Co-authored-by: Wayne <wayne@petml.com>`.

Closed automatically by the merge commit rather than as a rejection — `main` requires signed commits, which a fork PR can't satisfy from your side.

Three things checked before merging, all held:

- The filter matches your table exactly — 8/8, including `x-anthropic-version` kept because the name doesn't end in "header", and prose mentioning a header left alone.
- The header is real, not inferred: `x-anthropic-billing-header`, `cc_entrypoint`, `cc_version` and `sdk-cli` are all in the bundled CLI.
- **Session resume survives it.** This was the one that could have been ugly and invisible: `systemContext` gets shorter after filtering, so if it fed session identity, every in-flight conversation would have failed to resume the moment this shipped. It doesn't — `getConversationFingerprint` hashes first-user-message + cwd, `computeLineageHash` takes messages only, and `fingerprint.ts:39` documents *why* systemContext is deliberately excluded. Worth knowing the margin was there.

Your `usePreset` analysis also holds — all six call sites pass a `boolean` merged over `DEFAULT_FEATURES`, so the `??` never falls through to the `systemContext` check.

And the framing was right: transport metadata isn't an instruction, and you said plainly it's hygiene rather than a measured behaviour fix. That's the correct justification to put on it — I kept that line in the PR body rather than upgrading the claim.

PR #764, opened 2026-08-04
fix: recognize shorter Claude limit wording

justprosh:
## Context

I exhausted the five-hour quota on the preferred profile in a two-profile `priority` setup and expected Meridian to switch to the fallback profile. Instead, the request failed with:

```text
Claude Code returned an error result: You've hit your limit · resets 6:40pm (UTC)
```

Meridian returned this as `500 api_error`, so priority routing did not mark the profile exhausted or try the fallback.

This is a shorter wording variant of the quota error handled in #680: the classifier recognizes `You've hit your session limit`, but not `You've hit your limit`.

## Fix

- classify the shorter wording as `429 rate_limit_error`
- add classifier and priority-failover regression tests

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -490,6 +490,12 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     expect(r.status).toBe(429)
   })
 
+  it("maps the CLI's shorter 'You've hit your limit' wording to rate_limit_error", () => {
+    const r = classifyError("Claude Code returned an error result: You've hit your limit \u00b7 resets 6:40pm (UTC)")
+    expect(r.type).toBe("rate_limit_error")
+    expect(r.status).toBe(429)
+  })
+
   it("maps 'usage limit reached' to rate_limit_error", () => {
     const r = classifyError("usage limit reached | resets at 5pm")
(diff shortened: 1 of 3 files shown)


rynfar:
Thanks @justprosh — this shipped in #788.

To explain the close, since GitHub gives you no signal about it: `main` requires signed commits, so PRs from forks can't be merged directly no matter how green the checks are. It surfaces only as `mergeStateStatus: BLOCKED` with no reason attached, and there was nothing you could have done from your side. The way through is for a maintainer to cherry-pick, which re-signs the commit while preserving you as its author — your commit is on `main` unchanged, and the squash carries a `Co-authored-by:` trailer for you.

Worth noting: @WarGloom opened #787 in the same window for `"You've hit your weekly limit"`. Neither wording subsumes the other — `"hit your limit"` isn't a substring of `"hit your weekly limit"` — so both were real, and both commits are in #788.

Two variants in one window suggested the literal list would keep growing, so there's a commit on top replacing it with `/hit your (?:[\w-]+ )?limit/`, covering daily, monthly, and 5-hour too. Your placement of the `failureMessage` reset in the file-level `beforeEach` was the correct scope and is what the merged branch keeps — #787 had it at describe scope, where it leaked into two later describes.

PR #768, opened 2026-08-07
fix(passthrough): never leave a passthrough continuation unanswered

justprosh:
## Problem

A passthrough turn can complete as a success — `end_turn`, HTTP 200, `error: null` — while delivering nothing the client can act on. Four causes:

1. **Early stop** leaves the stored session on an interrupted tail; resuming it yields meta text (`No response requested.`) or nothing.
2. **Client abort** never settled the session; every following continuation came back empty, each empty turn becoming the next tail.
3. **Spent deny at the boundary seam**: the persisted `end your turn now` deny sits immediately before the continuation's user turn, and the model obeys it — empty text block, clean `end_turn`.
4. **90s idle limit** kills a long-thinking turn; the kill is classified `unknown`, captured tool calls are dropped, the client gets a finished empty message.

## Fix

- **Boundary capture** — store the uuid of the last persisted deny; continuations fork there (`resumeSessionAtUuid`) instead of attaching to the interrupted tail. Prompt cache survives (91–99% hit).
- **Settlement** — the client-abort path now stores or evicts the session and resolves the pending store (`clientAbortDisposition`, pure).
- **Discharge** — boundary continuations get a lead-in stating the deny is spent (`framePassthroughContinuation`); bare resumes are byte-identical to today.
- **Idle limit override** — `MERIDIAN_UPSTREAM_IDLE_MS`, default unchanged (90s).
- **Outcome guard** — every completed turn is classified in `turnOutcome.ts`: a terminal envelope must carry text or a tool call; `thinking` doesn't count. A silent turn — or an announce-only one on a deny-boundary continuation of a short conversation — gets one recovery turn, forked from the boundary, with a nudge that discharges the deny. Kill switch `MERIDIAN_SILENT_TURN_RECOVERY=0`. Failed turns now emit the error before `message_stop` and report `max_tokens` instead of claiming `end_turn`.

Telemetry: `response.silent_turn`, `response.announce_turn`, `passthrough.client_abort_settled`.

## Verification

Early stop reproduced on a harness (Sonnet 10/10 degenerate → 0/20 fixed). Causes 2–4 diagnosed from production transcripts and telemetry, unit-tested at their decision points. The guard has an injection E2E (`MERIDIAN_DEBUG_FORCE_SILENT_TURN`, `scripts/e2e-silent-turn.mjs`): recovery off — 2/2 fail, on — 2/2 recover. Full suite, typecheck, build green.

### DIFF
--- a/E2E.md
+++ b/E2E.md
@@ -90,6 +90,7 @@ kill $(lsof -ti :3456)
 | E35 | [SDK boundary assumptions (#694/#708/#710)](#e35-sdk-boundary-assumptions-694708710) | **Automated**: `bun scripts/e2e-sdk-boundary.mjs` — real SDK: rate-limit reset units land in a sane window, every live content-block type is classified for hashing, resume survives a client dropping thinking blocks, and reports whether the gitStatus block still misstates its provenance. **Run after any `@anthropic-ai/claude-agent-sdk` bump** and before releases touching lineage, rate limits, or the system prompt | 2026-07-29 |
 | E36 | [Client detection after an upgrade (#733)](#e36-client-detection-after-an-upgrade-733) | **Automated**: `bun scripts/e2e-client-detection.mjs` — drives each installed client against a local stub, captures its real headers, and asserts the adapter Meridian resolves. **Run after upgrading any client**; costs no tokens | 2026-07-31 |
 | E37 | [WebFetch preflight scope (#748)](#e37-webfetch-preflight-scope-748) | **Automated**: `bun scripts/e2e-webfetch-preflight.mjs` — stubbed `claude` + isolated HOME: the toggle reaches the right adapter's `--settings`, and only `cherry` can actually run the built-in WebFetch, so the documented scope is asserted rather than assumed. **Run before releases touching sdkFeatures, query settings, or tool config**; costs no tokens | 2026-08-03 |
+| E38 | [Silent turns (#768)](#e38-silent-turns-768) | **Automated**: `bun scripts/e2e-silent-turn.mjs` — real CLI, SSE mode, second turn of a fresh session: asserts the outcome (did the client get text or a tool call?) rather than any one cause, so a cause not yet found fails it too. Pair a `MERIDIAN_DEBUG_FORCE_SILENT_TURN=1` run against `MERIDIAN_SILENT_TURN_RECOVERY=0` for the before/after. **Run before any release touching the passthrough tool loop, prompt assembly, or session resume** | 2026-08-11 |
 
 | P1 | [Profile: List & Auth Status](#p1-profile-list--auth-status) | `/profiles/list` returns profiles with emails, login status, auth timestamps | - |
 | P2 | [Profile: Switch via API](#p2-profile-switch-via-api) | `POST /profiles/active` switches profile; health endpoint reflects new email | - |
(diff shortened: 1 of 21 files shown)


rynfar:
digging into this

justprosh:
@rynfar thanks! can be a little bit messy, stumbled multiple times on different occasions and tried to fix

rynfar:
Thanks for this @justprosh — the diagnosis is right and the four structural fixes are going in as you wrote them. Continued in #793, where your seven commits are cherry-picked with your authorship intact (`main` requires signed commits, so fork PRs sit at `BLOCKED` with no visible reason — nothing you could have done from your side; the squash carries a `Co-authored-by:` trailer for you).

Your harness is what made the rest of this possible. Measuring the outcome rather than any one cause was the right call, and the `MERIDIAN_DEBUG_FORCE_SILENT_TURN` injection is what turned "we think it works" into something checkable. Nearly everything below was found by running it.

**Three bugs in the guard layer**, all confirmed against a live model:

1. Recovered content arrived *after* the terminal `message_delta`. A probe on a forced-silent turn put the recovered answer at index 9 and the `end_turn` delta at index 8 — so any client that finalizes at `message_delta` discarded the repair. `message_delta` is now withheld the way `message_stop` already was. Related: the unseen-tool-use path emitted a second `message_delta`, so a recovery that made a tool call gave one message two conflicting endings.

2. The recovery fork was never persisted. `storeSession` runs before the recovery block, so the fork's answer lived in a session nothing pointed at — the next request resumed the silent tail, and a recovery tool call would return a `tool_result` for a `tool_use` the resumable session never saw. Now the fork's id and boundary are captured and re-stored.

3. The recovery query bypassed `guardUpstreamIdle`, so a stalled recovery could hold an open SSE response and its concurrency slot indefinitely.

**The announce removal, which reverses a deliberate call of yours**, so it deserves the reasoning rather than just the diff.

`fad37764` argued that a false positive costs one redundant closing turn while a false negative costs the conversation. That trade is sound if the numbers hold. Measured on the exact shape it targets — deny-boundary continuation, short text, short conversation — it fired on **12 of 12 healthy turns with zero genuine silences**. In one of those the recovery came back with a tool call, so a turn that had already answered in prose went out as `stop_reason: tool_use` and the client would have executed it.

That window is the shape of an ordinary concise answer after a tool round-trip, which is among the most common things a passthrough client does. So the false-positive cost was not a redundant turn — it was acting on a question already answered, plus an extra model turn on roughly every opening exchange.

Text is treated as an answer now. Your silence invariant is untouched and does the work: forced-silent turns go 3 detected / 3 recovered / 0 reaching the client, with a third turn proving the session still resumes. Natural traffic is 0 of 12 where it used to be 12 of 12.

If the announce stall is real — and your transcript suggests you saw one — it needs a signal that distinguishes it from a short answer. Length and conversation depth don't. If you have the transcript, I'd genuinely like to look: something in that turn that isn't its length may separate the two, and it'd be worth adding back on that basis.

**Small things in the harness**, mentioned for completeness: `inst.stop?.()` is not a method (`ProxyInstance` exposes `close()`), so optional chaining swallowed it and a *passing* run never exited — the failure path's `process.exit(1)` hid it. The attribution counters read `console.debug` while `diagnosticLog.session()` writes to the ring buffer, so `recovered` was structurally always 0. And it defaulted to Sonnet despite E2E.md's Haiku convention. All fixed in #793, along with assertions for ordering and post-recovery resume.

PR #771, opened 2026-08-08
feat(profile): create the profile when `profile login` names an unknown one

Nowaker:

Nowaker is still iterating on this. It is opened early so the diff is visible;
it will be marked ready when it is actually ready.

---

`meridian profile login weekend-account` currently exits 1 with:

```
Profile not found. Run: meridian profile add weekend-account
```

Two commands for one intention, and the error does not say that the second
one is what you meant. This makes `login` create the profile when it names
an unknown one, after printing a red notice saying so.

### What it does

- An unknown ID prints a notice and falls straight into `profileAdd`, which
  stays the **single creation path** - config dir, `profiles.json` entry,
  auth flow and verification keep one definition of what a profile is made
  of, rather than gaining a second one that can drift.
- A **known** profile is unchanged, `oauth-token` profiles are unchanged.
  Both are covered by golden tests.
- The notice is red only on a TTY; piped output stays free of escape codes.

### The ID validation, and why it is on one branch only

`login <id>` on an unknown ID is the only path that turns an ID into a
directory under `~/.config/meridian/profiles`, so `login ../../etc` must not
become one. Validation therefore happens on that branch, before anything is
created.

`profileAdd` already had that check; it is now a shared `isValidProfileId`
rather than a third copy of the regex. An ID that **already** has a profile
is deliberately not re-validated, so `login` keeps working on entries written
by hand.

### Tests

The decision is a pure `planProfileLogin`, so all four outcomes are unit
tested without spawning `claude auth login`.

```
bun test src/__tests__/profile-login-plan.test.ts   8 pass, 0 fail
bun run test                                        full chain, 0 fail
tsc --noEmit                                        exit 0
```

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -148,7 +148,7 @@ ANTHROPIC_API_KEY=your-secret-key ANTHROPIC_BASE_URL=http://meridian-host:3456 o
 | `meridian profile add <name> --oauth-token [TOKEN]` | Add a headless profile from a `claude setup-token` value (prompts when `TOKEN` is omitted) |
 | `meridian profile list` (alias `profile ls`) | List all profiles and their auth status |
 | `meridian profile switch <name>` | Switch the active profile (requires running proxy) |
-| `meridian profile login <name> [--headless]` | Re-authenticate an expired profile (browser-login profiles only); `--headless` uses the URL/code flow |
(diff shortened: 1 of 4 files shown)


rynfar:
Hey there looks like youve been working on some things.  What is your plan with all of these PR's?  are you continueing ot work through them?

Nowaker:
@rynfar Yeah, I'm working on several more things, and want to submit them for review when I'm done with all of them. :)

rynfar:
sounds good!  look forward to see what you’ve got!

PR #772, opened 2026-08-08
feat(dev): MERIDIAN_CREDENTIALS_READONLY, for a second instance on shared credentials

Nowaker:

Nowaker is still iterating on this. It is opened early so the diff is visible;
it will be marked ready when it is actually ready.

---

`MERIDIAN_CREDENTIALS_READONLY=1` lets a **development** instance run beside a
production one, against the same credential files, without being able to
corrupt them.

### The hazard

Two Meridian processes holding the same OAuth refresh token. Refresh happens
proactively on boot and every 45s, and a `refresh_token` grant rotates the
token server-side - so whichever process refreshes first can invalidate the
other's copy. Recovery is an interactive `claude login` per account.

**Sharing the credential files is not opt-in.** `MERIDIAN_CONFIG_DIR`
relocates `settings.json` only; `profiles.json` is a module-level constant in
`profiles.ts`, so every instance on a machine reads the same profile list
pointing at the same `claudeConfigDir` directories. Any second instance
therefore holds the first one's real accounts from the moment it boots. This
flag is what stops it writing to them.

*(A separate change making `MERIDIAN_CONFIG_DIR` relocate the whole directory
is in progress and will be its own PR. It reduces the blast radius but does
not remove the need for this: a dev instance that deliberately shares
credentials is exactly the case this exists for.)*

### What the flag does

- The background refresh scheduler is **never armed** - never-arm rather than
  arm-and-refuse, so `isBackgroundRefreshActive()` stays an honest answer.
- The profile refresh loop and the auth keepalive are not scheduled. The
  keepalive matters despite looking read-only: it shells out to
  `claude auth status` against the profile's `CLAUDE_CONFIG_DIR`, which is the
  CLI's own opportunity to rotate and rewrite the credential file, in a
  subprocess this process cannot intercept.
- `refreshOAuthToken` and `ensureFreshToken` return before the network call.
  Refusing only the *write* is insufficient: the grant rotates the token
  server-side, so calling the endpoint and then declining to persist the
  replacement would invalidate the other instance's credentials **and** discard
  the only thing that could repair them.
- Every credential write is refused **at the store seam** and logged to stderr.
  The guard wraps both builders in `tokenRefresh.ts` and the design token
  store, so a call site nobody anticipated fails loudly rather than silently
  corrupting a token file the other instance depends on. The log names the
  operation and the store, never a value.
- The auth-status cache is additionally invalidated when the credential file's
  mtime changes, so a rotation performed by the *owning* instance is picked up
  on the next tick instead of up to a TTL later.
- The mode is announced once at startup.

`design.ts` is included because `defaultDesignTokenPath()` ignores
`MERIDIAN_CONFIG_DIR` entirely: a second instance writes the first's design
token no matter how its config dir is set, and that token carries a refresh
token that rotates on use. It is the concrete example of why the refusal
belongs at the store rather than on the timers.

**With the flag absent, behaviour is unchanged.** The mtime helper returns 0
without `stat()`ing, so every comparison is equal and the cache keeps exactly
its current time-based path.

### Measured, not assumed

Whether the Claude Agent SDK subprocess writes a rotated token back is outside
any guarantee this flag can make, so it was tested rather than reasoned about.

On 2026-08-07, against a live proxy (Linux, Claude Code 2.1.198,
`claude-haiku-4-5`), hashing one profile's `.credentials.json` around a single
request routed to it:

| | result |
|---|---|
| control window spanning two 45s ticks | file unchanged |
| after a real request (HTTP 200, genuine SDK round trip) | byte-identical - same SHA-256, mtime and size |

The subprocess is demonstrably willing to write that directory - the same
request updated `.claude.json`, `policy-limits.json`, `remote-settings.json`,
`backups/`, `jobs/` and `sessions/` - it simply did not touch
`.credentials.json`.

**Not measured:** whether it writes when refreshing an expired access token
itself, which needs either an ~8h wait or edits to a live credential file, and
which the proxy's own 5-minute refresh buffer reaches first in this
deployment. Documented as a stated limit in `docs/configuration.md` rather
than papered over.

### Tests

The flag skips both timers, the store refuses a direct write and logs, the
OAuth request is never sent, and the mtime change re-reads.

The mtime tests live in `models-auth-status.test.ts` because four files in the
main run call `mock.module("../proxy/models")`, which is global; that file
already runs as its own isolated invocation for the same reason.

```
bun test src/__tests__/credentials-readonly.test.ts   17 pass, 0 fail
bun test src/__tests__/models-auth-status.test.ts     15 pass, 0 fail
bun run test                                          full chain, 0 fail
tsc --noEmit                                          exit 0
```

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -48,8 +48,75 @@ Environment variables, endpoints, authentication, SDK feature toggles, passthrou
 | `MERIDIAN_SILENT` | `CLAUDE_PROXY_SILENT` | unset | Set to `1` to suppress startup output (used by embedding plugins) |
 | `MERIDIAN_PLUGIN_DIR` | — | `~/.config/meridian/plugins` | Plugin auto-discovery directory |
 | `MERIDIAN_PLUGIN_CONFIG` | — | `~/.config/meridian/plugins.json` | Plugin manifest path |
+| `MERIDIAN_CREDENTIALS_READONLY` | `CLAUDE_PROXY_CREDENTIALS_READONLY` | unset | Set to `1` to forbid this instance from refreshing or writing OAuth credentials. For a second instance sharing another's credential files — see [Read-only credentials](#read-only-credentials). |
 
 †Sonnet 1M requires Extra Usage on all plans including Max ([docs](https://code.claude.com/docs/en/model-config#extended-context)). Opus 1M is included with Max/Team/Enterprise at no extra cost. Fable 1M is also included at no Extra Usage cost, verified live on both Max and Team.
+
+### Read-only credentials
+
+`MERIDIAN_CREDENTIALS_READONLY=1` makes an instance incapable of modifying
+credentials. It exists so a second instance — a development build, a staging
+copy — can run beside a production one **against the same credential files**
+without being able to corrupt them.
+
+The hazard is two instances holding the same OAuth refresh token. Meridian
+refreshes proactively on boot and every 45s. If Anthropic rotates a refresh
+token on use, whichever instance refreshes first can invalidate the other's
+copy, and the only recovery is an interactive `claude login` per account.
+
+Sharing the files is usually not a choice. `profiles.json` is read from
+`~/.config/meridian/` unconditionally — `MERIDIAN_CONFIG_DIR` relocates
+`settings.json` only — so any second instance sees the same profile list
+pointing at the same credential directories. This flag is what makes that
+safe.
+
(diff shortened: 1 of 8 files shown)


rynfar:
Incorporated into main in #1064 (commit b7b820e9d5f42f1bff7693aa7b2cf8347a51024b), retaining original commit authorship. Thank you @Nowaker!

PR #773, opened 2026-08-08
feat(cli): print the dashboard when the port is already serving Meridian

Nowaker:

Opening this early so it is visible; I am still iterating on it. Please hold off
on review until this banner is gone.

---

## The problem

Running `meridian` on a machine where Meridian is already running prints a full
usage screen, then:

```
Error: Port 3456 is already in use

Another instance of Meridian may be running.
Try: meridian --port <other-port>
```

It says "another instance of Meridian **may** be running" without ever checking,
and then suggests moving out of its way. But asking what is running is not a
mistake, and on the overwhelmingly common path the process holding that port
*is* Meridian. So this answers instead: the `/` page, rendered for the terminal,
exit 0.

## What it does

**Probe before deciding.** On a port that will not bind, the CLI issues
`GET /health` against the host and port *this invocation* was going to use, so
`MERIDIAN_HOST` / `MERIDIAN_PORT` are honoured and a second config pointed at
3457 probes 3457.

`/health` is the right question because it is the only identifying route that is
public - `createProxyServer` gates `/v1/*`, `/telemetry*`, `/profiles*`,
`/plugins*`, `/settings*`, `/metrics` and `/auth/*` behind `requireAuth` and
deliberately leaves `/` and `/health` open - and because it is free: it reads
cached auth status, spends no model quota, and echoes no credential.

**Identification is by shape, not status code.** `/health` answers 503 when the
account is logged out and 401 when a key is required, and both are still
Meridian answering. A JSON body carrying a known `status` and a `version`, or
Meridian's own `authentication_error` envelope, is the instance identifying
itself - which matters most precisely when it is unhappy, because that is when
somebody runs the command to find out why.

**Three outcomes, kept distinct:**

| what answers the probe | result |
|---|---|
| Meridian | print the dashboard, **exit 0** |
| something else | an error naming what it found - status, `Server` header, content type, never the body - **exit 1** |
| nothing, within 2s | the same error, worded for a port held by something silent - **exit 1** |

A wedged listener cannot hang the CLI.

**The check runs before the server starts**, rather than from the `EADDRINUSE`
handler, so the dashboard is the whole output - by the time a bind fails, the
pre-flight auth check and the plugin loader have already printed.
`startProxyServer`'s own `EADDRINUSE` handler stays where it is and still covers
losing the race.

**Reachable on purpose too:** `meridian status` prints the same thing without
having to make a mistake first, and exits 1 with a fitting message when there is
nothing of ours on that port.

## The renderer

It is the same page, not a summary of it: same values, same conditions, and the
same colour meanings as `src/telemetry/landing.ts`, with the four payloads it
fetches as its only input.

- **Pure.** Colour is a caller-supplied flag, so piped output carries no escape
  bytes and the tests can assert that.
- The theme's four semantic tokens (`--ok`, `--warn`, `--err`, `--dim`) map to
  xterm-256 in one table rather than through the renderer.
- A field the instance does not return prints as an explicit unknown, **never as
  `0`**. A section that cannot be read says why instead of rendering empty.
- Not printing credentials falls out of the same rule: the page renders neither
  `email` for a configured profile nor anything from `apiKey`, so neither is
  read.

## Sample output

Against a live instance with ten profiles (values redacted where they identify
an account):

```
MERIDIAN  ● Operational  ·  corp3 claude-max

  ANTHROPIC_BASE_URL=http://127.0.0.1:3456
  … · passthrough · port 3456

ACCOUNTS

  ● corp3  ACTIVE                         $24.98
    30 requests · est. API value · 24h
    5h        ████████████░░░░░░░░░░░░░░░░   42%  in 3h 19m
    7d        █████████████░░░░░░░░░░░░░░░   46%  in 1d 14h
    7d Fable  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░    0%
    pace      █████████████░░░░░░░░░┃░░░░░  -31%  ~60% by reset

  ● randall-corp                           $0.00
    no traffic · 24h
    no usage data yet

LAST 24 HOURS

  Requests           1000  55 errors
  Tokens Out       905.7k  8.4k in
  Cache Hit           82%  prompt cache
  Est. API Value     $492  list prices
```

## Verification

Run against a real, live Meridian rather than only unit-tested:

| case | result |
|---|---|
| occupied port that IS Meridian (3456) | dashboard rendered, **exit 0** |
| occupied port that is NOT Meridian (a plain HTTP listener) | `Error: 127.0.0.1:34571 is already in use. Something is listening there, and it is not Meridian (HTTP 200 OK, text/plain).` **exit 1** |
| free port (34572) | proxy starts normally, unchanged |
| piped to a file | **0** escape-sequence lines |
| `bun test src/__tests__/cli-dashboard.test.ts` | 24 pass, 0 fail |
| `bun test src/__tests__/status-probe.test.ts` | 26 pass, 0 fail |
| `bunx tsc --noEmit` | exit 0, 0 error lines |

## Files

```
 ARCHITECTURE.md                     |   2 +
 bin/cli.ts                          |  41 +++
 docs/configuration.md               |   3 +-
 src/__tests__/cli-dashboard.test.ts | 275 ++++++++++
 src/__tests__/status-probe.test.ts  | 271 ++++++++++
 src/proxy/statusProbe.ts            | 257 +++++++++++
 src/telemetry/cliDashboard.ts       | 606 ++++++++++++++++++++++++++++
 7 files changed, 1454 insertions(+), 1 deletion(-)
```

---

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -55,6 +55,7 @@ src/
 │   ├── sessionStore.ts        ← Shared file store (cross-proxy session resume)
 │   ├── profiles.ts            ← Multi-profile support: resolve, list, switch auth contexts (leaf)
 │   ├── profileCli.ts          ← CLI commands for profile management (leaf, I/O)
+│   ├── statusProbe.ts         ← Asks a busy port whether it is Meridian, and collects what / shows
 │   ├── agentDefs.ts           ← Subagent definition extraction from tool descriptions
 │   ├── agentMatch.ts          ← Fuzzy agent name matching
 │   ├── design.ts              ← Claude Design MCP proxy (token store/refresh, auth precedence, login flow)
@@ -74,6 +75,7 @@ src/
 │   ├── pricingStore.ts        ← User pricing overrides (persisted JSON)
 │   ├── profileBar.ts          ← Shared profile switcher bar (injected into HTML pages)
 │   ├── profilePage.ts         ← Profile management page HTML
+│   ├── cliDashboard.ts        ← The landing page rendered for a terminal (pure)
 │   └── types.ts               ← Telemetry types
 └── plugin/
     └── claude-max-headers.ts  ← OpenCode plugin for session header injection
--- a/bin/cli.ts
+++ b/bin/cli.ts
@@ -25,6 +25,7 @@ Usage: meridian [command] [options]
 
 Commands:
   (default)        Start the proxy server
+  status           Show what a running instance is doing (the / page, in the terminal)
   setup            Configure the OpenCode plugin (run once after install)
   profile          Manage Claude account profiles (add, list, switch, remove)
   refresh-token    Refresh the Claude Code OAuth token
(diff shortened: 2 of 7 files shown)


rynfar:
Incorporated into main in #1065 (commit a80a15e227b5f292b177c1ef31d6af3d876009ac), retaining original commit authorship. Thank you @Nowaker!

PR #774, opened 2026-08-08
fix(config): make MERIDIAN_CONFIG_DIR relocate the directory, not one file in it

Nowaker:

I am still working on this branch. Opening it early so the diff is visible and
so a maintainer can say "no thank you" before more effort goes into it. I will
remove this banner when it is ready for review.

---

## The problem

`MERIDIAN_CONFIG_DIR` is named for a directory and moves one file in it.

`settingsFile()` (`src/proxy/settings.ts`) honours it. Nothing else does, and
the most consequential omission is `profiles.json`, which is a **module-level
constant** in `src/proxy/profiles.ts`:

```ts
const CONFIG_FILE = join(homedir(), ".config", "meridian", "profiles.json")
```

Computed at import, so it is fixed before anything can set the variable, and
unsettable by a test afterwards.

The consequence is that a second Meridian pointed at a fresh directory **is not
a second instance**. It still reads the production profile list, the per-profile
`claudeConfigDir` paths, the adapter instances and the feature toggles from
`~/.config/meridian`. A development build running beside a packaged one holds
the production account list and is one `profile switch` away from moving that
traffic.

That is how I found it: I wanted a dev instance to work on the profile UI, and
discovered there is no supported way to give one a different profile set.

## What this changes

Resolution moves into one module, `src/proxy/configDir.ts`, read from
`process.env` **per call**. `settings.ts` uses it too, so there is a single
definition of where the directory is rather than two that can drift.

Now relocated: `profiles.json`, `profiles/<id>`, `adapter-instances.json`,
`sdk-features.json`, `model-pricing.json`, `telemetry.db` — and `settings.json`,
whose behaviour is unchanged. Only the **defaults** move;
`MERIDIAN_PRICING_CONFIG` and `MERIDIAN_TELEMETRY_DB` still win for their own
file.

**Each 5s disk cache is keyed by the resolved path.** Without that, a reader
that changes the variable gets the previous directory's list served from a warm
cache — which is the isolation failure this change exists to prevent, passing
for the wrong reason.

## Deliberately left alone

- `~/.cache/meridian` — a cache, with its own `MERIDIAN_SESSION_DIR`.
- `~/.claude` and any explicit `claudeConfigDir` — Claude Code's directories,
  not Meridian's.
- `~/.config/opencode` — another tool's config that `meridian setup` writes.
- `plugins/`, `plugins.json` and `design-token.json` **do** belong to this
  directory and should follow it, but they resolve in `server.ts` and
  `design.ts`. Keeping the diff to the modules that decide where the directory
  *is* seemed the better trade for review; `MERIDIAN_PLUGIN_DIR`,
  `MERIDIAN_PLUGIN_CONFIG` and `MERIDIAN_DESIGN_TOKEN_PATH` relocate them
  individually today, and `docs/configuration.md` now names the gap rather than
  implying it is not one. Happy to fold them in if you would rather have it in
  one piece.

`XDG_CONFIG_HOME` is still not honoured, for the reason `settings.ts` already
gives: it would relocate the configuration of everyone who has it set.

## The migration hazard, warned about rather than papered over

Anyone already setting this variable **kept their profiles precisely because
only `settings.json` moved**. After this change their instance starts with none,
which reads as profiles lost rather than left where they always were.

So: on the first disk read that finds no `profiles.json` at the configured
location while the default location has one, Meridian says so once and names
both paths. It stays silent when the variable is unset (both paths are then the
same file) and under `MERIDIAN_PROFILES`, where disk discovery is off and an
empty list is expected. The decision is a pure function, so all four outcomes
are unit tested without touching either location.

## What this does not fix

Two instances sharing **one** directory are still not safe: profiles resolving
to the same `claudeConfigDir` share credential files, and either instance will
refresh those tokens. That is #772's subject. What this fixes is narrower and
was the blocker — a second instance can now have its own profile list at all.

## Verification

- `bun run test` — **2510 pass, 0 fail**, using the repo's own script (which
  runs several files in separate processes).
- `bunx tsc --noEmit` — exit 0, zero errors.
- New unit tests cover the resolver, the per-path cache keying, and all four
  outcomes of the migration notice.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -29,6 +29,7 @@ Environment variables, endpoints, authentication, SDK feature toggles, passthrou
 | `MERIDIAN_PROFILE_ORDER` | — | *(config order)* | Priority-mode pool order, comma-separated, highest priority first (e.g. `work,personal`). Also editable at `/settings`. |
 | `MERIDIAN_PASSTHROUGH_EARLY_STOP` | — | `1` | Set to `0` to disable [digest-turn elimination](#how-tool-calling-works-in-passthrough) and restore the old end-of-turn behavior |
 | `MERIDIAN_SUPPRESS_SCRATCHPAD` | — | `1` | Set to `0` to let the SDK advertise its proxy-host scratchpad directory in passthrough mode |
+| `MERIDIAN_CONFIG_DIR` | — | `~/.config/meridian` | Meridian's own config directory. Moving it moves everything inside it — see [below](#relocating-the-config-directory). |
 | `MERIDIAN_PRICING_CONFIG` | `CLAUDE_PROXY_PRICING_CONFIG` | `~/.config/meridian/model-pricing.json` | Path to the model pricing overrides file used by cost estimation |
 | `MERIDIAN_PROFILES` | — | unset | JSON array of profile configs (overrides disk discovery). See [Multi-Profile Support](profiles.md). |
 | `MERIDIAN_DEFER_TOOL_THRESHOLD` | — | `15` | Number of tools before non-core tools are deferred via ToolSearch. Set to `0` to disable. |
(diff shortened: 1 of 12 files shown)


rynfar:
Incorporated into main in #1066 (commit c07cefd4226ec3cbd1ccde5dd3176978c10b3ebd), retaining original commit authorship. Thank you @Nowaker!

PR #775, opened 2026-08-08
feat(profiles): reorder the profile pool by drag or keyboard, on both pages

Nowaker:

Adds drag-and-drop and keyboard reordering of the profile pool to `/profiles` **and** the home page, with one shared implementation and one shared order.

## Why

The order is load-bearing. `profileOrder` is what `resolvePriorityOrder()` consumes, so with `routing = "priority"` it decides which account serves a request and which one takes over when that account runs out.

Until now the only way to change it was the ↑/↓ list on `/settings`, which lives inside `#routing-order-wrap` and is therefore `display:none` **unless the mode is already `priority`** — so the order could not be arranged *before* switching to the mode that uses it. Meanwhile `/profiles`, the page named after these objects, rendered them in config order and offered no control at all.

## What it does

- Every card gets a real `<button>` drag handle, so it is in the tab order by default rather than by a `tabindex` somebody has to maintain.
- Dragging reorders and saves.
- **↑/↓ on a focused handle** moves that card one position and saves — the whole feature without a pointer, and the only path that works on touch, where HTML5 drag events do not fire at all.
- **Dragging near the top or bottom edge of the viewport scrolls the page** and keeps scrolling while the pointer stays there.

**No new endpoint.** `PUT /settings/api/routing` already accepts `profileOrder`, validates every id against the configured set, and persists through `setSetting`. This is a second client for it. The `/settings` list keeps working unchanged and every surface reads back the same saved order.

## The three defects fixed by the second commit

Reported against the first commit, from a live twelve-account fleet.

### 1. A drag could not reach past the current viewport

HTML5 drag-and-drop does not auto-scroll the page. Once the list is taller than the window — which twelve accounts comfortably are — the only reachable drop targets are the ones already on screen, so an item could be moved one or two positions and no further.

`edgeScrollVelocity(clientY, viewportHeight)` returns a per-frame scroll delta, ramped by how deep into the edge zone the pointer is, so one gesture serves both a nudge and a fast travel. It is a pure function with unit tests. Three details that are not guesses:

- A pointer dragged *past* the window edge reports a `clientY` outside the viewport, which **clamps** to full speed rather than accelerating without bound.
- The zone is capped at **half the viewport**, so on a short window the two zones cannot overlap and fight over the direction.
- Scrolling stops if no `dragover` has arrived for 700ms. `dragend` covers the ordinary paths, but a drag that leaves the window stops producing events while the last velocity is still in force, which would otherwise scroll to the end on its own.

The listener is on `document`, not `#content`: the pointer is over the header or the gap between cards as often as it is over a card.

### 2. The two pages disagreed about the order

`/profiles` sorted by the persisted `profileOrder`. The home page rendered `/profiles/list` in raw config order and never fetched the routing config at all — so a reorder took effect on one page and was invisible on the other. This was a read-path bug on `/`, not a write-path bug.

The persisted order is now the base order on every page: `/` fetches `/settings/api/routing` alongside its other calls and applies the same `sortByOrder`.

### 3. The home page had no drag at all

It has the same handles, the same gesture, the same auto-scroll and the same persistence, because it is the same code.

One home-page-specific guard: the card there is itself a click-to-switch button, so without a handle check every *grab* of a handle would also change the active account.

## Structure — one implementation, not two that agree today

The first commit followed the house pattern of transcribing helpers into the page's inline script, as `profileUsage.ts` does. That is exactly how defect 2 happened, and a second page would have doubled it.

`src/telemetry/profileOrder.ts` now owns the whole display half and exports the browser code as `reorderCss` / `reorderClientJs` / `reorderLiveRegionHtml`, which both pages interpolate. There is one copy of the gesture in the codebase; `profilePage.ts` goes from 644 lines to 463. A test asserts both pages interpolate the module rather than carrying their own transcriptions.

The pure parts (`moveInOrder`, `sortByOrder`, `edgeScrollVelocity`) stay directly unit-tested. `sortByOrder` appends unlisted profiles at the end in their original order, matching `resolvePriorityOrder()`, so a profile added after the order was saved renders where the router would put it.

## Two things that are still not obvious from the diff

- **The 10s poll is suspended while a drag is in flight**, and the drop handler reads the order back off the DOM rather than from the cached routing config. A poll landing mid-drag would replace the very cards being dragged; and the cached config can name a profile `/profiles/list` no longer returns, which the endpoint rejects as an unknown id.
- **When `MERIDIAN_PROFILE_ORDER` is set the handles are not rendered** and the list says so, because the env var wins over the setting. Offering a control that silently does nothing is worse than not offering it.

## Decision: dragging while a temporary view sort is active

This branch's home page has no sort control, so the case cannot arise in this diff. It arises the moment this meets #777, which adds a temporary "most spent / least spent" view sort to the same page.

**The rule is that dragging is offered only while the persisted order is the one on screen.** The drag persists whatever sequence the cards are currently in, so a drag under a spent-ranked view would silently overwrite the routing pool order with the spent ranking — an ordering the user never chose and cannot see they have chosen.

Mechanically the gate is one condition on the `reorderable` line in `landing.ts`: `multi && !meridianReorder.envPinned() && viewSort === 'configured'`, with the existing `noteHtml` explaining why the handles are absent, exactly as it already does for the env-var case. Whichever of the two PRs merges second owns that line.

## Accessibility

The `aria-label` names the profile and reads "position N of M", so the order is available without seeing the list; each move is announced through an `aria-live="polite"` region, because the visual result of a move is a reordered list a screen-reader user has no reason to re-read; and focus is restored to the same profile's handle after the re-render, so a run of ↑ presses moves one card repeatedly instead of moving one card and then whatever landed under the cursor.

Keyboard reorder is unaffected by defect 1 — it never needed the viewport — so the feature remained usable on a long list throughout.

## Verification

- `bunx tsc --noEmit` exit 0.
- `bun run test` (the project's own chain) — **2515 pass, 0 fail**, pipeline exit 0.
- Driven in a real browser (headless Chromium over CDP) against a development instance on its own port and its own config dir, with an eleven-account list and a 700×600 viewport so both pages overflowed:

| check | result |
|---|---|
| `/` renders the persisted order | matches `profileOrder` exactly, and differs from the raw `/profiles/list` order it used to render |
| edge auto-scroll, `/` (1708px page) | bottom → top edge scrolled 1108 → 633 in 400ms; top → bottom edge scrolled 0 → 475 |
| edge auto-scroll, `/profiles` (4987px page) | 4387 → 3912 in 400ms |
| scrolling stops | `dragend` halts it with no drift; a `dragover` in the middle of the viewport holds position |
| drag on `/` | moved a card from position 11 to 1, announced it, persisted to `settings.json` on disk |
| grabbing a handle on `/` | active profile unchanged |
| keyboard move on `/profiles` | moved, announced, focus kept on the moved card |
| round trip | a reorder on `/` showed up on `/profiles` and vice versa |
| console | no errors |

### DIFF
--- a/src/__tests__/profile-order.test.ts
+++ b/src/__tests__/profile-order.test.ts
@@ -1,0 +1,82 @@
+/**
+ * Unit tests for the profile-reordering helpers — pure functions, no mocks.
+ */
+import { describe, expect, it } from "bun:test"
+import { moveInOrder, sortByOrder } from "../telemetry/profileOrder"
+
+describe("moveInOrder", () => {
+  const base = ["a", "b", "c", "d"]
+
+  it("moves an item down", () => {
+    expect(moveInOrder(base, 0, 2)).toEqual(["b", "c", "a", "d"])
+  })
+
+  it("moves an item up", () => {
+    expect(moveInOrder(base, 3, 1)).toEqual(["a", "d", "b", "c"])
+  })
+
+  it("moves to the ends", () => {
+    expect(moveInOrder(base, 2, 0)).toEqual(["c", "a", "b", "d"])
+    expect(moveInOrder(base, 0, 3)).toEqual(["b", "c", "d", "a"])
+  })
+
+  it("returns an unchanged copy for a no-op move", () => {
+    const out = moveInOrder(base, 1, 1)
+    expect(out).toEqual(base)
+    expect(out).not.toBe(base)
+  })
+
+  it("ignores out-of-range indices instead of corrupting the order", () => {
+    expect(moveInOrder(base, -1, 2)).toEqual(base)
+    expect(moveInOrder(base, 0, 9)).toEqual(base)
+    expect(moveInOrder(base, 9, 0)).toEqual(base)
+    expect(moveInOrder(base, 0, -1)).toEqual(base)
+  })
+
+  it("never mutates the input", () => {
+    const input = ["a", "b", "c"]
+    moveInOrder(input, 0, 2)
+    expect(input).toEqual(["a", "b", "c"])
+  })
+
+  it("handles empty and single-item lists", () => {
+    expect(moveInOrder([], 0, 0)).toEqual([])
+    expect(moveInOrder(["only"], 0, 0)).toEqual(["only"])
+  })
+})
+
+describe("sortByOrder", () => {
+  const profiles = [{ id: "alpha" }, { id: "beta" }, { id: "gamma" }]
+
+  it("applies the saved order", () => {
+    expect(sortByOrder(profiles, ["gamma", "alpha", "beta"]).map((p) => p.id))
+      .toEqual(["gamma", "alpha", "beta"])
+  })
+
+  it("keeps config order when nothing is saved", () => {
+    expect(sortByOrder(profiles, []).map((p) => p.id)).toEqual(["alpha", "beta", "gamma"])
+  })
+
+  it("appends profiles missing from the saved order, in their original order", () => {
+    // "beta" was added after the order was last saved.
+    expect(sortByOrder(profiles, ["gamma", "alpha"]).map((p) => p.id))
+      .toEqual(["gamma", "alpha", "beta"])
+  })
+
+  it("ignores ids in the order that no longer exist", () => {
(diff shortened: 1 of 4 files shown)


rynfar:
Incorporated into main via #1078 with desktop app parity and full integration tests. Thank you @Nowaker!

PR #779, opened 2026-08-08
feat(settings): make the per-harness SDK Features sections tabs

Nowaker:

Nowaker is still iterating on this. Filed early so the change is visible and
reviewable when it is ready; please hold off until the DRAFT marker is gone.

---

## What

Turns the per-harness **SDK Features** cards on `/settings` into a tab strip
over one body. Picking a tab shows that harness's settings and nothing else,
including its two system-prompt toggles - the built-in Claude Code prompt and
the connecting agent's own.

## Why

Ten adapters rendered as ten stacked cards is a long scroll whose only structure
is repetition, and it pushes everything below it off the first screen. That is
also what buried the Routing controls (sibling PR).

## Details worth a reviewer's attention

- **The strip is built from the same `Object.keys(currentConfig)` the flat list
  used**, so an adapter the API returns appears the moment it exists, labelled by
  `ADAPTER_LABELS` or by its raw name. No hardcoded list to keep in sync.
- **Tabs reuse `dashboard.ts`'s `.tab` rules verbatim**, so the two pages read
  alike rather than gaining a second tab style.
- **They wrap rather than scroll.** Ten labels do not fit 900px on one row, and a
  strip that hides harnesses off-screen would reintroduce the discoverability
  problem this is meant to fix.
- **A tab whose harness is off its defaults carries a small dot.** Stacked cards
  showed every harness's Active/Default badge at once; one body at a time cannot,
  and losing that signal silently would be a regression.

## Verification

- `bunx tsc --noEmit` exit 0, 0 errors
- `bun run test` (the project's own chain) - **2494 pass, 0 fail**, pipeline exit 0
- Rendered in a browser against a development instance on its own port and config
  dir: every harness tab selects and shows its own toggles, no console errors

## Scope

Presentation only - no adapter behaviour, no settings API change.

Cut from `main`, one commit, independent of the sibling Routing PR. They touch
the same file and will conflict if both are taken; the conflict is textual and
trivial, and either can be taken alone.

---

### DIFF
--- a/src/__tests__/sdk-features-unit.test.ts
+++ b/src/__tests__/sdk-features-unit.test.ts
@@ -207,4 +207,10 @@ describe("settings UI adapter coverage", () => {
     const { settingsPageHtml } = require("../telemetry/settingsPage") as typeof import("../telemetry/settingsPage")
     expect(settingsPageHtml).toContain("ADAPTER_LABELS[adapter] || adapter")
   })
+
+  it("renders the harnesses as tabs over a single selected body", () => {
+    const { settingsPageHtml } = require("../telemetry/settingsPage") as typeof import("../telemetry/settingsPage")
+    expect(settingsPageHtml).toContain('id="adapterTabs"')
+    expect(settingsPageHtml).toContain("selectedAdapter")
+  })
 })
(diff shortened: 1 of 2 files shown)


rynfar:
Consolidated and merged into main via #1080.

PR #781, opened 2026-08-09
fix(quota): back off OAuth usage rate limits

CrazyCoder:
## Summary

- track a per-profile cooldown after the OAuth usage endpoint returns HTTP 429
- honor `Retry-After` with a 60-second floor, including forced refreshes
- keep serving the last-good quota snapshot during the cooldown

## Problem

Forced quota refreshes bypassed the normal cache and retried Anthropic's private usage endpoint on every request. A 429 response could therefore keep the endpoint rate-limited until the bounded last-good snapshot expired, which cleared clients' quota displays.

## Verification

- `bun test src/__tests__/oauth-usage.test.ts src/__tests__/proxy-usage-quota-route.test.ts`
- `bun run typecheck`
- `bun run build`
- manual Pi 0.84.1 round trip through plugin-enabled Meridian using the `pi` adapter: model response completed and the live five-hour/seven-day quota footer remained populated

### DIFF
--- a/src/__tests__/oauth-usage.test.ts
+++ b/src/__tests__/oauth-usage.test.ts
@@ -171,6 +171,57 @@ describe("oauthUsage", () => {
     expect(result).toBeNull()
   })
 
+  test("backs off repeated 429 responses per profile even when forced", async () => {
+    const { fetchImpl, getCalls } = countingFetch(() =>
+      new Response("rate limited", { status: 429, headers: { "Retry-After": "120" } }))
+    const store = makeStore("t")
+
+    const first = await fetchOAuthUsage({ force: true, store, profileId: "limited", fetchImpl })
+    const second = await fetchOAuthUsage({ force: true, store, profileId: "limited", fetchImpl })
+    const otherProfile = await fetchOAuthUsage({ force: true, store, profileId: "other", fetchImpl })
+
+    expect(first).toBeNull()
+    expect(second).toBeNull()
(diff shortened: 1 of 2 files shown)


rynfar:
I will try to get this reviewed tonight

rynfar:
Thanks for this, @CrazyCoder — the diagnosis was right, and the fix shipped in #785.

To explain the close, since GitHub gives you no signal about it: `main` requires signed commits, so PRs from forks can't be merged directly no matter how green the checks are. It surfaces only as `mergeStateStatus: BLOCKED` with no reason attached, and there was nothing you could have done from your side. The way through is for a maintainer to cherry-pick, which re-signs the commit while preserving you as its author — so your commit is on `main` unchanged, and the squash carries a `Co-authored-by:` trailer for you.

One change on top, as a separate commit: the cooldown had a floor but no ceiling, so a `Retry-After` longer than the 15-minute `STALE_MAX_MS_DEFAULT` window could outlive the last-good snapshot it was meant to protect. `Retry-After: 3600` gave 45 minutes of `null` usage with zero retries — the inverse of serving stale data through the cooldown. It's now capped at `staleMaxMs`, floored so a small `staleMaxMs` can't disarm the throttle.

A follow-up in #786 also fixes a related mislabel your change made visible: entering the cooldown with an empty cache reported `error: "no_token"`, which the UI renders as "Run `claude login`" — misleading when the credentials are fine.

Apologies that the close landed without context.

PR #783, opened 2026-08-09
fix(usage): report why a usage fetch failed instead of calling every failure no_token

Nowaker:

`/v1/usage/quota/all` documents `error: "no_token" | "upstream_error"` and only
ever emits the first, because `fetchOAuthUsage` collapses every failure into
`null` and the route has nothing else to go on. A rate-limited read therefore
reaches the UI as an account that has lost its credentials.

## How it was found

Measured on a ten-account box running two Meridian instances against the same
credential files. Anthropic's usage endpoint answered **429 for all ten
accounts** — one per-account budget, two readers — and the dashboard put a
**"needs login" pill on every one of them, including the account it was
serving from at that moment**.

That is the part worth stating plainly: `claude login` was the advertised
remedy for a problem that was neither the account's nor the operator's, and
following the advice would have re-authenticated ten healthy accounts. Two of
them are borrowed, so it would also have meant asking their owners.

The accounts recovered on their own once the rate limit passed. Nothing was
ever wrong with any token.

## The change

`fetchOAuthUsageResult` returns the reason beside the snapshot, and the two
values are exactly the two available responses:

| reason | means |
|---|---|
| `no_token` | needs a human |
| `upstream_error` | needs waiting |

**A 401 that could not be refreshed is `upstream_error`, not `no_token`.** The
token is present, and an instance may decline to refresh it deliberately —
`MERIDIAN_CREDENTIALS_READONLY` is exactly that case — so a missing login is
the one thing it is not.

A snapshot served stale after a transient failure still reports no error,
because the caller did get numbers.

`fetchOAuthUsage` stays as the snapshot-only form for callers that act on the
figures rather than reporting a status, so the exhaustion refinement and the
single-profile route are unchanged.

## Scope

Three files, +120/-19, cut clean off `main`. The new behaviour is covered in
`src/__tests__/oauth-usage.test.ts`; `bun run test` is green (2692 pass, 0
fail) and `tsc --noEmit` exits 0 on the integration branch carrying it.

### DIFF
--- a/src/__tests__/oauth-usage.test.ts
+++ b/src/__tests__/oauth-usage.test.ts
@@ -9,7 +9,7 @@
  */
 
 import { describe, expect, test, beforeEach } from "bun:test"
-import { fetchOAuthUsage, resetOAuthUsageCache } from "../proxy/oauthUsage"
+import { fetchOAuthUsage, fetchOAuthUsageResult, resetOAuthUsageCache } from "../proxy/oauthUsage"
 import type { CredentialStore } from "../proxy/tokenRefresh"
 
 const SAMPLE_RESPONSE = {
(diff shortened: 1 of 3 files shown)


rynfar:
Thanks @Nowaker — this shipped in #789, with your commit and authorship intact.

To explain the close, since GitHub gives you no signal about it: `main` requires signed commits, so PRs from forks can't be merged directly no matter how green the checks are. It surfaces only as `mergeStateStatus: BLOCKED` with no reason attached, and there was nothing you could have done from your side. A maintainer cherry-picks instead, which re-signs the commit while preserving you as its author — the squash carries a `Co-authored-by:` trailer for you.

Apologies for the conflict you'd have hit here: #785 and #786 landed in the meantime and rewrote the same functions. #786 was mine and addressed the same bug from a worse angle — a module-state lookup *after* a null return, which could only ever distinguish a 429. Your design supersedes it and `explainMissingOAuthUsage` is gone. The 5xx and refresh-refusal cases only get fixed the way you did it, by threading the reason out of the fetch; the refresh-refusal one especially, since a read-only instance declines by design and telling that operator to re-authenticate every account is precisely backwards.

Two changes on top, both noted in #789:

- Your union was `no_token | upstream_error`, with a 429 as `upstream_error`. Because #785's backoff can hold that state for minutes at a stretch, it's split into its own `rate_limited` value — so your test asserting `upstream_error` for a 429 now asserts `rate_limited`. Narrowed, still never `no_token`.
- The `__setFetchOAuthUsageOverride` shim returned `no_token` for any null, which would have let it contradict a cooldown the same process had just recorded. It now consults the cooldown.

Nowaker:
@rynfar Thanks - no worries. I'll force-push the ones still pending. I got sidetracked with other work, but will be wrapping up all my contributions soon.

PR #787, opened 2026-08-11
fix: recognize weekly Claude limit wording

WarGloom:
## Summary

- classify the live Claude Code wording `You've hit your weekly limit` as `429 rate_limit_error`
- let priority routing retry the next configured profile instead of returning a generic 500
- add classifier and HTTP-level priority failover regression coverage

This complements #764, whose shorter `You've hit your limit` pattern does not match the observed weekly wording.

## Verification

- red phase: both new regressions failed as `api_error` / HTTP 500
- `bun test src/__tests__/errors.test.ts src/__tests__/priority-routing-integration.test.ts` - 101 passed
- `npm run build` - passed, including declaration generation
- full suite comparison: the patch run and untouched `origin/main` reproduce the same five unrelated SDK-feature/system-prompt failures; no changed-path test fails

Observed production error:

```text
Claude Code returned an error result: You've hit your weekly limit · resets 2pm (Asia/Jerusalem)
```

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -490,6 +490,12 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     expect(r.status).toBe(429)
   })
 
+  it("maps the CLI's 'You've hit your weekly limit' to rate_limit_error", () => {
+    const r = classifyError("Claude Code returned an error result: You've hit your weekly limit \u00b7 resets 2pm (Asia/Jerusalem)")
+    expect(r.type).toBe("rate_limit_error")
+    expect(r.status).toBe(429)
+  })
+
   it("maps 'usage limit reached' to rate_limit_error", () => {
     const r = classifyError("usage limit reached | resets at 5pm")
     expect(r.type).toBe("rate_limit_error")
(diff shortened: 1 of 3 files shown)


rynfar:
Thanks @WarGloom — this shipped in #788.

To explain the close, since GitHub gives you no signal about it: `main` requires signed commits, so PRs from forks can't be merged directly no matter how green the checks are. It surfaces only as `mergeStateStatus: BLOCKED` with no reason attached, and there was nothing you could have done from your side. The way through is for a maintainer to cherry-pick, which re-signs the commit while preserving you as its author — your commit is on `main` unchanged, and the squash carries a `Co-authored-by:` trailer for you.

Worth noting: @justprosh opened #764 in the same window for the shorter `"You've hit your limit"`. Neither wording subsumes the other — `"hit your limit"` isn't a substring of `"hit your weekly limit"` — so both were real, and both commits are in #788, each with its own regression test.

Two variants in one window suggested the literal list would keep growing, so there's a commit on top replacing it with `/hit your (?:[\w-]+ )?limit/`, covering daily, monthly, and 5-hour too.

One adjustment to the test: the `failureMessage` reset moved from the `priority routing` describe's `beforeEach` to the file-level one. Since `failureMessage` is module-level and two later describes never reset it, the describe-scoped version was benign only because that block's tests happen to run last — adding or reordering a test would leak the weekly wording into `keyless priority affinity`, whose timing assumptions depend on the retry ladder that wording doesn't trigger.

PR #794, opened 2026-08-11
feat(profiles): create a profile from the web UI

Nowaker:

**Stacked on #792.** This branch is cut from `feat/profile-login-from-web-ui`, not
from `main`, because it reuses the PKCE login machinery that PR introduces. GitHub
cannot express that across a fork, so the diff below carries **two** commits — take
`be4c279` as the change under review here and #792 as its prerequisite.

Adding an account currently needs a shell on the machine Meridian runs on. This
finishes what #792 started: the Profiles page grows an "Add a profile" box — name
it, sign in to Claude in the tab that opens, paste back what Claude shows (the bare
code or the whole callback URL). The profile appears with its own config directory,
ready to use with no restart.

## It is its own route, not a loosened guard

`/profiles/login/start` refuses unknown ids **deliberately**, and that refusal is
what stops "re-authenticate `enrique-corp`" from silently creating `enrique-crop`
on a typo. Creating an account slot is a different act from re-authenticating one,
so it gets its own button, its own routes and its own refusals.

## One definition of what a profile is made of

`profileAdd` welded id validation, a collision check, `mkdir` and the
`profiles.json` entry into the middle of an interactive CLI flow. The creation half
is now `createProfileSlot`, called by the CLI at each of its three success paths and
by the new route. `isValidProfileId` is lifted out of a regex that had been written
inline twice — the id becomes a directory name, so having one answer to "what is a
legal id" is the whole point.

## Nothing is written until the credentials are in hand

The exchange with Anthropic happens first; the `profiles.json` entry only once it
succeeds. A sign-in that is abandoned, rejected or never finished leaves **no
profile behind at all** — no half-made slot stuck at "not logged in" for someone to
find and clean up. The residue of a failure is at most an empty directory the next
attempt reuses.

Every refusal happens at `/start`, before a sign-in tab opens: an illegal name, a
name that already exists (pointing at "Log in from browser", which is what that
button is for), and `MERIDIAN_CREDENTIALS_READONLY`. Refusing *after* sign-in would
burn a one-time code for nothing.

## Superseding, not refusing, a second sign-in for the same name

Refusing was the first implementation, and driving it in a browser proved it wrong:
cancelling the panel, reloading the page and closing the tab all abandon a sign-in
without telling the server, so the name was locked out for the full ten minutes by a
message telling the user to cancel something they had no way to cancel. One entry per
name still holds — that is what stops two sign-ins for one new name from both
completing; superseding only decides which survives.

## What actually guards this, stated rather than implied

Both routes sit under `/profiles/*` and inherit `requireAuth`, so they are behind
`MERIDIAN_API_KEY` when one is set. **When it is not — the default — the only thing
between the page and a new profile is whatever can reach the port.** Measured on the
author's box: `GET /profiles` answers `200` with no auth header, on loopback and
through a reverse proxy alike. Adding a profile is the most privileged thing that
page does, so it is kept to exactly that: no delete, no rename, no edit. The docs say
this rather than implying the routes are protected on their own.

Nothing from the code or the token response is logged; the refusal log records a
reason code and nothing else. `loadProfileIds` returns ids rather than profiles so a
collision check can never reach an `apiKey`.

## The `~/.claude` import stays CLI-only

`meridian profile add` on a host whose default config dir is already signed in offers
to adopt those credentials. The UI always signs in fresh: silently claiming whichever
account you happen to be logged in as on that machine is not something a button press
should be able to do. The docs say so.

## Tests

43 across two new files — id validation including traversal-shaped names, collisions
against both the effective profile list and `profiles.json`, readonly refusal,
supersede, expiry, single use, both paste forms, API-key enforcement on both routes,
and the slot **not** being written when the exchange fails. Also driven in a real
browser against two throwaway instances, one with `MERIDIAN_CREDENTIALS_READONLY`,
which is where the lockout above was found.

`bun run test` green and `bun run typecheck` clean on the branch.

---

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -102,6 +102,8 @@ adapter lets the subprocess run the built-in WebFetch at all.
 | `POST /profiles/active` | Switch the active profile |
 | `POST /profiles/login/start` | Begin an OAuth login for a profile — returns the authorize URL (see [Re-authenticating from the web UI](profiles.md#from-the-web-ui-re-authenticate-a-profile-without-a-terminal)) |
 | `POST /profiles/login/complete` | Finish that login with the code the user pasted back |
+| `POST /profiles/add/start` | Begin creating a new profile — validates the name and returns the authorize URL (see [Adding from the web UI](profiles.md#from-the-web-ui-add-a-profile)) |
+| `POST /profiles/add/complete` | Finish that creation — writes the profile only once Anthropic returns credentials |
 | `GET /v1/usage/quota` | Usage windows for the active profile (JSON) |
 | `GET /v1/usage/quota/all` | Usage windows for every profile (JSON) |
 | `GET /settings` | SDK feature toggles + model pricing UI |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -100,6 +100,8 @@ adapter lets the subprocess run the built-in WebFetch at all.
 | `GET /profiles` | Profile management page |
 | `GET /profiles/list` | List profiles with auth status (JSON) |
 | `POST /profiles/active` | Switch the active profile |
+| `POST /profiles/login/start` | Begin an OAuth login for a profile — returns the authorize URL (see [Re-authenticating from the web UI](profiles.md#from-the-web-ui-re-authenticate-a-profile-without-a-terminal)) |
+| `POST /profiles/login/complete` | Finish that login with the code the user pasted back |
 | `GET /v1/usage/quota` | Usage windows for the active profile (JSON) |
 | `GET /v1/usage/quota/all` | Usage windows for every profile (JSON) |
 | `GET /settings` | SDK feature toggles + model pricing UI |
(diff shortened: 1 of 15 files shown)


Nowaker:
Combined into #792, which now carries both halves of the same journey — the page that adds an account is the page that re-authenticates one.

`be4c279` was cherry-picked onto that branch and its three conflicts (`docs/configuration.md`, `src/proxy/server.ts`, `src/telemetry/profilePage.ts`) were resolved as unions, so nothing here is lost: `src/proxy/profileAdd.ts`, `src/proxy/settings.ts` and both add-profile test files arrived byte-identical, and the shared `renderOauthPanel` from this branch now backs both paste panels.

Consolidation went that direction rather than this one because this branch's first commit, `fff5043`, is an older copy of the login work — #792's version is roughly twice the size and carries the redirect-back flow that replaced the code paste. Landing on top of `fff5043` would have regressed it.

Closing in favour of #792. The branch is left in place.

PR #795, opened 2026-08-11
fix(profiles): persist the account plan at headless OAuth login

Nowaker:

`meridian profile add --headless` writes four fields into `.credentials.json` — `accessToken`, `refreshToken`, `expiresAt`, `scopes` — and drops the account's plan, even though `OAuthCredentials` has declared `subscriptionType` and `rateLimitTier` all along.

Measured on a ten-account host: a credential file written by `claude login` is **471 B** and carries `subscriptionType`; one written by Meridian's own headless login is **405 B** and does not. Eight of the ten profiles on that host were created by Meridian and are therefore plan-blind.

## Why that matters downstream

Everything reads the plan back out of that file via `claude auth status`, so a headless-created profile stays plan-unknown **forever**:

- `meridian profile list` and `/profiles/list` report no subscription type
- `/health` omits it
- the `max`-only branch in `/v1/models` — the one advertising the larger context window — silently never fires

## The plan is not on the token response

This is the part worth knowing, because the obvious fix does not work. Claude Code reads only `access_token`, `refresh_token`, `expires_in`, `scope`, `account.{uuid,email_address}` and `organization.uuid` off a token response. It derives `subscriptionType` / `rateLimitTier` from a **separate** authenticated `GET https://api.anthropic.com/api/oauth/profile`, mapping the wire `organization_type` (`claude_max`) onto the stored value (`max`).

The headless login now makes the same call, using the `user:profile` scope it already requests. The mapping was verified against the shipped `claude` binary rather than inferred: `api/oauth/profile`, `organization_type`, `rate_limit_tier` and all four wire values (`claude_max` ×17, `claude_pro`, `claude_team`, `claude_enterprise`) are present in it.

## Properties

- **Best-effort; never fails the login.** Three separate catch paths and a 10s timeout. On any error the plan stays unknown and the profile is still created — a profile whose plan is unknown is strictly better than no profile, and every consumer already treats the field as optional.
- **Fields are spread in only when known.** `subscriptionType: undefined` would write a null-ish key into a file the real CLI also parses.
- **Nothing credential-shaped is logged.** The three warnings carry an error message or an HTTP status, never a token or a response body.
- **`buildLoginCredentials` is split out and pure**, so the on-disk shape is directly assertable. The defect being closed was a field missing from an object literal, which no test of the surrounding prompt/fetch I/O would have caught.
- **`doRefresh`'s spread over the existing credential is what keeps the value alive** across each ~8h refresh. A test pins that, so it is not later rewritten as an explicit field list — which would re-open the same hole one refresh after every login.

## Forward-looking only

A profile created by an earlier version **cannot be repaired by a token refresh**: the value is written at login and nowhere else, and Anthropic's usage endpoint does not carry it. Operators must re-run `meridian profile login <name> --headless` on affected profiles. `docs/profiles.md` says so.

## Tests

19 new assertions in `src/__tests__/profile-login-plan-fields.test.ts`: the wire→stored mapping including unrecognized and absent types, field omission when unknown, every failure path of the lookup returning `{}` rather than throwing, the credential record's exact on-disk shape, and the refresh spread preserving an existing plan.

`bun run test` green, `tsc --noEmit` exit 0.

---

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -31,6 +31,10 @@ Open the printed URL in a browser, sign in to the target Claude account, then pa
 meridian profile login work --headless
 ```
 
+The same login also records the account's plan (`subscriptionType`, `rateLimitTier`), read from Anthropic's OAuth profile endpoint — the token exchange itself returns no plan information. That is what lets `meridian profile list`, `/profiles/list`, `/health` and the dashboard tell a Max account from a Team one, and what makes `/v1/models` advertise the larger context window Max accounts actually have. If the lookup fails the login still succeeds and the plan simply stays unknown.
+
+> **⚠ A profile created by an older Meridian has no plan recorded, and a token refresh cannot backfill it** — the value is only ever written at login, and Anthropic's usage endpoint does not carry it. Re-run `meridian profile login <name> --headless` to repair such a profile.
+
 #### Headless / CI: register an OAuth token
 
 When a browser isn't available (containers, CI runners, remote shells), generate a long-lived OAuth token with `claude setup-token` and register it as a profile:
--- a/src/__tests__/profile-login-plan-fields.test.ts
+++ b/src/__tests__/profile-login-plan-fields.test.ts
@@ -1,0 +1,270 @@
+/**
+ * Unit tests for the plan fields a headless OAuth login persists.
+ *
+ * Anthropic's token endpoint returns no plan information, so the headless login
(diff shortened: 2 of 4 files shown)


rynfar:
Thank you @Nowaker! This has been incorporated and merged via #1070 with original commit authorship preserved, resolving merge conflicts with the recent credential store hardening and extending account plan badges to the desktop application.

PR #803, opened 2026-08-12
feat(profiles): say what plan an account is on, and how much usage it buys

Nowaker:

> Stacks on #795. The first commit here is that PR's; the two after it are this one. Read/merge #795 first.

`Plan: max` does not say how much work an account can do.

Personal Max **5x** and Personal Max **20x** both report `max` through `claude auth status` and differ by **4x** in Claude Code allotment. So a Profiles page showing only the plan cannot tell the reader which of two accounts has room left — and on the landing page it leaves every usage bar unqualified, because 70% of a 20x account is several times the work left in 70% of a 5x one.

## Where the distinguishing field lives

`rateLimitTier` (`default_claude_max_20x`) is the field that separates them, and it is **never** in `claude auth status` — that reports `subscriptionType` and nothing else. It exists only in the profile's own `.credentials.json`, written at login. So the derivation reads the credential store rather than the live auth status.

`readRefreshTokenExpiry` already read that file on the `/health` path and cached it per `refreshKey` for 5 minutes, because on macOS each read spawns `security find-generic-password`. Rather than add a second read with its own cache, that one is generalized to `readCredentialFacts` and now carries the plan fields alongside the refresh-token expiry — they arrive in one file and are wanted by the same request. `/health` and `/profiles/list` therefore gain the allowance at **no extra store hit**.

## A Team seat is sized by its seat, not by its rate-limit tier

`rate_limit_tier` cannot size a Team account, and the two ways it fails are opposite. Measured across twelve live accounts through `/api/oauth/profile`:

| seat | `organization.rate_limit_tier` | `organization.seat_tier` |
|---|---|---|
| Team Premium (x4) | `default_claude_max_5x` | `team_tier_1` |
| Team Standard | `default_raven` | `team_standard` |
| personal Max | `default_claude_max_5x` / `default_claude_max_20x` | `null` |

A Premium seat's tier is **byte-identical** to what a personal Max 5x reports, so reading it understates a 6.25x seat as 5x and names it after somebody else's product. A Standard seat's is a codename naming no published allotment, so reading it yields nothing at all.

`seat_tier` separates them, is absent on every personal account, and is therefore consulted **first**; `rate_limit_tier` still sizes personal plans, which is the only thing it is reliable for.

On the page this was two visible defects. A Team Premium seat rendered `Plan: Team` with `Allowance: 5x of a Pro plan's Claude Code usage`, understating it by 25%. A Team Standard seat rendered a bare lowercase `Plan: team` with **no Allowance row at all**, because `default_raven` matches no tier.

## The mapping

`planAllowance()` (`src/proxy/planAllowance.ts`) is a pure leaf module — no I/O, no imports — mapping a normalized tier to a multiplier, matching Anthropic's published allotments:

| tier | plan | allowance |
|---|---|---|
| `max_20x` | Max 20x | 20x |
| `max_5x` | Max 5x | 5x |
| `pro` | Pro | 1x |
| `team_tier_1` / `team_premium` | Premium seat | 6.25x |
| `team_standard` | Standard seat | 1x |

Normalization strips the `default_claude_` / `default_` prefix and underscores first, so one entry covers every spelling of a tier rather than one entry per spelling. `seat_tier` is looked up in the same table and also with a `team ` prefix, since that field is already scoped to a team — a bare `standard` there means what `team_standard` means in `rate_limit_tier`.

Bare `max` and bare `team` deliberately resolve to **nothing**. Each names a family with two differently-sized members, so answering would mean picking one at random — and rendering `1x` for an account that is really 20x is worse than rendering nothing, because a reader acts on it.

## Plan and account are separate answers

Two fields answer two questions and either can be known without the other: an account whose family is declared but whose seat is missing still has a trustworthy family. So `accountType` (Personal / Team / Enterprise) and `planName` (the name on Anthropic's own pricing page) are derived and rendered separately, and `planName` degrades to null — rather than asserting a product name nobody can confirm — when the declared family contradicts the tier's.

## A live status must not erase what was read off disk

`/health` and `/profiles/list` both spread the live `claude auth status` over the stored plan fields. `subscriptionType: undefined` **overwrites** rather than defers, so an account whose status omits the field lost its stored plan entirely — which is why a personal Max rendered no Plan row at all. The live value is now spread only when there is one.

## What clients see

Both routes emit `rateLimitTier`, `seatTier`, `allowance`, `allowanceWeight`, `planLabel`, `accountType` and `planName`. Every field is additive and nullable, so a client that predates them is unaffected. `allowanceWeight` is the same value as a number, so a caller can weight or rank accounts without re-parsing the label.

- `/profiles` renders `Account` and `Plan` rows beside `Allowance: 20x of a Pro plan's Claude Code usage`, with the raw seat or tier in the `title`.
- `/` renders a `20x` chip beside the profile name, where it qualifies the usage bars below it.

`seat_tier` is carried through `extractPlanFields`, the credential file, the facts cache and `getStoredPlanFields`, so it survives a token refresh the way the other two plan fields already did.

## Verification

`npm run typecheck` clean. The derivation's own unit tests are green — **39 pass, 0 fail** across `plan-allowance.test.ts` and `profile-login-plan-fields.test.ts`, covering every tier, both spellings of Team Premium, the prefix stripping, the deliberate refusal on bare `max`/`team`, the `subscriptionType`-only fallback, and the two measured Team seats above as real wire payloads.

The full suite's failure set is **identical with and without this change**, verified as a control rather than by comparing two runs: the change was stashed in the same worktree and the suite re-run at the same moment, giving the same failures. (Several integration tests in this repo read the developer's real `~/.config/meridian`, so their outcome moves with the live account list and two runs taken minutes apart are not comparable.)

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -31,6 +31,10 @@ Open the printed URL in a browser, sign in to the target Claude account, then pa
 meridian profile login work --headless
 ```
 
+The same login also records the account's plan (`subscriptionType`, `rateLimitTier`), read from Anthropic's OAuth profile endpoint — the token exchange itself returns no plan information. That is what lets `meridian profile list`, `/profiles/list`, `/health` and the dashboard tell a Max account from a Team one, and what makes `/v1/models` advertise the larger context window Max accounts actually have. If the lookup fails the login still succeeds and the plan simply stays unknown.
+
+> **⚠ A profile created by an older Meridian has no plan recorded, and a token refresh cannot backfill it** — the value is only ever written at login, and Anthropic's usage endpoint does not carry it. Re-run `meridian profile login <name> --headless` to repair such a profile.
+
 #### Headless / CI: register an OAuth token
 
 When a browser isn't available (containers, CI runners, remote shells), generate a long-lived OAuth token with `claude setup-token` and register it as a profile:
--- a/src/__tests__/plan-allowance.test.ts
+++ b/src/__tests__/plan-allowance.test.ts
@@ -1,0 +1,109 @@
+/**
+ * Unit tests for the plan → allotment multiplier derivation.
+ *
+ * The module is pure, so these are direct assertions with no mocks. The cases
+ * that matter are the ones where a wrong answer is worse than no answer: a
+ * `max` that could be 5x or 20x, and a tier string Anthropic has not shipped
+ * yet.
+ */
+
+import { describe, it, expect } from "bun:test"
+import { normalizeRateLimitTier, planAllowance } from "../proxy/planAllowance"
+
+describe("normalizeRateLimitTier", () => {
+  it("strips the wire prefix and the underscores", () => {
+    expect(normalizeRateLimitTier("default_claude_max_20x")).toBe("max 20x")
+    expect(normalizeRateLimitTier("default_team_tier_1")).toBe("team tier 1")
+  })
+
+  it("accepts the bare spelling of the same tier", () => {
+    expect(normalizeRateLimitTier("max_5x")).toBe("max 5x")
+    expect(normalizeRateLimitTier("MAX_5X")).toBe("max 5x")
+  })
+
+  it("treats absent and blank as unknown rather than as a tier", () => {
+    expect(normalizeRateLimitTier(null)).toBeNull()
+    expect(normalizeRateLimitTier(undefined)).toBeNull()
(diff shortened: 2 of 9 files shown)


rynfar:
Thank you @Nowaker! This has been incorporated and merged via #1081, including full desktop app parity across the main window and menu bar tray.

PR #804, opened 2026-08-12
feat(profiles): fill in a missing account plan at the next token refresh

Nowaker:

> Stacks on #795. The first commit here is that PR's; the second is this one. Read/merge #795 first.

The plan fields are written **once, at login, and never again**.

A credential file created before Meridian persisted them — or by a `claude login` that recorded `subscriptionType` without `rateLimitTier` — stays plan-blind for the life of the profile, because nothing else in the lifecycle ever asks. Measured on a ten-account host: **eight of ten profiles carried no plan at all**, and re-running an interactive login on each one just to learn the plan is not a reasonable ask.

## Why the refresh path

`GET /api/oauth/profile` needs a valid access token. A token refresh is the **one other moment** in a profile's life that holds one — every other consumer reads credentials rather than minting them. So it is the only place the gap can be closed without an interactive re-login.

## What it costs

One extra GET **per profile**, not per refresh. The lookup is gated on either field being absent, so the next refresh reads the value this one wrote and skips it. A steady-state fleet does zero extra requests.

Three further properties, all deliberate:

- **One store write.** The fetched fields are merged into the same `claudeAiOauth` object before `store.write()`, so a backfilling refresh still writes once.
- **Disk always wins.** The fetched fields are spread *under* the stored ones, so a backfill can never overwrite what a login recorded — it can only fill a hole.
- **Best-effort.** `fetchOAuthPlanFields` already swallows its own failures and returns `{}`, so a profile endpoint that is down, slow or 401ing leaves the refresh itself completely untouched. A refresh never fails because of this.

`token_refresh.success` now logs which fields were backfilled, so a run that filled a gap is distinguishable from one that had nothing to do.

## The module move

The plan lookup moves from `src/proxy/profileCli.ts` to a new leaf module `src/proxy/oauthPlan.ts`, unchanged apart from the new `planFieldsMissing` predicate.

`profileCli` already imports `tokenRefresh` for the credential store, so the refresh path importing back would close a cycle. Both callers — login and refresh — now depend on the leaf instead. The existing plan-fields tests were repointed at the new module; no assertion changed.

## Verification

`npm run typecheck` clean. `npm test` green: **2568 pass, 0 fail**, including 10 new tests for the backfill — that it fires when either field is missing, that it does *not* fire when both are present, that a fetched value never overrides a stored one, that a failing lookup leaves the refresh succeeding, and that the whole thing is still a single store write.

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -31,6 +31,10 @@ Open the printed URL in a browser, sign in to the target Claude account, then pa
 meridian profile login work --headless
 ```
 
+The same login also records the account's plan (`subscriptionType`, `rateLimitTier`), read from Anthropic's OAuth profile endpoint — the token exchange itself returns no plan information. That is what lets `meridian profile list`, `/profiles/list`, `/health` and the dashboard tell a Max account from a Team one, and what makes `/v1/models` advertise the larger context window Max accounts actually have. If the lookup fails the login still succeeds and the plan simply stays unknown.
+
+> **⚠ A profile created by an older Meridian has no plan recorded, and a token refresh cannot backfill it** — the value is only ever written at login, and Anthropic's usage endpoint does not carry it. Re-run `meridian profile login <name> --headless` to repair such a profile.
+
 #### Headless / CI: register an OAuth token
 
 When a browser isn't available (containers, CI runners, remote shells), generate a long-lived OAuth token with `claude setup-token` and register it as a profile:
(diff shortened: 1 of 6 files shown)


rynfar:
Thank you @Nowaker! This has been incorporated and merged via #1070 with original commit authorship preserved, resolving merge conflicts with the recent credential store hardening and extending account plan badges to the desktop application.

PR #805, opened 2026-08-12
feat(auth): log every property Anthropic returns during authentication

Nowaker:

> Stacks on #795. The first commit here is that PR's; the second is this one. Read/merge #795 first.

Three payloads decide what Meridian knows about an account — the token endpoint's response, `GET /api/oauth/profile`, and `claude auth status` — and **none of them was observable**.

When a field is absent, renamed, or carries a plan tier nobody has seen before, the only symptom downstream is a profile that reads `unknown`. Nothing in any log says whether the field was missing from the response or dropped on the way to disk, and that distinction is the whole of the diagnosis.

## Why `sanitize()` was not enough

`logger.ts`'s `sanitize()` redacts **by key name**. That is the wrong polarity for this problem: it protects the names somebody already thought of, and a field discovered tomorrow is by definition not one of them. Two of these three payloads carry credentials, so they cannot simply be handed to it.

## The inversion

`src/proxy/authDiscovery.ts` flips the default. **Every string is redacted to its length** unless its key is on an allow-list of values known to carry no secret — durations, timestamps, published enums, and the scope list that already appears in plaintext in the authorize URL.

A field nobody has seen before is therefore reported by **name and type** — enough to see that it arrived and to decide what it is — while its contents stay out of the log until somebody deliberately adds the key to the list.

Non-string primitives print as they are: a boolean or a number cannot be a credential, and they are exactly the fields worth reading (`loggedIn`, `expires_in`, whatever `has_*` flag comes next).

Keys that identify a **person or an organization are deliberately absent** from the allow-list. `email`, `email_address`, `name`, `display_name` and `uuid` are reported by length only, because a debug log is routinely pasted into an issue.

## Two fields per site

Every site logs `fields` (the dotted key list of what arrived) *and* `payload` (the described value), because they answer different questions. Comparing two logins reduces to comparing two key lists — which answers "did this account's response even contain the field" without reading a single value.

| event | what it captures |
|---|---|
| `auth.token_discovered` | the token exchange, logged **before** the required-token check, so a response missing one still reports what it did contain |
| `auth.profile_discovered` | the profile endpoint, alongside the plan fields actually extracted from it |
| `auth.credentials_built` | what is about to be written to disk — so a lost field localizes to the response or to this build |
| `auth.status_discovered` | `claude auth status`, from **both** readers: the sync CLI one and the TTL-cached async one every HTTP route goes through, which can disagree with it |

Each failure branch gained a matching `auth.*_failed` line, so a lookup that never happened is distinguishable from one that returned nothing.

## Cost when switched off

All of it is behind `OPENCODE_CLAUDE_PROVIDER_DEBUG`, which already gates `claudeLog`. **With the variable unset this changes no output at all** — and no behaviour is altered on any path, in either case.

## Verification

`npm run typecheck` clean. `npm test` green: **2572 pass, 0 fail**, including 14 new unit tests for the describer — that an unknown string key is reduced to its length, that every allow-listed key prints through, that nesting and arrays are walked, that identifying keys stay redacted, and that the dotted path list includes nested keys.

### DIFF
--- a/docs/profiles.md
+++ b/docs/profiles.md
@@ -31,6 +31,10 @@ Open the printed URL in a browser, sign in to the target Claude account, then pa
 meridian profile login work --headless
 ```
 
+The same login also records the account's plan (`subscriptionType`, `rateLimitTier`), read from Anthropic's OAuth profile endpoint — the token exchange itself returns no plan information. That is what lets `meridian profile list`, `/profiles/list`, `/health` and the dashboard tell a Max account from a Team one, and what makes `/v1/models` advertise the larger context window Max accounts actually have. If the lookup fails the login still succeeds and the plan simply stays unknown.
+
+> **⚠ A profile created by an older Meridian has no plan recorded, and a token refresh cannot backfill it** — the value is only ever written at login, and Anthropic's usage endpoint does not carry it. Re-run `meridian profile login <name> --headless` to repair such a profile.
+
 #### Headless / CI: register an OAuth token
 
 When a browser isn't available (containers, CI runners, remote shells), generate a long-lived OAuth token with `claude setup-token` and register it as a profile:
--- a/src/__tests__/auth-discovery.test.ts
+++ b/src/__tests__/auth-discovery.test.ts
@@ -1,0 +1,168 @@
+/**
+ * Unit tests for the auth-payload describer.
+ *
+ * The module is pure, so these are direct assertions with no mocking. The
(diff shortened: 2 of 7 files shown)


rynfar:
Incorporated and merged via maintainer PR #1083 (squash commit onto main), bringing safe auth discovery logging with length-based redaction and full test coverage. Thank you @Nowaker!

PR #807, opened 2026-08-14
fix(session): retry a refused resume before abandoning the session

justprosh:
## Problem

The CLI refuses a `--resume` that lands while the previous subprocess for the same session is still exiting. One of its wordings — `No conversation found with session ID …` — reads like the session no longer exists, but it is intact on disk and the next attempt succeeds.

The proxy took that first refusal as terminal: it evicted the session and replayed the whole conversation as a fresh one. Nothing surfaces to the client — the response is a normal 200 — while the turn silently loses the context the SDK side held, re-primes the prompt cache from zero, and costs a full-history replay.

The other wording of the same race, `is currently running as a background agent`, already received retries. So one phrasing of one transient refusal was handled and the other destroyed a live session.

## Solution

Classify the refusal once, in `errors.ts`, instead of matching its text at each call site. Three verdicts, because the refusal genuinely distinguishes three situations with three different recoveries:

- **`busy`** — the session exists and is held by a running agent, so it can be branched.
- **`unresumable`** — it would not open. This is *not* proof it is gone.
- **`missing-message`** — a message inside it is gone, so an identical attempt fails identically.

Both "not right now" verdicts retry the same resume with linear backoff. Only `busy` is forked afterwards — an unresumable session offers nothing to branch, where the previous shape reported a doomed fork into nothing. A fresh replay becomes the last resort for a session that refused to open and then spent its retries, and stays the immediate answer for a lost message.

Since the wording can alternate inside one request, the replay is keyed on having seen an unresumable refusal at all rather than on what the last attempt was refused with; otherwise a mixed sequence exhausts the budget and the refusal escapes to the client.

Telemetry follows the same split: `session.resume_retry` / `session.resume_replay` carry the verdict, so a retry is no longer filed under a cause it may not have. The replay log line gains the request id it was missing. The delay env var keeps its name — renaming it would drop existing overrides.

**Cost of the change:** a session that truly cannot be resumed now pays the retry budget (0.5 s + 1 s + 1.5 s) before the fresh replay it previously got immediately. That includes the undo path when the rollback fails with the missing-conversation wording. Losing a live session was judged the worse outcome.

## Verification

`src/__tests__/proxy-resume-refusal.test.ts`, five cases through the HTTP layer with a mocked SDK. Against `main`, three fail:

- a transient refusal is retried with the same resume, streaming and non-streaming — on `main` the second query carries no resume at all, the session is already evicted;
- a session refused on every attempt still reaches a fresh full-history replay, and the resume is attempted exactly `1 + retries` times — on `main` it replays after a single refusal;
- a wording that alternates mid-request still replays instead of surfacing a 500 (red on an intermediate shape of this branch, green here);
- a refusal naming a lost message is answered by one fresh replay and never retried.

`errors.test.ts` pins each verdict, and pins that a rate limit, an auth failure and a bare exit code stay unclassified so the surrounding handlers keep them. The existing busy-refusal tests pass unchanged — that path's behavior, including its fork, is untouched.

Full suite green, `tsc --noEmit` clean.

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -2,7 +2,7 @@
  * Unit tests for classifyError — pure function, no mocks needed.
  */
 import { describe, it, expect } from "bun:test"
-import { classifyError, extendedContextHint, isStaleSessionError, isBusySessionError, isExtraUsageRequiredError, extractSdkTermination, formatSdkTermination, isAccountFailoverError, isQuotaRefusal } from "../proxy/errors"
+import { classifyError, extendedContextHint, classifyResumeRefusal, isBusySessionError, isExtraUsageRequiredError, extractSdkTermination, formatSdkTermination, isAccountFailoverError, isQuotaRefusal } from "../proxy/errors"
 
 describe("classifyError", () => {
   describe("authentication errors", () => {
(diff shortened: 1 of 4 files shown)


rynfar:
Merged via #811 (`8455bd66`), with your authorship preserved on the commit and a `Co-authored-by:` trailer on the squash.

This wasn't a problem with the change — `main` requires signed commits, so PRs from forks sit at `BLOCKED` regardless of how green the checks are. Cherry-picking re-signs the commit with a key the branch protection accepts while keeping you as `Author`. Nothing you could have done from your side.

Thanks for this one — the writeup made it easy to review. Two things I verified independently and both held: reverting `errors.ts` and `server.ts` to `main` fails exactly the three tests you named, and the replay is genuinely one-shot (`yield*` then `return`), so the sticky `sawUnresumableRefusal` flag can't drive an eviction loop — which is the first thing I went looking for.

One note for the release: the telemetry keys `session.busy_retry` → `session.resume_retry` and `session.stale_uuid_retry` → `session.resume_replay` are a silent break for any external dashboard. Nothing in-repo consumes them, so it's a release-note line rather than a change request.

PR #810, opened 2026-08-14
fix(concurrency): stop internal dispatch hops from taking a second queue slot

### LINKED ISSUE #809: Priority routing deadlocks the SDK concurrency queue: one request takes two slots
With `routing="priority"`, one external request consumes **two** of the
`MAX_CONCURRENT_SESSIONS` slots. Once concurrent requests reach the limit, the
queue stops draining and never recovers — every later `/v1/messages` hangs
until the process is restarted, while `/health` keeps answering `healthy`.

## Mechanism

`dispatchPriority` dispatches by re-entering the app over `app.fetch`
(`src/proxy/server.ts:663`), and the route it re-enters is the queued one:

```
POST /v1/messages
  └─ handleWithQueue → acquireSession()          ← slot #1 (outer pass)
      └─ dispatchPriority → app.fetch(/v1/messages)
          └─ handleWithQueue → acquireSession()  ← slot #2 (inner pass)
```

The outer pass holds slot #1 for as long as the inner pass runs. With N slots
and N concurrent requests, every slot is held by an outer pass waiting on an
inner pass that can never be admitted, and `releaseSession` is only reached
after the inner one returns — a deadlock, not a slowdown. Raising
`MERIDIAN_MAX_CONCURRENT` only moves the threshold.

## Reproduction

Deterministic — with a single slot, one request already needs two:

```
MERIDIAN_MAX_CONCURRENT=1 MERIDIAN_ROUTING=priority   # two profiles configured
curl -sS localhost:3456/v1/messages -H 'content-type: application/json' \
  -d '{"model":"claude-sonnet-4-5","max_tokens":64,"stream":false,
       "messages":[{"role":"user","content":"hi"}]}'
# never returns; /health still answers
```

Seen on a 1.61.0 deployment: the log counter climbs in steps of two (one step
per external request), five concurrent `stream=false` requests reached
`active=10/10` at the default limit, and everything after that timed out until
the unit was restarted.

## Fix

The slot belongs to the external request; the inner pass is that same request
continuing, and it is the only one of the two that spawns an SDK subprocess.
PR #810 runs it on the slot the outer pass already holds, marking it via
`AsyncLocalStorage` — not a header, which would reach the limiter from the wire
and let any client opt out of the limit.

## Related, not in the PR

The inner `Request` is built without a `signal`, so a client disconnect never
reaches the inner hop: the outer pass keeps waiting, and keeps its slot, on
work nobody reads any more. Happy to file separately.

justprosh:
Fixes #809.

`dispatchPriority` re-enters the app over `app.fetch`, and the route it
re-enters is the queued one — so one external request makes two passes through
`handleWithQueue` and takes two slots, the outer one held for the whole life of
the inner. At N concurrent requests every slot is held by an outer pass waiting
on an inner pass that can never be admitted: the queue stops draining for good,
while `/health` keeps answering because it never enters the queue.

The slot belongs to the external request, not to the pass. The inner pass is
that same request continuing, and it is the only one of the two that spawns an
SDK subprocess — so it now runs on the slot the outer pass already holds.
Failover stays sequential, so one request still means one subprocess at a time.

The marker rides an `AsyncLocalStorage` rather than a header on purpose: a
header reaches the limiter from the wire, and any client could then opt out of
the limit. It also carries the outer pass's queue timings, so the inner pass —
the one that records telemetry — still reports the wait actually served instead
of a zero of its own.

## Tests

New `src/__tests__/proxy-concurrency-limiter.test.ts`, three cases, all failing
on `main` before this change: one request against one slot (deadlock), two
concurrent requests against two slots (only one admitted), a burst of six
against two (deadlock). The first is deterministic — with a single slot one
request already needs two. The last also asserts peak SDK concurrency stays
within the limit, so the fix can't degenerate into not queueing at all.

`bun run test`, `bun run typecheck` and `bun run build` all exit 0.

### DIFF
--- a/src/__tests__/proxy-concurrency-limiter.test.ts
+++ b/src/__tests__/proxy-concurrency-limiter.test.ts
@@ -1,0 +1,172 @@
+/**
+ * The SDK concurrency limiter under priority routing.
+ *
+ * Priority routing dispatches by re-entering this same app over `app.fetch`
+ * (dispatchPriority), so one external request produces two passes through the
+ * queued route. The limiter counts passes, not external requests, so the outer
+ * pass holds a slot while waiting for an inner pass that needs one of its own.
+ * At N concurrent external requests every slot is held by a waiter and nothing
+ * can be admitted — a permanent deadlock, not a slowdown: on staging the
+ * counter reached 10/10 and every subsequent request timed out until the
+ * process was restarted, while /health kept answering.
+ *
+ * These tests fix the invariant the limiter is meant to express: one external
+ * request occupies exactly one slot, and the limit still bounds how many SDK
+ * subprocesses run at once.
+ */
+import { describe, it, expect, mock, beforeEach, afterEach } from "bun:test"
+import { assistantMessage } from "./helpers"
+
+let sdkActive = 0
+let sdkPeak = 0
+/** Held open while a test needs SDK calls to overlap observably. */
+let sdkGate: Promise<void> = Promise.resolve()
+let openGate: () => void = () => {}
+
+function gateClosed() {
+  sdkGate = new Promise<void>((resolve) => { openGate = resolve })
+}
+
+mock.module("@anthropic-ai/claude-agent-sdk", () => ({
+  query: () => (async function* () {
+    sdkActive++
+    sdkPeak = Math.max(sdkPeak, sdkActive)
+    try {
+      await sdkGate
+      yield assistantMessage([{ type: "text", text: "ok" }])
+    } finally {
+      sdkActive--
+    }
+  })(),
+  createSdkMcpServer: () => ({ type: "sdk", name: "test", instance: {} }),
+  tool: () => ({}),
+}))
+
+mock.module("../logger", () => ({
+  claudeLog: () => {},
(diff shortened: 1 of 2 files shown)


rynfar:
Merged via #813 (`1bfaaa20`), authorship preserved, with a `Co-authored-by:` trailer on the squash.

Nothing was wrong with the PR — `main` requires signed commits, so fork PRs sit at `BLOCKED` no matter how green the checks are. Cherry-picking re-signs it with a key branch protection accepts while keeping you as `Author`.

Three things I verified independently rather than taking from the description, all of which held:

- Reverting `server.ts` to `main` fails all three tests with 5s/5s/15s timeouts — the deadlock signature, not a flake.
- The re-entry is an in-process `await app.fetch(...)`, so the `AsyncLocalStorage` context genuinely propagates. Worth stating explicitly since this would silently not work across a real network boundary.
- Composition holds. Only `/v1/messages` and `/messages` are queued, so `/v1/chat/completions` (which re-enters from outside the queue) takes exactly one slot before and after, and stacking all three layers still yields one.

The `sdkPeak > 0` assertion alongside `sdkPeak <= 2` is the detail I appreciated most — without it the fix could have degenerated into not queueing at all and the test would still have passed. Thanks for both of these.

PR #818, opened 2026-08-16
fix(profiles): an account with no stored token is not logged in

Nowaker:

`/profiles/list` reports `loggedIn` straight from `claude auth status`, which
answers from its own view of the account rather than from the credential a
request will actually present.

Measured on a ten-account fleet: **three accounts whose stored `accessToken`
was the empty string were reported `loggedIn: true`.** The profiles page showed
them as ordinary idle accounts - "no usage data yet" - while every request
routed to them came back `401 authentication_error`:

```
$ curl -H 'x-meridian-profile: corp' .../v1/messages
HTTP 401  error.type=authentication_error

$ curl .../profiles/list | jq '.profiles[] | select(.id=="corp") | .loggedIn'
true
```

A fourth account in the same state *did* read as needing login. It differed
only in never having authenticated within that process's auth-cache lifetime,
so which of the four told the truth was an accident of timing rather than a
fact about the credential.

## The change

`/profiles/list` now reads the profile's own credential store and reports
`loggedIn: false` when it holds no access token. The credential is what a
request presents, so an access token that is not there outranks a cheerful
probe.

**Only a credential that was read successfully and carries no access token
demotes.** `CredentialStore.read()` answers `null` both for a credential that
is absent and for one it could not parse, so from the caller the two are
indistinguishable - and a Keychain that momentarily refuses, or a file caught
mid-write, must never report a working account as logged out. That case is
`unknown` and leaves the existing answer alone.

## Scope

Deliberately `/profiles/list` alone. `/health` has the same blind spot for the
active profile, but it answers 503 on a logged-out account, and a false 503 on
a monitoring endpoint is a worse failure than the one being fixed here. Happy
to extend it if you would rather.

## Testing

- 6 new unit tests covering present / empty-string / missing-oauth-block /
  non-string token / `null` read / throwing read.
- `tsc -p tsconfig.json --noEmit` clean.
- Full suite 2443 pass / 102 fail, against an **identical failure set** on
  `main` at 7c8d2b4 - the same 102 tests, no difference in either direction.
- Verified live against a real instance: the four affected accounts flipped to
  `loggedIn: false`, the seven healthy ones stayed `true`.

### DIFF
--- a/src/__tests__/credential-presence-unit.test.ts
+++ b/src/__tests__/credential-presence-unit.test.ts
@@ -1,0 +1,60 @@
+/**
+ * Unit tests for readStoredCredentialPresence — the cross-check that stops
+ * /profiles/list calling a logged-out account authenticated.
+ *
+ * The case that motivated it, measured on a ten-account fleet: three accounts
+ * whose stored accessToken was the empty string were reported `loggedIn: true`
+ * by `claude auth status`, so the profiles page showed them as fine while every
+ * request routed to them returned 401 authentication_error.
+ *
+ * The store is dependency-injected rather than the platform one, so the file
+ * touches no real credential and is safe to run beside every other suite.
+ */
+
+import { describe, expect, test } from "bun:test"
+import { readStoredCredentialPresence, type CredentialStore } from "../proxy/tokenRefresh"
+
+function storeReading(value: unknown): CredentialStore {
+  return {
+    async read() { return value as any },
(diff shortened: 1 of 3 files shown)


rynfar:
Incorporated into main via PR #1069 with original commit authorship preserved and extended with desktop status visibility ('Needs login' / 'Sign-in required'). Thank you @Nowaker!

Issue #478, opened 2026-05-04
After update to v1.41.0 getting error: Script not found "stream-json"


stefanpartheym:
I updated to meridian v1.41.0 and I'm now getting the following error whenever i try to send a prompt: 

```
sdk_termination reason=process_exit exit=1 model=opus[1m] resume=false deferred=false stderr="error: Script not found \"stream-json\""
```

rynfar:
do you have any more context for this by chance you can share?  Anything you can tell me that lead up to this?  Is it happening on every request now?

Which harness are you using?  Opencode?  Any telemetry you can share would be helpuful!  @stefanpartheym

stefanpartheym:
Hey @rynfar , yes sorry that was a bit brief.

- It is happening on every request (even with v1.40.0).
- I'm using OpenCode.
- Logs from meridian:

```json
[
    {
        "level": "error",
        "category": "error",
        "message": "b0e73953-5bf1-4c9b-94b9-9c1061d05813 sdk_termination reason=process_exit exit=1 model=opus[1m] resume=false deferred=false stderr=\"error: Script not found \\\"stream-json\\\"\"",
        "requestId": "b0e73953-5bf1-4c9b-94b9-9c1061d05813",
        "timestamp": 1777913765421
    },
    {
        "level": "info",
        "category": "session",
        "message": "b0e73953-5bf1-4c9b-94b9-9c1061d05813 adapter=opencode model=opus[1m] stream=true tools=11 lineage=new session=new agent=primary active=1/10 msgCount=1",
        "requestId": "b0e73953-5bf1-4c9b-94b9-9c1061d05813",
        "timestamp": 1777913765416
    },
    {
        "level": "error",
        "category": "error",
        "message": "10128d26-4ecc-4a76-beb2-cd5fb5482080 sdk_termination reason=process_exit exit=1 model=opus[1m] resume=false deferred=false stderr=\"error: Script not found \\\"stream-json\\\"\"",
        "requestId": "10128d26-4ecc-4a76-beb2-cd5fb5482080",
        "timestamp": 1777913757402
    },
    {
        "level": "info",
        "category": "session",
        "message": "10128d26-4ecc-4a76-beb2-cd5fb5482080 adapter=opencode model=opus[1m] stream=true tools=11 lineage=new session=new agent=primary active=1/10 msgCount=1",
        "requestId": "10128d26-4ecc-4a76-beb2-cd5fb5482080",
        "timestamp": 1777913757397
    },
    {
[...]

PR #819, opened 2026-08-17
feat(health): /livez and /readyz, because /health answers neither question

Nowaker:

`/health` is the only health surface Meridian has, and it answers a question
that is neither of the two an orchestrator asks.

**As a liveness probe it is a restart loop waiting to happen.** It verifies the
*active profile's* auth and returns 503 when that profile is logged out. No
number of restarts re-authenticates an account, so a supervisor pointed at it
kills a perfectly healthy proxy over and over until a human runs `claude login`.

**As a readiness probe it fails in the other direction.** It inspects one
profile out of however many are configured, and its own catch-all returns
`degraded` with status **200** - so an instance that cannot verify auth *at
all* reads as ready.

This adds the two endpoints that answer the two questions, named and shaped
after kube-apiserver's own so a Kubernetes probe, a Caddy `health_uri` and a
person with `curl` all read the same thing:

| | question | checks |
|---|---|---|
| `GET /livez` | is this PROCESS turning - would a restart help? | none |
| `GET /readyz` | should traffic come to THIS instance rather than another? | `profiles`, `claude-executable` |

```
$ curl -s localhost:3456/readyz
ok

$ curl -s 'localhost:3456/readyz?verbose'
[+]profiles ok
[+]claude-executable ok
readyz check passed

$ curl -s localhost:3456/readyz          # an instance with no profiles.json
[-]profiles failed: no profiles configured
readyz check failed
```

## Liveness passes unconditionally, and that is the design

Reaching the handler at all means the event loop is turning and the HTTP server
is answering, which is the entire question. Every dependency added there - a
subprocess, a credential read, a reachable upstream - is another way for this
process to be killed over somebody else's outage, and a liveness probe that
restarts a working process is strictly worse than no liveness probe at all.

## Readiness admits only PER-INSTANCE checks

The point of readiness is to move traffic somewhere better, so a check that two
instances fail *together* cannot move it anywhere - it can only take every
instance out at once and turn a clear error into a 502.

That rule is why the credentials are deliberately **not** checked here, even
though a proxy with no usable token serves nothing: instances share those files,
so their loss is no reason to prefer one instance over another, and the honest
answer to the caller is the provider's own 401 rather than a gateway error.

What is left is genuinely local to one process:

- **`profiles`** - an instance whose config and whose `profiles.json` are both
  empty has no account to serve from while a neighbour reading a populated one
  is fine. In a container that is entirely per-instance: the file is under that
  container's own `HOME`, so a missing mount takes exactly one replica out.
- **`claude-executable`** - resolved from `MERIDIAN_CLAUDE_PATH`, this install's
  own `node_modules`, or its platform package, all per-instance, and without one
  no request can be served (#478). The cached answer is used first so the steady
  state costs nothing; the sync lookup runs only before the first SDK call has
  populated that cache, where the alternative is reporting a freshly started
  instance unready.

## Both are public, on purpose

They emit `ok`, or a check *name* and pass/fail - no account, no email, no
token, no profile id. That is strictly less than `/health` beside them already
gives any unauthenticated caller, which answers with the signed-in email and
subscription type.

Gating them would be worse than useless: a 401 is what a load balancer reads as
"this backend is down", so setting `MERIDIAN_API_KEY` would make every instance
look unhealthy at once and take down whatever sits in front - the exact failure
this pair exists to prevent. The auth-audit test added in #477 caught the new
routes immediately, which is the test doing its job; its allowlist now carries
that review beside the entry rather than in a commit message nobody will find.

## Measured, not only unit-tested

Against a real instance on a spare port:

- with no `profiles.json` reachable at all, `/readyz` answered **503** naming
  the failed check while `/health` answered **200**;
- with the real config, both answered 200 and `?verbose` listed both checks;
- a probe costs **5ms** end to end, including `curl`'s own startup.

## Tests

12 new unit tests covering every branch of the decision logic and the render
shape, including the two that matter most: liveness passing with no checks at
all, and the terse failing form still naming the failed check, because a 503
nobody can explain is a 503 somebody disables.

`tsc --noEmit` clean. Full suite run on this branch and on `main`, failure sets
compared with `comm` in both directions rather than by count: **identical**
(102 pre-existing failures either side, none introduced, none fixed).

## Where this is used

Behind a Caddy `reverse_proxy` with two Meridian instances and
`lb_policy first`, `/readyz` is what decides which one serves. Verified by
stopping the primary: the vhost kept answering 200, served by the standby, and
returned to the primary by itself when it came back.

### DIFF
--- a/src/__tests__/probes-unit.test.ts
+++ b/src/__tests__/probes-unit.test.ts
@@ -1,0 +1,94 @@
+/**
+ * Unit tests for the liveness/readiness probes.
+ *
+ * The decision logic is pure so every branch is provable without an HTTP
+ * server, a subprocess or a credential file - which is the same reason the
+ * probes themselves touch none of those at request time.
+ */
+
+import { describe, expect, test } from "bun:test"
+import { livenessReport, readinessReport, renderProbe } from "../proxy/probes"
+
+const READY = { profileCount: 3, claudeExecutableResolved: true }
+
+describe("livenessReport", () => {
+  test("passes", () => {
+    expect(livenessReport().ok).toBe(true)
+  })
+
+  // Pinned deliberately: the value of a liveness probe is everything it does
+  // NOT depend on. A check added here is a way for a working process to be
+  // killed over somebody else's outage, so growing this list is the
+  // regression, not a feature.
+  test("has no checks at all, which is the design", () => {
+    expect(livenessReport().checks).toEqual([])
+  })
+})
+
+describe("readinessReport", () => {
+  test("is ready with profiles and a resolved executable", () => {
+    const report = readinessReport(READY)
+    expect(report.ok).toBe(true)
+    expect(report.checks.every(c => c.ok)).toBe(true)
+  })
+
+  // MERIDIAN_CONFIG_DIR relocates the whole config directory, so an instance
+  // pointed at an empty one has no accounts while its neighbour is fine. That
+  // is exactly the per-instance failure readiness exists to route around.
+  test("is not ready with no profiles, and says which check failed", () => {
+    const report = readinessReport({ ...READY, profileCount: 0 })
+    expect(report.ok).toBe(false)
+    const failed = report.checks.find(c => c.name === "profiles")
+    expect(failed?.ok).toBe(false)
+    expect(failed?.detail).toContain("no profiles configured")
+  })
+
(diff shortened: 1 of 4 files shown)


rynfar:
Thank you @Nowaker! This has been incorporated and merged via #1072 with original commit authorship preserved, resolving import conflicts with recent telemetry/model exports and integrating the liveness probe into the desktop application.

PR #821, opened 2026-08-17
fix: coordinate SDK/session concurrency and add graceful shutdown

debin-ge:
## Summary

- coordinate SDK subprocesses with an abortable process-wide semaphore and serialize turns for the same client session
- add graceful draining/shutdown behavior and expose separate session/SDK queue timing in telemetry
- document the new concurrency and shutdown controls and add focused regression coverage

## Testing

- `npm test` — 2576 passed, 1 skipped, 0 failed in the main suite; all serial suites passed
- `npm run build` — passed

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -37,6 +37,8 @@ Agent (OpenCode) ◄── SSE Response ◄────────────�
 src/
 ├── proxy/
 │   ├── server.ts              ← HTTP layer: routes, SSE streaming, concurrency, request orchestration
+│   ├── concurrency.ts         ← Abortable SDK query semaphore and concurrency config parsing
+│   ├── shutdown.ts            ← Bounded HTTP drain and connection tracking
 │   ├── adapter.ts             ← AgentAdapter interface (extensibility point for multi-agent support)
 │   ├── adapters/
 │   │   ├── opencode.ts        ← OpenCode adapter (session headers, CWD extraction, tool config)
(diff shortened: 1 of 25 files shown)


debin-ge:
CI and Docker are currently awaiting administrator approval because this is a fork pull request. Local validation has completed successfully: `npm test` passed (2576 passed, 1 skipped, 0 failed in the main suite; all serial suites passed) and `npm run build` passed. Please approve the workflow runs when convenient.

rynfar:
Thanks for this — the design here was sound, and it's now merged to `main` via #825 as commit 83b2f5c.

Your commit was cherry-picked unmodified onto current `main` (#815 had landed in the meantime), so **you remain its git author**, and you're credited as co-author on the squashed commit. The only thing I changed was normalising the author email — the original had literal spaces (`gedebin@MacBookPro.lan guest wan`), which meant GitHub couldn't link the commit to your account.

Replacing the `activeSessions` counter plus the `AsyncLocalStorage` re-entrancy hack with an abortable FIFO semaphore and a per-session turn coordinator was the right call, and calling `handleMessages` in-process rather than re-entering `app.fetch` is the correct fix for the `insideSessionSlot` deadlock. Holding the SDK permit for the whole streaming lifecycle also fixed a real pre-existing bug: the old code released it before the stream body ran, so streaming subprocesses were effectively unbounded.

Review turned up four things that were corrected on top, each with a test that fails without the fix:

1. The drain contract wasn't actually honoured on `/v1/chat/completions` and `/v1/responses` — the gate was only on the internal `/v1/messages` hop, and both routes flattened the 503 into `upstream_error`, dropping the `x-meridian-draining` header.
2. `sdkQueueWaitMs` was credited only from a granted lease, so an aborted wait landed in `proxyOverheadMs`; and a cancel while waiting for the session lease recorded no row at all.
3. `forkAttemptMeta` rewrote a caller's `x-request-id` per attempt. The comment justified it by saying same-id rows overwrite each other — they don't; SQLite always INSERTs against a non-unique `request_id` and the memory store is a ring buffer.
4. A second SIGINT/SIGTERM was swallowed, making the process unkillable for the full 30s grace window.

Three deeper items are filed as #827, #828 and #829 rather than rushed into the merge.

Closing this in favour of #825 — the work itself shipped. Appreciate the contribution.

PR #822, opened 2026-08-17
feat(profiles): show the organization an account belongs to, and its details on hover

Nowaker:

Two related additions to the profile UI, sharing one new field.

**1. The organization an account belongs to.**

A fleet of accounts named `work2`, `corp3`, `personal2` gives no way to tell whose
subscription each one actually is. Anthropic already answers that: the same
`GET /api/oauth/profile` call Meridian makes for the plan also returns
`organization.name`. This keeps it.

It is stored in a Meridian-owned `organizations.json` beside `settings.json`,
**not** in `.credentials.json`. That file's keys are Claude Code's own, and
Claude Code rewrites it on its next token refresh - a Meridian-invented key
there would both break the "indistinguishable from `claude login`" property and
be dropped within hours. Fetching live per render is also out: eleven profiles
would mean eleven authenticated round trips on every page load. The cache is
refreshed opportunistically, off the request path, with a retry floor.

`GET /profiles/list` gains `organizationName`, present on every profile with
`null` when unknown, so a consumer can tell "no organization" from "this
Meridian is too old to answer". The `/profiles` card gains an `Organization`
row, omitted entirely when absent rather than rendered as "unknown".

The name is deliberately **not** added to `SAFE_AUTH_STRING_KEYS` - that
allowlist governs debug logs, which get pasted into issues. Rendering it on the
operator's own screen is fine; putting every account's organization into a
shareable log is not. The discovery log records the name's *length*, matching
what `authDiscovery.ts` already does for the same field.

**2. An (i) overlay on the home page.**

Each account card gets a small (i) after the status dot and before the name.
Hovering or focusing it draws a panel with the same label/value list the
`/profiles` card shows - type, status, organization, plan, when it was last
verified. Not a `title` attribute: the native tooltip cannot render a
label/value grid.

Both surfaces render from one shared builder (`src/telemetry/profileFacts.ts`),
so the two lists cannot drift - adding a row adds it in both places. That
builder ships browser source and its tests evaluate that exact text, so the
tested function and the shipped function are the same string.

The icon opts out of the card's click and keydown handlers: the card is itself
the switch button, so reading an account must not move all traffic to it. The
10s poll is suppressed while an overlay is open, so it cannot be yanked away
mid-read.

**Testing.** `tsc --noEmit -p tsconfig.json` clean. Full suite compared against
`main` with `comm` in both directions on the deduped failure set: identical,
zero new and zero disappeared, plus 48 new passing tests.

### DIFF
--- a/bin/cli.ts
+++ b/bin/cli.ts
@@ -204,6 +204,9 @@ export async function runCli(
     enableDiskProfileDiscovery()
   }
 
+  const { enableOrganizationLookup } = await import("../src/proxy/organizationName")
+  enableOrganizationLookup()
+
   const proxy = await start({ port, host, idleTimeoutSeconds, pluginDir, pluginConfigPath, profiles, defaultProfile, version, installProcessErrorHandlers: true })
 
   // Handle EADDRINUSE — preserve CLI behavior of exiting on port conflict
--- a/src/__tests__/organization-name.test.ts
+++ b/src/__tests__/organization-name.test.ts
@@ -1,0 +1,274 @@
+/**
+ * Unit tests for organizationName.ts.
+ *
+ * The pure readers are exercised directly. The persistence round trip goes
+ * through the real module with MERIDIAN_CONFIG_DIR redirected, the same
+ * arrangement settings-unit.test.ts uses — a re-implementation of the JSON
+ * handling would pass regardless of what the module actually did.
+ */
+import { describe, test, expect, beforeEach, afterEach } from "bun:test"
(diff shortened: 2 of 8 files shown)


rynfar:
Incorporated and merged via maintainer PR #1082 (commit 7651b3ea8bfb2f7b57bc0d93720a60dfb785acf0), which brings this organization display to the web UI, /profiles page, and desktop app cards/tray with unified facts. Thank you @Nowaker!

PR #824, opened 2026-08-17
feat: keep the last good usage reading when a check fails

Nowaker:

## What this changes

When the OAuth usage endpoint refuses, the profiles page no longer throws away the last good reading. It keeps the bars, percentages and reset countdowns on screen, says when that reading was taken, and says which check failed, why, and how many have failed in a row.

Before, `renderUsageSection` replaced the entire usage block with one sentence:

> Usage data is rate limited upstream — retrying shortly.

A reading that was often only minutes old vanished along with it. Across a twelve-account fleet the endpoint refuses often enough for this to be a daily occurrence rather than a theoretical one.

## It turned out to be a rendering change, not a new cache

The first thing I checked was whether a cache already existed. It does: `src/proxy/oauthUsage.ts` keeps a per-profile snapshot with a 30-second TTL and a 15-minute stale fallback, and `staleOr` already served the last good snapshot on transient failure.

So no new cache was added, which is much the better outcome. What was missing was **provenance** — the data reached the consumer as bare numbers, with no way to tell three states apart:

| state | before | after |
|---|---|---|
| never read | empty `windows`, `error` set | empty `windows`, `stale: false`, `fetchedAt: null` |
| read and fresh | numbers | numbers, `stale: false`, `failure: null` |
| read then stale | numbers, indistinguishable from fresh | numbers, `stale: true`, `fetchedAt` dated, `failure` explaining |

A window sitting at 0% and a profile that has never been read flatten to the same numbers, so the distinction has to travel as its own field rather than being inferred from emptiness.

## API

`fetchOAuthUsageResult` gains two fields:

- `failure` — `{ reason, consecutiveFailures, lastFailureAt }`, or `null` when the last check succeeded. The reason reuses the `OAuthUsageError` vocabulary from #789, so a rate limit stays distinguishable from an auth failure: one is temporary and the other wants `claude login`.
- `lastGood` — the last successful snapshot at any age, including past the 15-minute bound within which `staleOr` will serve it as `snapshot`.

`toUsageEntry` shapes those into the per-profile wire fields `/v1/usage/quota/all` serves, adding `stale` and `failure` next to the existing `windows`, `extraUsage`, `fetchedAt` and `error`.

**`error` keeps its exact previous meaning** — set only when there was nothing fresh *and* no recent-enough reading to stand in. A consumer that ignores the new fields behaves as it did before.

## Why the failure count skips cooldown-suppressed polls

`consecutiveFailures` counts checks that actually reached the credential store or upstream. The 429 cooldown turns polls away before either, and those do not advance the count.

The profiles page polls every 10 seconds; the backoff runs for a minute or more. Counting suppressed polls would report six failures for one refusal — a number that describes the poll interval, not the account.

## What it looks like

Read and fresh, unchanged:

> **USAGE**  ·  updated just now
> `[5h 36%]` `[7d 5%]`

Read then stale:

> **USAGE**  ·  updated 6m ago
> Last check failed (rate limited upstream) — figures below are the last successful read.
> `[5h 36%]` `[7d 5%]`

> **USAGE**  ·  updated 41m ago
> Last 4 checks failed (usage endpoint unavailable) — figures below are the last successful read.
> `[5h 36%]` `[7d 5%]`

Never read at all, which is the one case where there is genuinely nothing to show:

> **USAGE**
> No reading yet — last check failed (rate limited upstream), retrying.

The note uses `--yellow` at 11px, matching `.usage-empty`'s size for informational lines, and sits between the section title and the bar grid. Reasons map through a fixed vocabulary in `describeFailedRun` rather than being echoed, so nothing the server sends reaches the DOM.

## Testing

Nine new unit tests covering the state machine:

- a first-ever failure with nothing cached
- a failure after a success, which keeps the reading and dates it
- three consecutive failures counting to 3
- a success clearing the run rather than zeroing it
- cooldown-suppressed polls not advancing the count
- `toUsageEntry` for each of never-read, read-and-fresh, and stale-past-the-bound

`bunx tsc --noEmit` is clean.

The provenance work adds 16 tests in `src/__tests__/cached-facts.test.ts`: thirteen over the pure module (each of the three states, the omit case, escaping, ordering of the marker after the value, and the status-class join) and three asserting the page itself renders the marker.

`bunx tsc --noEmit` and the full `bun run test` gate both exit 0 on this branch: **2781 passing before the provenance commit, 2797 after**, with no failures in either run. An earlier revision of this description reported 105 pre-existing failures; that reflected the local `~/.config/meridian` state at the time, since several integration tests read the developer's real account list. Measured today the suite is green.

A control run confirms the new tests are load-bearing rather than decorative: with `cachedFacts.ts` present but `profilePage.ts` and `server.ts` reverted to their previous state, **exactly the three page-contract tests fail** while the thirteen module tests pass. With the module removed as well, the test file fails to import at all.

The rendered output of all six states was verified in a browser against a fixture server serving the real page module — no console errors or warnings, checked at desktop width and at 420px where the bar grid reflows to one column.

## Marking a remembered value as remembered

Keeping the reading was only half of it. A figure the check took thirty seconds ago and a figure it took an hour ago and could not re-confirm rendered identically, so a reader still had no way to tell a current number from a remembered one — which is the whole point of having kept it.

Every value now carries its provenance immediately to the right of itself:

> **USAGE**  ·  updated 6m ago
> Last check failed (rate limited upstream) — figures below are the last successful read.
> `[5h 36% (cached)]` `[7d 5% (cached)]`

Per value rather than once per card, because a card routinely mixes the two — the auth status can be live while the email and plan beside it are what an earlier check left behind. One banner across the card would mislabel whichever half it did not apply to. A value read by the check that just ran carries no marker at all.

### "No value" was two different facts wearing one blank

An account whose plan became unreadable and an account that never had a plan both rendered as a missing row. That is the more misleading of the two: a missing row reads as "no such thing" rather than "could not look".

`src/telemetry/cachedFacts.ts` names the three states and renders each distinctly:

| state | meaning | rendering |
|---|---|---|
| `live` | read by the check that just ran | the value, unmarked |
| `cached` | read earlier; the check that just ran could not confirm it | the value, followed by `(cached)` |
| `never` | never read successfully for this profile | `never read`, muted |

`renderFactValue` returns `null` for a fourth case — a value absent from a check that *succeeded*, which is genuinely not applicable to that account rather than unreadable — so those rows stay omitted exactly as before.

### `/profiles/list` gains `authProvenance`

`getClaudeAuthStatusAsync` already returns its `lastKnownGood` when a check fails, so `email`, `subscriptionType` and `loggedIn` were routinely remembered values wearing the exact shape of fresh ones, and the route had no way to admit it. `authProvenance` is `"live"`, `"cached"` or `"never"` accordingly.

The field is additive. A consumer ignoring it sees exactly what it saw before.

### The inline mirror is checked, not assumed

`profilePage.ts` is one template literal that runs in the browser and cannot import, so it mirrors the module in its inline script — the convention `profileUsage.ts` already documents. Rather than trust that the copy still matches, the inlined block was extracted from the rendered page, executed, and diffed against the module across all three states plus a hostile value, an empty string, and both status-class variants. All eleven cases agree.

## Scope

The diff stays inside the usage section of `src/telemetry/profilePage.ts`, which has concurrent changes landing in other areas of the same file.

### DIFF
--- a/src/__tests__/cached-facts.test.ts
+++ b/src/__tests__/cached-facts.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Provenance markers for remembered values — pure functions, no mocks, plus the
+ * profile page's contract that it actually renders them.
+ *
+ * The behaviour under test is a reader's question: is this number one the
+ * check just took, one it took earlier and could not confirm, or one it has
+ * never taken at all? The first two look identical on screen without a marker,
+ * and the third used to look like nothing at all.
+ */
+
+import { describe, it, expect, test } from "bun:test"
+import {
+  CACHED_TEXT,
+  NEVER_READ_TEXT,
+  cachedTag,
+  factProvenance,
+  renderFactValue,
+} from "../telemetry/cachedFacts"
+import { profilePageHtml } from "../telemetry/profilePage"
+
+describe("factProvenance", () => {
+  it("calls a value from the check that just ran live", () => {
+    expect(factProvenance("max", false)).toBe("live")
+  })
+
+  it("calls a value the failed check inherited cached", () => {
+    expect(factProvenance("max", true)).toBe("cached")
+  })
+
+  it("calls an absent value never, whether or not the check failed", () => {
+    expect(factProvenance(null, true)).toBe("never")
+    expect(factProvenance(undefined, false)).toBe("never")
+  })
+
+  it("treats an empty string as absent rather than as a live reading of nothing", () => {
+    expect(factProvenance("", false)).toBe("never")
+    expect(factProvenance("", true)).toBe("never")
+  })
+})
+
+describe("cachedTag", () => {
+  it("marks a cached value", () => {
+    expect(cachedTag("cached")).toContain(CACHED_TEXT)
+    expect(cachedTag("cached")).toContain("cached-tag")
+  })
+
+  it("leaves a live value unmarked", () => {
+    expect(cachedTag("live")).toBe("")
+  })
+
+  it("does not mark a value that was never read — there is nothing to date", () => {
+    expect(cachedTag("never")).toBe("")
+  })
+})
+
+describe("renderFactValue", () => {
+  it("renders a live value with no marker at all", () => {
+    const html = renderFactValue("alice@example.com", false)
+    expect(html).toBe('<span class="detail-value">alice@example.com</span>')
+    expect(html).not.toContain(CACHED_TEXT)
+  })
+
+  it("renders a cached value with the marker to the right of it", () => {
+    const html = renderFactValue("alice@example.com", true) ?? ""
+    expect(html).toContain("alice@example.com")
+    expect(html).toContain(CACHED_TEXT)
+    // "to the right of each value" — the marker follows the value, never leads it.
+    expect(html.indexOf("alice@example.com")).toBeLessThan(html.indexOf(CACHED_TEXT))
+  })
+
+  it("renders never-known differently from known-but-stale rather than collapsing both", () => {
+    const stale = renderFactValue("20x", true)
+    const never = renderFactValue(null, true)
+
+    expect(stale).not.toBe(never)
+    // The stale one still carries the figure; the never one must not invent it.
+    expect(stale).toContain("20x")
+    expect(never).not.toContain("20x")
+    expect(never).toContain(NEVER_READ_TEXT)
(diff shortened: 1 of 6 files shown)


rynfar:
Thank you @Nowaker! This has been incorporated and merged via #1071 with original commit authorship preserved, and extended to provide cached usage provenance and indicators in the desktop app and system tray.

PR #825, opened 2026-08-17
fix: coordinate SDK/session concurrency and add graceful shutdown (hardened)

rynfar:
Supersedes #821. The concurrency/shutdown work is **@debin-ge's** — their commit is cherry-picked here unmodified and they remain its git author, so credit shows up in the log and the contributor list. Our review corrections land as separate follow-up commits.

Rebased onto current `main` (#815 landed since #821 was opened). The auto-merge was verified, not just accepted: #815's `resolveCooldownUntil` refactor survives intact and the stale `PRIORITY_COOLDOWN_CAP_MS` is correctly gone.

## What the original does

Replaces the ad-hoc `activeSessions` counter and `AsyncLocalStorage` re-entrancy hack with an abortable FIFO semaphore plus a per-session turn coordinator, adds graceful draining/shutdown, and exposes session/SDK queue timing in telemetry. Calling `handleMessages` in-process instead of re-entering `app.fetch` is the correct fix for the `insideSessionSlot` deadlock.

## What we corrected

Each fix has a test that **fails without it** — verified by stashing the source changes and re-running.

**1. The drain contract was false for the OpenAI-compatible routes.**
`docs/configuration.md` promised `/v1/chat/completions` and `/v1/responses` return `503` + `x-meridian-draining: 1`. They didn't — the gate lived only on the internal `/v1/messages` hop, and both routes flattened every inner failure into `upstream_error`, discarding the header and disguising the new session-conflict `400` ("your history is stale, refetch it") as an upstream fault a gateway would retry or fail over instead of resolving. There was also a real race: both routes `await c.req.json()` and translate *before* the internal fetch, so a request accepted while healthy could be refused mid-translation.

Admission now happens at each public entrypoint. The internal hop is exempted by a **per-instance random token**, not a fixed header name, so the exemption cannot be claimed from the wire — there is a test for that too.

**2. Queue telemetry was wrong exactly when it mattered.**
`sdkQueueWaitMs` was credited only from a *granted* lease, so an aborted wait dumped its whole duration into `proxyOverheadMs` — corrupting the one metric that means "the proxy is the bottleneck" precisely under the load that makes clients cancel. Separately, a request cancelled while waiting for the session lease returned before `RequestMeta` existed and recorded **nothing at all**, making "clients give up waiting on the session lock" invisible — the exact symptom this feature can cause. It now files a 499 row.

**3. `forkAttemptMeta` rewrote a caller-supplied request id.**
It rewrote `x-request-id` to `<id>.1` per failover attempt, justified by a comment claiming same-id rows overwrite each other. They don't: SQLite always INSERTs against a non-unique `request_id` and the memory store is an append-only ring buffer. A client sending `x-request-id: abc` could not find the row that actually served it. The id now survives failover byte-identical; attempts stay distinguishable by `profileId` and status.

**4. Two robustness guards.**
A second SIGINT/SIGTERM now forces exit (`130`) instead of being swallowed for the full 30s window — a wedged stream previously made the process unkillable short of SIGKILL. And `closeAllConnections()` is called optionally, which the adjacent comment already assumed but the code did not do.

## Verification

- `npx tsc --noEmit` clean.
- Full suite: **2591 passed, 0 failed, 1 skipped** across 155 files. Baseline on the bare cherry-pick was 2585 across 153 — the delta is exactly the 6 new tests; nothing else moved.
- Each new test confirmed red on the unmodified original.

## Needs your call

- **`/health` is a Stable API Contract entry.** It gains a `status: "draining"` shape and a `503`, and `close()` now keeps the listener serving for up to 30s. `AGENTS.md` requires owner sign-off, so this is flagged rather than assumed.

## Deliberately deferred

- **The SDK permit is released before the subprocess is actually dead.** `sdkQuery.close()` is fire-and-forget and the SDK's process transport waits ~2s before SIGTERM (up to 5s more before SIGKILL), so `MERIDIAN_MAX_CONCURRENT` can be transiently exceeded on the abort path — when the host is already under stress. Worth its own PR.
- **Forced close awaits HTTP teardown, not request settlement**, then the CLI calls `process.exit(0)`; SDK cleanup is async and can be cut short.
- **The SQLite migration test builds the current schema**, so an omitted `ALTER` would not fail it. Both new ALTERs are correct; the test is what is weak.
- **The error-path telemetry row carries no `profileId`**, so "which account refused" is missing. Pre-existing on `main`.
- Nits: the turn lease key is not namespaced by adapter; `processSessionTurns` is module-global, so two instances in one process serialize each other's sessions with no per-instance opt-out.

One review claim we checked and **dismissed**: removing `pendingSessionStores` is not a regression — that wait was already gated on `profileSessionId`, so the lease strictly supersedes it.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -37,6 +37,8 @@ Agent (OpenCode) ◄── SSE Response ◄────────────�
 src/
 ├── proxy/
 │   ├── server.ts              ← HTTP layer: routes, SSE streaming, concurrency, request orchestration
+│   ├── concurrency.ts         ← Abortable SDK query semaphore and concurrency config parsing
+│   ├── shutdown.ts            ← Bounded HTTP drain and connection tracking
 │   ├── adapter.ts             ← AgentAdapter interface (extensibility point for multi-agent support)
 │   ├── adapters/
 │   │   ├── opencode.ts        ← OpenCode adapter (session headers, CWD extraction, tool config)
@@ -51,7 +53,8 @@ src/
 │   │   ├── index.ts           ← Barrel export
 │   │   ├── lineage.ts         ← Pure functions: hashing, lineage verification
 │   │   ├── fingerprint.ts     ← Conversation fingerprinting, client CWD extraction
-│   │   └── cache.ts           ← LRU session caches, lookup/store operations
+│   │   ├── cache.ts           ← LRU session caches, lookup/store operations
+│   │   └── turnCoordinator.ts ← Process-wide strict serialization for reliable session IDs
 │   ├── sessionStore.ts        ← Shared file store (cross-proxy session resume)
 │   ├── profiles.ts            ← Multi-profile support: resolve, list, switch auth contexts (leaf)
 │   ├── profileCli.ts          ← CLI commands for profile management (leaf, I/O)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -58,6 +58,8 @@ OpenCode-specific behavior is documented in `ARCHITECTURE.md` under "Agent-Speci
 
 ```
 server.ts          → HTTP routes, SSE streaming, concurrency (orchestration only)
+concurrency.ts     → Abortable SDK query semaphore, max-concurrency config
+shutdown.ts        → Bounded HTTP drain, socket tracking, forced close
 adapter.ts         → AgentAdapter interface (extensibility point)
 adapters/
   opencode.ts      → OpenCode-specific: headers, CWD, tool config
@@ -73,6 +75,7 @@ session/
   lineage.ts       → Hashing, lineage verification (PURE — no I/O)
   fingerprint.ts   → extractClientCwd, getConversationFingerprint
   cache.ts         → LRU caches, lookupSession, storeSession (stateful)
+  turnCoordinator.ts → Process-wide strict serialization for reliable session IDs
 ```
 
 ## Stable API Contract
@@ -90,7 +93,6 @@ External plugins depend on these interfaces. **Do not change without project own
 | `POST /v1/messages` request/response format | `server.ts` | All agents (Anthropic API contract) |
 | `GET /profiles/list` response shape | `server.ts` | Profile management UI and CLI |
 | `POST /profiles/active` request/response | `server.ts` | Profile switching from CLI and UI |
-
 If you need to modify any of these, open an issue first — breaking changes affect downstream plugin authors.
 
 ## Git & Workflow
--- a/bin/cli.ts
+++ b/bin/cli.ts
@@ -212,6 +212,26 @@ export async function runCli(
       process.exit(1)
     }
   })
+
+  // Graceful shutdown: close() itself drains in-flight requests (bounded by
+  // MERIDIAN_SHUTDOWN_GRACE_MS) before releasing the port — see close() in
+  // startProxyServer. This handler just owns the process-exit decision, since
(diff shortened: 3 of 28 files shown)


rynfar:
### CI caught a latent test-isolation bug the local run couldn't

The first CI run failed on a test this PR never touched:

```
(fail) SDK concurrency limiter > still bounds how many SDK calls run at once
Expected: <= 2
Received: 6
```

It passes locally, on `main`, and on the bare cherry-pick. The cause is real, not flake:

`getProcessSdkSemaphore()` memoises a process-wide singleton on first use. `proxy-concurrency-limiter.test.ts` requests its bound through the **env var** (`createTestApp("2")` sets `MERIDIAN_MAX_CONCURRENT` and lets `createProxyServer` fall through to the singleton), so it only ever measured its own limit by being the first file in the run to construct a proxy. Once any earlier suite constructed one, the singleton was already frozen at the default 10 and the "2" was silently ignored — leaving the burst bounded only by scheduling. Confirmed directly: after the cache is warm, `getProcessSdkSemaphore()` returns the *same instance* even though the env var changed.

So the assertion was decorative under any ordering but the lucky one, and this PR's new files changed the ordering. Fixed by resetting the singleton on both sides in both suites, so each gets a semaphore that reflects the limit it actually asked for.

Worth noting the production-side implication, which is unchanged and not a bug today: `MERIDIAN_MAX_CONCURRENT` is read once per process, so it cannot be changed at runtime. That's fine for the CLI, but it does mean the per-instance `maxConcurrent` config field is the only way to get a distinct limit — as `server.ts:507-509` already does.

All checks green: `test`, `build-push`, `windows-smoke`.

rynfar:
### `/health` — resolved, and it does not need a contract waiver

I flagged this as needing owner sign-off. Checking `main` shows that was over-cautious; the change is additive within the existing contract, on all three axes:

1. **The 503 is not new.** `main`'s `/health` already returns `503` for `status: "unhealthy"`. Any plugin polling this endpoint has always had to handle a 503.
2. **`status` was already a varying string** — `healthy` / `degraded` / `unhealthy`. `draining` is a fourth value in a field callers already had to switch on.
3. **The healthy `200` body is byte-identical.** The drain check is a pure prepend; the rest of the handler is untouched.

And the state it replaces was strictly worse: on `main`, `close()` calls `server.close()` immediately, so a health check during shutdown got a **closed socket**. A `503` with `status: "draining"` is a strict improvement over ECONNREFUSED — it's the standard readiness-gate behaviour, and it's what makes the drain window useful to a fleet manager at all.

Pinned in `graceful-shutdown.test.ts`: an undrained `/health` never reports `draining`, keeps `status`/`version`, and gains no new status code until `beginDrain()` is actually called.

### Also completed in this round

**The SQLite migration test was decorative.** It opened the DB through `createSqliteStores`, which creates *today's* schema — so every `ALTER TABLE` could have been deleted and it would still have passed. It now hand-builds the pre-migration table, asserts the new columns really are absent, then checks the back-filled defaults on the legacy row, the values on a fresh row, and idempotency on reopen. Verified honestly: deleting one `ALTER` leaves the old test green and fails the new one.

### Filed rather than rushed

- #827 — SDK permit released before the subprocess actually exits (can exceed `MERIDIAN_MAX_CONCURRENT` on the abort path)
- #828 — forced shutdown awaits HTTP teardown, not request settlement
- #829 — error-path telemetry rows don't record which profile refused (pre-existing)

Each needs either evidence this PR can't produce (#827) or scope this PR shouldn't take (#828, #829). Everything fixed here is covered by a test that fails without the fix.

All checks green.

PR #826, opened 2026-08-17
feat(routing): say when an account is refusing, and route around it

Nowaker:

## The problem

An account that has run out keeps being routed to, and keeps reporting itself healthy while it happens.

Measured on a 12-account fleet: one account rendered `5h 67% / 7d 7%` on `/profiles` while every request through it was being refused. Meridian's own `/telemetry/requests` recorded those requests as `status: 200` with `error: null`. The connected agent, seeing an error frame in a stream that had already returned 200, could not map it to a status, recorded it as a non-retryable 500, and stopped permanently. Sessions sat dead for 25 minutes with nothing anywhere in Meridian saying an account had run out.

The mechanism is that **a refusal is not an HTTP status**. By the time the SDK fails, the response has already committed `200` plus `text/event-stream` headers, so Anthropic's refusal can only arrive as an `event: error` frame inside the body. Anything watching status codes is structurally unable to see it. `sniffQuotaFailure` does read that frame, but it is called from exactly one place - `dispatchPriority` - so in `active` and `sticky` mode nothing inspected the stream at all.

### 1. A refused account now says so, in every routing mode

Refusal bookkeeping moved to the two terminal error paths (streaming and non-streaming), which is the only point where the account id, the raw CLI wording, and the failure are all in hand at once. It marks the profile exhausted, records what was refused, and publishes an event.

This is arguably the most valuable line of the change: it makes an exhausted account visible in **every** mode, not only in the one an operator may not have switched to yet. It also fixes a second defect found on the way - a failed request's telemetry row carried `profileId: null`, so even when a failure *was* recorded you could not tell which account it belonged to.

### 2. Meridian works out *which* allowance ran out, and says how confident it is

New leaf module `limitDetection.ts` with a four-rung evidence ladder, reported alongside the verdict:

| source | meaning |
|---|---|
| `error_message` | Anthropic named it. `"You've hit your session limit - resets 12:30am (America/Chicago)"` gives both the bucket and a real reset instant, resolved through the named zone (DST-correct). |
| `sdk_event` | The SDK's own `rate_limit_event` reported `rejected` for a specific window. |
| `cached_usage` | Nothing named it, so it is deduced from the last-read windows: if the weekly is nowhere near its limit, the 5-hour one is what ran out. **Flagged as a guess.** |
| `unknown` | Say so rather than invent a bucket. |

When the prose names a five-hour reset, that instant also becomes a new **tier 0** for the priority cooldown, ahead of the existing 10-minute conservative default. Deliberately restricted to `five_hour`: a weekly reset printed as a bare clock time says nothing about which *day* it lands on, and adopting it would suppress an account for hours on a misreading.

The verdict is exposed as a field **separate from** the cached usage numbers, never overwriting them. One is what we last read; the other is what the API is doing right now. An account can be at a cached 67% and refusing, and both facts are true.

### 3. `active+priority`: the operator steers, Meridian catches the fall

A fourth routing mode. The active profile is always the pool head - whatever a human or a supervisor selected - and the existing `profileOrder` supplies the fallbacks behind it. It reuses that setting as-is and does **not** get a list of its own.

**Design decision worth reviewing:** priority mode has assignment affinity - an existing conversation stays pinned to its account while healthy, protecting a warm prompt cache (a live request here showed a 441k-token cache read at 99.8% hit rate). `active+priority` deliberately does **not** honour affinity for the *head*. In this mode, moving the active profile is how you move traffic; if an old assignment outranked it, switching the active profile would not move running conversations, which is the entire capability that distinguishes this mode from plain `priority`. Affinity is still honoured among the fallbacks, so a conversation that already failed over is not dragged back and forth.

Failover is ad-hoc re-proxying **inside one request**: the request is re-issued internally against the next candidate and the client receives a normal answer, never an error. It only sees a failure if every candidate refused.

`getRoutingMode`'s typo-safety is preserved - the three original modes stay exact-match, so every value that falls back to `active` today still does. Only the new mode accepts separator variants, because its canonical spelling carries a `+`, which a query string decodes to a space.

### 4. A failover surface a supervisor can actually poll

`profile.failover` previously went to the log and nowhere else, so nothing outside the process could learn that an account had dropped out. `GET /profiles/events` is a bounded ring with a `since` cursor:

```
GET /profiles/events?since=<seq>&limit=<n>
```

`nextSince` is the last **returned** event's seq, not the newest held, so a page truncated by `limit` resumes exactly where it stopped. `dropped` reports that the ring evicted events the caller had not read. Both properties exist because `/telemetry/requests` has neither - no cursor, and a refusal was never recorded as a failure in the first place - so polling it can silently miss a failover between two polls.

### 5. The UI stops contradicting itself

A refusing account now carries a red badge and a full-width banner on `/profiles` and `/`, stating what the API is doing right now and labelling a deduced bucket as a guess. The cached percentage bars stay exactly as they were, under a note that they are the last successful read and not live. A cheerful `67%` next to an account that cannot serve was precisely the bug; showing the refusal *without* the history would have been a different one.

## Verification

- `bun run test`: **2604 pass / 0 fail**, exit 0 (65 new).
- `npx tsc --noEmit`: clean.
- Both pages driven in a real browser against a fixture reproducing the measured `5h 67% / 7d 7%`-while-refusing state: badge, banner, guess labelling, unaffected accounts left clean, bars intact, zero console errors.
- One defect was found by its own test and fixed before commit: a duplicated id in `profileOrder` produced a repeated candidate, which would have made the dispatcher attempt the same account twice, breaking its at-most-once-per-request rule.

## Not included

No durable state - refusal marks and events are in-memory routing hygiene, consistent with the existing exhaustion tracker. No change to the three existing modes. No new profile-order list.


rynfar:
Incorporated via #1075 with active+priority routing, limit detection, refusal events, and desktop app indicators. Thank you @Nowaker!

PR #833, opened 2026-08-17
feat: make a refused account visible, and say where each request went and which accounts refused it

Nowaker:

## What this makes answerable

With a pool of accounts, three questions decide everything and Meridian could answer none of them.

**"Is this account actually working?"** An account that has run out keeps reporting itself healthy. Measured on a live fleet: `/profiles` rendered `5h 67% / 7d 7%` for an account while every single request through it was being refused, and the sessions pinned to it sat in backoff for 25 minutes. Nothing was wrong with the reading — it was a snapshot of the last *successful* read, and a refusal never touched it.

**"Where did this request actually go, and why there?"** `/telemetry` could name the profile that served a request but never why it was that one, so "went to corp2 because corp2 is active" looked identical to "went to corp2 because corp4 refused it first". Worse, the second case did not render as one request at all: a failover re-enters the proxy once per candidate, each hop wrote its own telemetry row, and the page showed two unrelated-looking requests.

**"Which accounts are carrying the load, and which are refusing it?"** Per-request answers do not aggregate into that. And the number that mattered most did not exist anywhere: a refusal that a failover hid from the client never reaches the client as an error, so it is in no error count, no status column and no log line anyone reads.

After this PR: a refused account is visible in **every** routing mode, with which allowance it ran out of and when that reopens; `/telemetry` tags each request with **how** its account was chosen and shows the **whole chain** when it failed over (`corp4 ✗ → corp2 ✓`); an Accounts table says what each account served, refused, and refused *on*; a supervisor can poll a lossless event stream instead of guessing from a ring buffer; and a new `active+priority` mode lets a human keep control of where traffic goes while still failing over automatically.

## Why a refusal was invisible

A refusal is not an HTTP status. By the time the SDK fails, the response has already committed `200` plus SSE headers, so Anthropic's refusal can only reach the client as an `event: error` frame in the body. Nothing that watches status codes can see it — the telemetry ring recorded those requests as status 200 with `error: null`. The one place that did read the frame, `sniffQuotaFailure`, ran only inside `dispatchPriority`, so in `active` and `sticky` mode nobody looked at all.

That is the root cause behind every part of this PR, and it is why the fix has to live at the terminal error paths inside the request handler: that is the only point where the account id, the raw CLI wording and the failure are all in hand at once.

## What is in it

**Refusal bookkeeping in every routing mode.** At both terminal error paths, a rate-limit failure marks the profile, records what was refused, and publishes an event. An exhausted account becomes visible whether or not anything routed around it.

**`limitDetection.ts` — which allowance was refused.** An evidence ladder that reports which rung it used. Anthropic often names it outright (`You've hit your session limit · resets 12:30am (America/Chicago)`), in which case both the bucket and a real reset instant are read from the prose — and that reset becomes a new tier 0 for the priority cooldown, ahead of the old 10-minute guess. Failing that, the SDK's own `rate_limit_event` is consulted, then the last-read usage windows are used to deduce it, flagged as a guess. The verdict is reported *separately* from the cached percentages and never overwrites them: one is what we last read, the other is what the API is doing right now.

**`active+priority`, a fourth routing mode.** The active profile is always the pool head — whatever the human or a supervisor selected — with the existing `profileOrder` behind it as fallbacks. It deliberately does *not* honour priority mode's assignment affinity for the head: in this mode, moving the active profile is how you move traffic, so an old assignment must not outrank it. Affinity still applies to the fallbacks, so a conversation that failed over is not dragged back and forth across accounts.

**`GET /profiles/events`** exposes refusals and failovers to a poller with a `since` cursor and a `dropped` flag, so a supervisor polling every few seconds can prove it missed nothing. `/telemetry/requests` cannot do that: no cursor, and until this PR a refusal was never recorded as a failure anyway.

**Route attribution on `/telemetry`.** Every metric carries `routeKind` — `pinned` for an explicit `x-meridian-profile`, `active`, `sticky`, `priority`, `active+priority`, or one of the `-hop` kinds for a single internal attempt — plus `routeGroupId`/`routeAttempt` correlating the hops of one client request. `GET /telemetry/requests` folds a group into one row carrying `routeChain`, so a failover renders as one row naming every account it touched, with the failure reason and the refused allowance on hover. `?hops=1` returns the raw per-account attempts for anyone who wants them.

The two pool modes keep separate hop kinds rather than sharing one with a flag, because the collapsed row has to name the mode that produced it: `priority` drained the pool in configured order, `active+priority` put the selected account at its head, and those are different answers to "why this account" even when the chains look alike.

**An Accounts table and a Failovers card.** One row per account: served, refused, which allowances those refusals named, estimated cost, and whether it is refusing or benched right now. Both are suppressed for single-account setups, which learn nothing from either. Three sources joined by profile id — the route tally, `costEstimate.byProfile` (already computed, never rendered until now), and live health.

**`GET /telemetry/routes`** is that tally as JSON: `requests`, `failedOver`, `unserved`, `byKind`, and a per-profile `served` / `refused` / `refusedBuckets`.

**`GET /profiles/health`** is a deliberately separate endpoint rather than a field on `/profiles/list`, which awaits an auth status per profile. That check is cached 60s on success but only **5s on failure**, so a page polling `/profiles/list` on a timer would spawn a `claude auth status` subprocess per unhealthy account per poll — worst exactly when an account has gone bad, which is when somebody is watching. `/profiles/health` is an in-memory snapshot and does no I/O.

**The page now says what it remembers.** The window selector offers "Last 24 hours" against a ring buffer of 1000 rows — under an hour at a working pace, and less in pool routing, where every hop takes a slot. A line under the selector reports what is actually held (`1 of 1,000 requests in memory · back to 3m · lost on restart`, or `4,812 requests in SQLite · 30d retention · back to 6d · 101 KB · survives restart`), and any window reaching past the data is suffixed `— only 3m held`. Those options stay selectable: the data grows, a partial answer is still an answer, and a selector that refuses what a user asked for is worse than one that says what they will get. Backed by `GET /telemetry/retention` and a new `describe()` on `ITelemetryStore`.

Two smaller corrections came with it. A failed persistence setup now reports what actually failed instead of asserting libsql is missing — libsql is a regular dependency, so for every failure except that one (an unwritable `MERIDIAN_TELEMETRY_DB`, a directory that cannot be created, a file owned by another user) the old message sent people to reinstall software they already had. And the diagnostic log ring honours `MERIDIAN_DIAGNOSTIC_LOG_SIZE`: at 500 entries it was the tightest window in the system and the only one with no way to change it, so the logs a bug report needs had usually scrolled away before anyone looked.

## No work added to the request path

Nothing here computes, joins or formats anything for the dashboard while a request is in flight.

- The request path only *records* values already in hand: two header reads, one flag and a pure classifier. Measured at **219 ns/request** over 2M iterations, against a live `proxyOverheadMs` of 582-789 ms on the same box — roughly six orders of magnitude below existing overhead.
- `routeRefusedBucket` is a field copy from the diagnosis the refusal bookkeeping already produced, not a second detection.
- The hop correlation and the tally run entirely on the **read** path. `collapseRouteChains` and `summarizeRoutes` have **zero importers under `src/proxy/`** (`rg -l` finds only `telemetry/{routes,index,routeChain}.ts` and their tests), so they are unreachable from the request handler. Measured: **2.8 µs** to collapse 1000 rows with no failovers, **359 µs** in the pathological case where every request failed over three times.
- `describe()` is O(1) on the memory store (read off the ring head) and an indexed `MIN(timestamp)` on SQLite.
- The ring buffer's size and retention are unchanged.

## Design notes worth knowing before reading the diff

The hops are correlated through `RequestMeta`, not through headers: `dispatchPriority` re-enters `handleMessages` directly, so there is no second HTTP request for a header to travel on. `forkAttemptMeta` already carries `requestId` across every attempt, so that id groups the chain and `routeAttempt` says where in it a hop sits — and marks that it is one.

`RouteKind` is declared in `telemetry/types.ts` rather than `proxy/routing.ts` because it is part of the `RequestMetric` wire shape, and telemetry may not import from proxy — dependencies flow proxy → telemetry only.

The failed hop is what names the account that refused, and a failed hop records from the outer `catch`, which carried no profile id. That catch now attributes both of its consumers — the refusal bookkeeping and the telemetry row — from a profile id seeded with the dispatcher's forced pin, so a hop that fails early still books a refusal against the right account and a chain reads `corp4 ✗` rather than `default ✗`. The routing mode is captured the same way, so a request that failed is described by the mode in effect when it ran rather than whatever `/settings` holds by the time it lands.

The tally counts only `rate_limit_error` and `billing_error` as refusals — the same set the pool router fails over on. An `overloaded_error` says nothing about the account, and counting it would blame an account for an outage, which is the exact misattribution the table exists to prevent.

`describe()`'s oldest timestamp is the oldest row **still held**, not the oldest ever seen. Reporting the latter would tell the page it covers requests it can no longer show, which is the same class of lie as the 24-hour window itself.

The tier-0 cooldown is restricted to `five_hour` on purpose: a weekly reset printed as a bare clock time says nothing about which *day* it lands on, so adopting it would suppress an account for hours on the strength of a misreading. Under-suppressing is the safe direction.

## Verification

`bun run test`: **2882 pass, 0 fail, exit 0.**

Live-verified against real accounts on spare loopback ports, with the first pool candidate pointed at a local mock that refuses with the CLI's own session-limit wording, and rendered in a real browser (zero console errors):

| surface | what was observed |
|---|---|
| `active` | unpinned request → `profileId=personal, routeKind=active` |
| `pinned` | `x-meridian-profile: personal` → `routeKind=pinned` |
| `active+priority` | active account refuses → collapsed row `routeChain=[{mock-limited, ok:false, 429, rate_limit_error, refusedBucket:"five_hour"}, {personal, ok:true, 200}]`; `?hops=1` shows both attempts sharing `routeGroupId` at attempts 1 and 2 |
| `GET /telemetry/routes` | `requests 2, failedOver 1, unserved 0`, `byProfile {personal: served 2, mock-limited: refused 1, refusedBuckets {five_hour: 1}}` |
| `GET /profiles/events` | `refused` (internalHop: true, bucket `five_hour`) then `failover` (mock-limited → personal) |
| Route column | `mock-limited ✗ → personal ✓` with badge `active+priority` |
| Accounts table | `mock-limited [active] │ 0 │ 1 │ five_hour ×1 │ $0.00 │ refusing (five_hour) / back in 6h` |
| Failovers card | `1` / `50.0% of 2 client requests` |
| Retention line | memory: `1 of 1,000 requests in memory · back to 1m · lost on restart`; SQLite: `1 request in SQLite · 30d retention · back to 3m · 101 KB · survives restart` |
| Window annotation | all four options suffixed `— only 1m held`, none disabled |

The client received a normal `200` throughout the failover — which is exactly the case where, before this, nothing anywhere recorded that an account had dropped out.


rynfar:
Merged into main via maintainer PR #1076 with full test validation, active-priority routing integration, and desktop app parity (failover chain visualization, route kind badges, failover overview card, and per-profile refusal counts). Thank you @Nowaker for this great telemetry and attribution work!

PR #834, opened 2026-08-17
feat(proxy): fail back on the next user turn

WarGloom:
## Summary

- add configurable `new-conversation` and `next-user-turn` priority failback policies
- promote an assigned fallback session to the preferred profile only on a genuine new human turn
- start a fresh preferred backing session while preserving fallback affinity, cooldowns, and compare-and-set assignment safety
- persist and document the new setting while using upstream session-turn coordination for keyed sessions

## Testing

- `./build.sh` from the four-repo `local/dev-stack` workspace: passed
- focused routing/settings/auth tests: 114 passed, 0 failed
- `npm run typecheck`: passed
- `npm run build`: passed
- `npm test`: 2614 passed, 5 failed; the same five unrelated SDK/system-prompt tests fail on clean `origin/main` (2593 passed, 5 failed)
- `git diff --check`: passed

## Integration note

`next-user-turn` uses OpenCode request ID/kind headers when available. Clients without those headers retain the default `new-conversation` behavior. A companion opencode-with-claude PR adds the header producer.


rynfar (changes requested):
The opt-in feature goal remains relevant, but this implementation cannot be safely rebased onto Meridian 1.64.0. It needs a durable redesign; do not merge the current head.

Blocking issues:

1. `{ profileId, requestId }` lives only in the process-local `AssignmentStore` map. Restart, LRU eviction, or another proxy loses the last human request ID and can move a same-turn tool continuation back to the preferred account. That contradicts the feature's core contract and the cross-process model added by #879.
2. `freshSession` blindly calls `evictSession()` before lookup and ignores its result. Current durable resume requires exact-generation invalidation, retained fences on uncertainty, journaled fresh-target preparation, and supported isolated forks. A mechanical port can retire a newer mapping or resume stale preferred history.
3. The dispatch rewrite omits the current `turnWatchdogSignal`, regressing cancellation/shutdown fencing for blocked candidate attempts.
4. Neither bundled OpenCode V1 nor exact-beta V2 produces the required trusted request-kind contract. V1 lacks the kind header; V2 lacks both headers. Depending on a separate external plugin makes the documented bundled setup ineffective.
5. OpenCode header parsing and adapter-name checks belong behind an adapter method, not directly in `server.ts`. These are control headers that must be bounded and overwritten/removed at the final plugin boundary so callers cannot spoof a destructive promotion.

Please supersede this with a new current-main PR rather than resolving conflicts in place. The new design needs a durable exact-CAS routing record coordinated with #879 publication/fences, preservation of watchdog signals, bundled V1/V2 turn metadata, and cross-process/restart tests. Also cover stream/non-stream promotion, pre/post-content failure, cancellation/shutdown/crash recovery, undo/compaction/tool loops, hidden V2 agents, explicit profile pins, other adapters, full Windows/Docker/Bun-Nix gates, and real supported V1/V2 clients. The Aug-17 green checks predate #879/#880 and are not evidence for this architecture.

rynfar:
Superseded by #897, which preserves all six of this PR's commits with Nikita Bige as author and carries explicit co-author credit on the merged commit.

The replacement added the production-safe durable route/session publication and crash-safe attempt-claim architecture, passed the full local and hosted Linux, Windows, Docker, Bun/Nix, Claude-availability, and live OpenCode V1/V2 gates, and merged as `2e51fdb8e17eed5ff0e26ed0f926de449fe8106e`.

Closing only now that the replacement has merged. Thank you, @WarGloom, for the original design and contribution.

PR #835, opened 2026-08-17
fix(errors): classify exhausted usage credits

WarGloom:
## Summary

- classify Claude Code's `You're out of usage credits` result as `rate_limit_error`
- return the existing 429 error shape instead of a generic API error
- allow priority routing to retry the next configured profile for this live-observed wording
- add direct classifier and HTTP failover regressions

## Testing

- `./build.sh` from the four-repo `local/dev-stack` workspace: passed
- focused classifier and priority-routing tests: 146 passed, 0 failed
- `npm run typecheck`: passed
- `npm run build`: passed
- `npm test`: 2595 passed, 5 failed; the same five unrelated SDK/system-prompt tests fail on clean `origin/main` (2593 passed, 5 failed)
- `git diff --check`: passed

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -611,6 +611,12 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     expect(r.status).toBe(429)
   })
 
+  it("maps the CLI's 'You're out of usage credits' to rate_limit_error", () => {
+    const r = classifyError("Claude Code returned an error result: You're out of usage credits. /model to switch models.")
+    expect(r.type).toBe("rate_limit_error")
+    expect(r.status).toBe(429)
+  })
+
   // #764 and #787 were the same bug twice: a new qualifier, a 500 instead of
   // failover, a PR. These pin the shape so the next variant is already covered.
   it.each([
--- a/src/__tests__/priority-routing-integration.test.ts
+++ b/src/__tests__/priority-routing-integration.test.ts
@@ -219,6 +219,16 @@ describe("priority routing", () => {
     expect(body.content[0]?.text).toContain("prof-personal")
   }, 20_000)
 
+  it("fails over on the CLI's 'You're out of usage credits' wording", async () => {
+    failureMessage = "Claude Code returned an error result: You're out of usage credits. /model to switch models."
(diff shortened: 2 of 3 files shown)


rynfar (changes requested):
The canonical Claude Code banner is still misclassified as `500 api_error` on 1.64.0, so this fix remains relevant. The current branch cannot merge unchanged.

Required changes:

1. Rebase onto current main and preserve the hardened `HTTP_429.test(lower)` logic. This branch conflicts with that later false-positive fix and is 23 commits behind.
2. Replace bare `lower.includes("out of usage credits")` with a canonical banner/prefix-shaped predicate. Incidental or negated prose containing the tail must not become a 429. In priority mode, a false positive marks an account exhausted and spends the rest of the pool, so this is a routing correctness issue—not only an HTTP status issue.
3. Add negative classifier and HTTP priority tests proving incidental/negated text does not return 429, mark the profile exhausted, or try another profile. Keep the positive direct and failover regressions.
4. Run fresh current-main full tests, typecheck, build, Windows, Docker, and Bun/Nix checks. Existing checks predate #879/#880.
5. Main requires signed commits and both current commits are unsigned. Update with verified commits, or allow a signed maintainer replacement that explicitly preserves WarGloom's authorship.

The fix belongs in the pure `errors.ts` leaf and should not require an SDK upgrade or adapter/session changes. Useful additional coverage is pre-content SSE failover, an assigned durable continuation on the fallback, and the all-profiles-exhausted final 429 payload. Please retain the contribution, but harden and rebase it before merge.

rynfar:
Superseded by #890, merged as `099741dd7a341daa767b08e20649456c51675b62` after fresh exact-head Linux, Windows, Docker, Bun/Nix, and Claude-availability checks passed.

Your two implementation/test commits were preserved on the replacement branch with `Nikita Bige <wargloom@gmail.com>` as Git author. The signed squash commit also carries:

`Co-authored-by: Nikita Bige <wargloom@gmail.com>`

The replacement retains the intended account failover while narrowing classification to the complete live usage-credit banner, so incidental documentation, MCP errors, negations, filenames, quotations, and arbitrary suffix text cannot exhaust a priority profile. Thank you for the original fix and tests.

PR #841, opened 2026-08-18
feat: rename a profile from the CLI and the web UI

Nowaker:

Adds renaming a profile, from the CLI and from the web UI, with the old name
left behind as a redirect so nothing breaks mid-flight.

## What you get

- `meridian profile rename <old> <new>` in the CLI.
- A pencil control to the right of the name on each `/profiles` card, which
  renames in place (Enter to save, Esc to cancel).
- `POST /profiles/rename` behind the same auth as the other profile routes.
- `x-meridian-profile: <old>` keeps routing, to the **new** profile, until the
  old name is claimed again.

## The rename moves everything that names a profile

Its entry in `profiles.json` is the obvious half. The other is the credential
directory at `~/.config/meridian/profiles/<id>` - browser-login profiles point
at it explicitly via `claudeConfigDir`, and oauth-token profiles are pinned to
it for SDK isolation. Renaming the entry alone would leave a profile naming a
directory that is no longer its own, which logs that account out over a
cosmetic change.

So the two are planned together and applied as one step: credentials move
first (a directory rename within one filesystem, atomic), and if
`profiles.json` cannot then be written, the move is undone. A profile that
imported `~/.claude` is left alone - those credentials were never ours to
rename.

## The alias is a redirect, not a second name

After `x` is renamed to `y`, `y` is the profile and `x` is a pointer that
exists only so in-flight callers and stale configs keep working.

- **Resolution checks real profiles first, aliases second.** A stale alias can
  never shadow a live account that has since taken the name.
- **Claiming the name back drops the redirect.** `meridian profile add x`, or
  renaming anything else to `x`, reclaims it at that moment.
- **Chains are collapsed on write, not followed on read.** Renaming x -> y and
  later y -> z leaves z holding both `x` and `y`, so resolution stays a single
  lookup and no alias can dangle at an intermediate name that no longer exists.

Former names are visible in `meridian profile list` and on the profile card - a
redirect nobody can see is a trap.

## Refusals change nothing

An unknown source, an occupied target, a name outside the charset
`profile add` already enforces (letters, numbers, hyphens, underscores), and a
destination credential directory that already exists are each rejected before
any write happens.

The active-profile pointer follows the rename, on disk and in the running
process, so a fleet is never left pointing at a name that is gone. An instance
started with `MERIDIAN_CREDENTIALS_READONLY=1` refuses on both surfaces, since
a rename writes credentials.

## Shape

The planner is pure and lives in its own leaf module (`src/proxy/profileRename.ts`),
so the CLI and the HTTP route share one set of rules rather than two
implementations that drift. `planProfileRename` decides; `applyProfileRename`
is the only thing that touches disk.

## Testing

`bun run test`: **2663 pass, 0 fail** across 156 files in the main suite, plus
all 10 sequential suites green (3, 12, 4, 4, 3, 12, 10, 11, 8, 79 - each 0
fail). 32 new unit tests cover chain collapse, reclaim dropping the alias,
aliases never shadowing a real profile, the credential-dir move, the rollback,
and every refusal; 4 new integration tests cover the route.

Manually exercised end to end against a throwaway `HOME`, never the developer's
real profiles: the CLI for all six behaviours above, and the pencil in a real
browser for the happy path, both refusal messages, Esc, poll suppression while
editing, and the header chip following a rename of the active profile.

### DIFF
--- a/bin/cli.ts
+++ b/bin/cli.ts
@@ -46,7 +46,7 @@ See https://github.com/rynfar/meridian for full documentation.`)
 }
 
 if (args[0] === "profile") {
-  const { profileAdd, profileAddOauthToken, profileList, profileRemove, profileSwitch, profileLogin, profileHelp } = await import("../src/proxy/profileCli")
+  const { profileAdd, profileAddOauthToken, profileList, profileRemove, profileRename, profileSwitch, profileLogin, profileHelp } = await import("../src/proxy/profileCli")
   const subcommand = args[1]
   const profileId = args[2]
   const headless = args.includes("--headless")
@@ -62,6 +62,7 @@ if (args[0] === "profile") {
   }
   else if (subcommand === "list" || subcommand === "ls") profileList()
   else if (subcommand === "remove" && profileId) profileRemove(profileId)
+  else if (subcommand === "rename" && profileId && args[3]) profileRename(profileId, args[3])
   else if (subcommand === "switch" && profileId) await profileSwitch(profileId)
   else if (subcommand === "login" && profileId) await profileLogin(profileId, { headless })
   else profileHelp()
--- a/src/__tests__/profile-rename.test.ts
+++ b/src/__tests__/profile-rename.test.ts
@@ -1,0 +1,362 @@
+/**
+ * Unit tests for profileRename.ts and the alias half of profiles.ts.
+ *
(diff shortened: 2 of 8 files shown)


rynfar:
Merged into main via maintainer PR #1079 with full test validation and desktop app parity. Thank you @Nowaker!

PR #843, opened 2026-08-18
fix(plugin): route hidden opencode agents on a distinct session key

dmitryrn:
# Why
This error started happening to me in opencode:
```
AI_APICallError: This session advanced while the request was waiting.
Retry with the latest conversation history or use a distinct session ID.
```
on meridian's npm version 1.62.5.

opencode-ai@1.18.18

# Why error occurred (regression)
- Aug 11 (be7c3ab): Latest npm version that day was 1.61.0 - worked on this version for me.
- Aug 16 (83b2f5c, PR #825 "fix: coordinate SDK/session concurrency and add graceful shutdown"): introduced the turn coordinator and the "session advanced while the request was waiting" error string. Shipped in 1.62.2. - this seems to have introduced this error for me

1.62.5's commit (#839, 91aeaf5) tried to fix this class of issue, it made declaresConcurrentFlow true when x-opencode-agent-mode: subagent is present, which exempts a request from the refusal. But the exemption applies to the subagent request (title), which is the lease holder, build is the one refused, and build is primary, so it isn't exempted. This is the gap this PR plugin fix closes (by routing title-gen on a distinct session key so it serializes against itself, not against build).

# Explanation
title/summary/compaction run concurrently with the primary build turn under the same opencode sessionID, so meridian's turn coordinator refuses the queued request as "session advanced while waiting". Override the built-in affinity headers with a distinct key for these one-shot agents so they're treated as independent and can't collide with build.

For context, you have this: https://github.com/rynfar/meridian/blob/main/docs/configuration.md#concurrent-requests-to-the-same-session, but the plugin was sending the same session key for these agents as for build, so they serialized against each other

# Tested
I tested in opencode, the error is gone

### DIFF
--- a/plugin/meridian.ts
+++ b/plugin/meridian.ts
@@ -50,6 +50,15 @@ const BUILTIN_AGENT_MODES: Record<string, string> = {
   compaction: "subagent",
 }
 
+/**
+ * Single-turn hidden agents (title/summary/compaction) run concurrently
+ * with the primary build turn under the SAME opencode sessionID, which
+ * meridian's turn coordinator refuses as "session advanced while waiting".
+ * They are one-shot, so route them on a distinct key instead of the shared
+ * one — meridian then treats them as independent and they can't collide.
+ */
+const HEADERLESS_AGENTS = new Set(["title", "summary", "compaction"])
+
 const MeridianPlugin: Plugin = async () => {
   // Modes from the merged config, per plugin instance. Replaced wholesale on
   // every config-hook fire so a reload can't leave stale entries behind.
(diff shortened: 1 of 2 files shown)


dmitryrn:
actually my clanker says dropping header is better, let me do it

dmitryrn:
actually no, this broke it

rynfar:
This is a correct fix, and I'm closing it only because the same defect shipped proxy-side in **v1.62.6** (#848) before I saw this PR. That's on me — I went from the symptom to the code without checking open PRs first.

Your mechanism genuinely works, and I verified it rather than assuming: overriding `x-session-affinity` (and `X-Session-Id`) reaches the wire because OpenCode spreads plugin headers *last*, so on the pre-1.62.6 adapter your change yields `ses_X:headerless:title` ≠ `ses_X` — distinct lease, distinct lineage, distinct priority-routing assignment.

What stands out is your commit trail: `44e6e5b` deleted the header, then `ac2b9a0` reverted it after you observed the collision come back. That observation is the crux of this whole area — a plugin cannot detach a request by *declining* to set a session header, because OpenCode always contributes `x-session-affinity` itself. PR #845 attempted exactly the delete-only approach and shipped green CI on a no-op. Your reverting commit is the reason I could tell the two apart quickly. (Small correction to the explanation there: the binary builds one object literal with the plugin spread last, so nothing happens "after" the hook — but the empirical conclusion is right, which is what mattered.)

Why the proxy-side fix is the one I kept, rather than a judgement about quality:

- It reaches every user on `npm i -g`, where a plugin fix only reaches those who re-run `meridian setup`.
- It reads `x-opencode-agent-mode` / `x-opencode-agent-name` — Meridian's *own* header names, set by Meridian's own plugin — so an upstream OpenCode header rename cannot break it. This PR depends on OpenCode-owned surface at two points (`x-session-affinity` being the name, and plugin headers landing last).

Had it gone in first, three things I'd have asked for, recorded in case they're useful elsewhere: gate on `isSubagent` as well as the name (a user-defined *primary* agent called `title` currently gets pushed onto the alternate key, so switching agents mid-session loses the lineage); drop `X-Session-Id`, which nothing on the OpenCode path reads; and rename `:headerless:`, since "headerless" already means "no session key at all" in this codebase and the misnomer ends up on the wire.

Your test also genuinely discriminates the change — `x-session-affinity` is never set by the plugin on `main`, so it fails without your diff. Worth knowing that this PR has run zero CI checks, so "no red" wasn't evidence either way; #852 reseeds that harness with OpenCode's real base headers so the whole class of "the plugin omitted a header it doesn't own" is now catchable.

For transparency: this couldn't have been merged as-is regardless — `main` requires signed commits, which isn't resolvable from a fork. Not a reflection on the work, and thank you for it.

PR #845, opened 2026-08-18
fix(plugin): detach title/summary requests from the session turn lease

justprosh:
## Problem

OpenCode's `title`/`summary` utilities run inside the parent session and fire in parallel with the user's real turn. The plugin tagged them with the same `x-opencode-session`, so both requests raced for the per-session turn lease — the loser died with `400 session_turn_conflict`. In practice the loser was usually the user's first turn, and the title exchange got committed into the session lineage.

## Fix

`chat.headers` now sends these one-shots:

- **without `x-opencode-session`** — no shared turn lease, nothing to conflict with;
- **with `x-meridian-source: subagent-<name>`** — an explicitly parallel stream, so the fingerprint fallback cannot glue them back onto the session (the title prompt embeds the conversation's own first message).

`compaction` keeps its session id: the proxy needs it for lineage and already exempts compaction from the conflict check.

## Tests

New describe block in `plugin-agent-mode.test.ts`: `title`/`summary` (string and legacy object shapes) go out detached, `build`/`compaction`/`general`/`explore` keep the session header.

## Note

Under sticky routing these one-shots lose profile affinity — they are cheap one-offs, any profile can serve them.

### DIFF
--- a/plugin/meridian.ts
+++ b/plugin/meridian.ts
@@ -50,6 +50,16 @@ const BUILTIN_AGENT_MODES: Record<string, string> = {
   compaction: "subagent",
 }
 
+/**
+ * One-shot utilities OpenCode runs inside the PARENT session instead of a
+ * child one (unlike general/explore, which carry their own session id), and
+ * concurrently with the user's turn. They are detached from the session in
+ * chat.headers — see the comment there.
+ * compaction is deliberately absent: the proxy excludes it from turn conflict
+ * itself and needs its session id to attach the compaction to the lineage.
+ */
+const PARENT_SESSION_ONE_SHOTS = new Set(["title", "summary"])
+
 const MeridianPlugin: Plugin = async () => {
   // Modes from the merged config, per plugin instance. Replaced wholesale on
   // every config-hook fire so a reload can't leave stale entries behind.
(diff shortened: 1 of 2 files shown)


rynfar:
Thank you for this — you found the right defect. OpenCode's `title`/`summary` agents running inside the parent session, racing the real turn for the turn lease, is exactly what was happening, and your diagnosis of the 400 is correct.

I have to close it, though, because the mechanism doesn't reach it: **omitting `x-opencode-session` does not make the request unkeyed.**

OpenCode's own header assembly always contributes `x-session-affinity` on the anthropic path (extracted from the 1.18.11 binary; plugin headers are spread *last*, so they can override that key but a plugin cannot remove it by declining to set one). And `openCodeAdapter.getSessionId` reads `x-opencode-session ?? x-session-affinity`. So the title request still resolves to the user's raw session id: same lease, same lineage, collision intact.

The `x-meridian-source: subagent-title` half is inoperative for the same reason — the independence guard it targets is `!agentSessionId && (fork-|isSubagent)`, and `agentSessionId` is truthy. The one thing that header does reach, `declaresConcurrentFlow`, was already true for these agents via the `x-opencode-agent-mode: subagent` the plugin has always sent.

**Your tests passed because the harness couldn't see it.** `headersFor` seeded `output.headers` with `{}`, modelling the plugin's contribution in isolation and never OpenCode's base object — so `expect(h["x-opencode-session"]).toBeUndefined()` read as "detached" but meant only "the plugin stopped setting one of three session headers". That's a harness defect, not yours, and it's fixed in #852: reseeded with OpenCode's real base headers, and asserting on the session key the *proxy derives* rather than on any single header. Against your change on the proxy version it targeted, the reseeded suite reports `Expected: 4, Received: 1` — all four agents on one key.

One thing worth flagging for any future pass: `PARENT_SESSION_ONE_SHOTS` matches on name only, ignoring mode. A user who defines their own *primary* agent named `title` would get `x-meridian-source: subagent-title` alongside `x-opencode-agent-mode: primary`; the proxy resolves that contradiction toward subagent, which silently drops them from `opus[1m]` to the 200k base tier.

The defect itself is fixed in **v1.62.6** (#848), proxy-side: `getSessionId` now scopes non-primary agents by agent name, reading the headers the plugin already sends. Verified live — first turn goes from a 400 after a 9.8 s stall to 200 with a 0 ms wait.

Also, for transparency: this PR could not have been merged as-is regardless, because `main` requires signed commits and that isn't something you can resolve from a fork. Not a reflection on the work.

PR #846, opened 2026-08-19
feat(plugin): add the OpenCode v2 plugin on session.model.request

justprosh:
Follow-up to #845, for the **opencode v2** line.

## Problem

opencode v2 has no `chat.headers` hook at all, so `plugin/meridian.ts` (v1) is inert on a v2 client — no agent tier, and nothing standing between v2 core and the proxy. Worse, v2's core (`SessionModelHeaders.make`) stamps `x-opencode-session` — plus `x-session-affinity` / `X-Session-Id` — on **every** model request, including the `title`/`summary` one-shots, which run inside the parent session and fire in parallel with the user's real turn. Both requests then carry the same session id, contend for the per-session turn lease, and the loser — in practice the user's first turn — dies with `400 session_turn_conflict` (`This session advanced while the request was waiting`).

Field-observed on a stock v2 beta client: refused 1-message requests with multi-second session-queue waits, and the title exchange committed into the session lineage. Reproduced against 1.62.3 and current `main`.

## Fix

New `plugin/meridian-v2.ts`, installed the same way (path in `opencode.json`'s `plugin` array). It hooks v2's `session.model.request`, which fires for the one-shots too and may rewrite headers in place:

- **title/summary** — session affinity headers stripped, `x-meridian-source: subagent-<name>` added: an explicitly parallel stream, which the proxy neither queues on the session's turn lease nor glues back via the lineage fingerprint fallback (the title prompt embeds the conversation's own first message);
- **every request** — `x-opencode-agent-name` / `x-opencode-agent-mode` for tier selection, resolving agent modes from the runtime with the built-in table as fallback (`compaction` keeps its session id: the proxy needs it for lineage and already exempts it from the conflict check).

## Tests

`src/__tests__/plugin-v2-headers.test.ts`: one-shots go out detached (all four affinity spellings gone, source header set), ordinary agents keep their session header, mode resolution from runtime/config/fallback, other providers untouched.

## Note for fronted deployments

A fronting proxy that owns the `x-meridian-*` namespace upstream (stripping it from client input as a control-channel boundary) must make this declaration itself — the plugin fixes direct v2 clients.

### DIFF
--- a/plugin/meridian-v2.ts
+++ b/plugin/meridian-v2.ts
@@ -1,0 +1,112 @@
+/**
+ * Meridian OpenCode plugin — **v2 line** (`@opencode-ai/cli@beta`, `opencode2`).
+ *
+ * The v1 plugin next to this file hooks `chat.headers`. That hook does not
+ * exist in opencode v2 at all, so on a v2 client the v1 plugin is inert: no
+ * agent tier, and — worse — no chance to fix what v2's core does on its own.
+ *
+ * What v2's core does: `SessionModelHeaders.make` stamps `x-opencode-session`
+ * (plus `x-session-affinity` / `X-Session-Id`) on **every** model request,
+ * including the `title` and `summary` one-shots that run inside the parent
+ * session and fire in parallel with the user's real turn. Both requests then
+ * carry the same session id, both contend for the proxy's per-session turn
+ * lease, and the loser — in practice the user's first turn — dies with
+ * `400 session_turn_conflict` ("This session advanced while the request was
+ * waiting"). Same failure the v1 fix addressed; v2 needs its own hook.
+ *
+ * This plugin uses v2's `session.model.request` hook, which fires for the
+ * one-shots too and may rewrite headers before the request goes out:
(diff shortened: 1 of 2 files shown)


rynfar:
Thank you, @justprosh — this work is now fully incorporated and shipped to `main` through #880.

Both commits from this PR were carried over patch-identically before the production hardening:

- `f669f576` → `5f57995c` (stable patch ID `ab76e4ed235a47926781f35f18698db46da446b5`)
- `4fdc5d4a` → `f7def95a` (stable patch ID `ed7e506476a0e318b2ce2df9e6b09a763038dbe8`)

#880 then added the exact beta-18314 pin, provider-scoped model/HTTP enforcement, compaction and subagent handling, V1/V2 setup detection, JSONC/duplicate protection, Windows support, packaging, and the full real-client validation matrix.

Your authorship is preserved on the carried commits and in the squash merge commit `40a69c7e` with an explicit `Co-authored-by` trailer. There is no remaining work unique to this PR, so I am closing it as superseded by #880.

PR #849, opened 2026-08-19
feat: configure telemetry storage from the web UI

Nowaker:

> **This PR stacks on #833.** GitHub cannot base a cross-fork PR on a branch that exists only in my fork, so this one targets `main` and therefore shows #833's commits underneath. **Only the final commit, `feat(settings): configure telemetry storage from the web UI`, is new here.** Read #833 first; if it lands, this rebases to a single commit.

## What this makes possible

You can turn on telemetry persistence, and set how long it is kept, without knowing an environment variable exists.

SQLite persistence has been fully implemented for some time — schema, migrations, retention, cleanup, and `libsql` as a regular dependency — and no page anywhere says so. The usual way to find out is to notice that `/telemetry` forgets everything on restart and never learn that it did not have to. `/settings` now has a **Telemetry Storage** card: persist on or off, retention in days, and both in-memory ring sizes.

## The part that makes this card different from every other setting on the page

These take effect **on the next start, not on the running proxy**. The stores are built once at startup, and swapping one out from under in-flight writes is not something a settings form should be able to do.

That makes the obvious implementation wrong. A form that saves a value and reads it back is claiming a change that has not happened — you would tick "persist", the page would show "persist", and the proxy would still be dropping everything on restart. So `GET /settings/api/telemetry` reports **three** states rather than one:

| | |
|---|---|
| `saved` | what is in `settings.json` — what the form shows |
| `wanted` | what a start right now would use, env precedence applied |
| `effective` | what the running stores actually are |

Every row renders its control next to the live value (`now: memory`), `pendingRestart` names the keys where `wanted` and `effective` disagree, and the card says so in as many words:

```
Persist to SQLite  [x]   now: memory
Retention          [30]  now: n/a while in memory
Request ring size  [  ]  now: 1,000
Diagnostic log     [  ]  now: 500

Holding 0 requests in memory — lost on restart
Saved, but not in effect: telemetryPersist. Restart Meridian to apply.
```

## It also says *how* to restart

"Restart to apply" on its own is a puzzle. The unit is named by whoever installed it, `systemctl restart` without `--user` silently addresses a different manager, and a proxy started by hand has no unit at all. All of that is already written in `/proc/self/cgroup`, so `proxy/supervision.ts` reads it and the card prints the exact command — `systemctl --user restart meridian-dev.service`, `sudo systemctl restart meridian.service`, or the honest fallback "Restart Meridian to apply." when nothing can be named.

A `.scope` leaf is deliberately **not** treated as a unit: a scope is how systemd accounts for a process it did not launch, so a proxy started in a terminal lands in one, and printing `systemctl restart session-2.scope` would send someone chasing a command that fails.

## Two fixes that fall out of the same work

**The default database path now follows `MERIDIAN_CONFIG_DIR`**, like every other file in that directory. It did not, and that is a collision rather than an inconsistency: two instances given separate config dirs *precisely so they do not share state* were both handed `~/.config/meridian/telemetry.db`, so one dashboard counted the other's requests. An explicit `MERIDIAN_TELEMETRY_DB` still wins. This PR makes persistence discoverable, so it also makes that collision much more likely to be hit.

**Persistence is resolved without `envBool`**, which cannot tell "unset" from `=0`. A checkbox in a browser must not be able to override an operator's explicit `MERIDIAN_TELEMETRY_PERSIST=0` in a unit file, and `envBool` reports both as plain `false`.

## Module layout

`settings.ts` moves from `src/proxy/` to `src/`, otherwise unchanged. It was already a leaf with no proxy dependencies, and telemetry has to read it to resolve these values; leaving it under `proxy/` would have made telemetry import from proxy, which `ARCHITECTURE.md` forbids in that direction (dependencies flow proxy → telemetry only). Four importers updated, no behaviour change.

The numeric bounds live beside the setting type rather than in the route, so the form and the validator read the same numbers — a page that offers a value the API rejects is a worse bug than either bound being slightly wrong.

## Verification

`bun run test`: **2906 pass, 0 fail, exit 0.**

**Supervision detection, against real cgroups rather than fixtures.** Run under a genuine transient unit:

```
$ systemd-run --user --unit=meridian-supervision-probe2 --pipe bun run probe.ts
{"kind":"systemd-user","unit":"meridian-supervision-probe2.service",
 "restartCommand":"systemctl --user restart meridian-supervision-probe2.service"}

$ bun run probe.ts          # same binary, started by hand
{"kind":"unknown","unit":null,"restartCommand":null}
```

**The card, in a real browser** on a throwaway config dir:

| check | result |
|---|---|
| initial render | four rows, `now:` values `memory` / `n/a while in memory` / `1,000` / `500`; note `Holding 0 requests in memory — lost on restart` |
| no premature warning | `Saved, but not in effect` absent |
| **toggle persist** | checkbox checked, **`now:` still reads `memory`** — the running store is correctly reported as unchanged |
| pending line | `Saved, but not in effect: telemetryPersist. Restart Meridian to apply.` — fallback wording, no `systemctl`, correct for a hand-started proxy |
| numeric round-trip | retention `30` survives the re-render |
| out-of-range | alert `telemetrySize must be an integer between 10 and 1000000, or null to unset.`; field reverts to empty, the rejected value is not shown as saved |
| console | one error, and it is the deliberate `400` from the out-of-range check |

The third row is the assertion that matters: ticking the box must not make the page claim persistence is live. It does not.


rynfar:
Consolidated and merged into main via #1080.

Issue #547, opened 2026-06-22
bug: Many context leaks


alon-z:
Hey, I have been using the proxy for a few weeks now, and many times I get context leaks. Suddenly one session talks about another.
I see sometimes logs like this:

> [PROXY] 0255b19a-7112-4e5b-a35c-53a2d201aba4 TOKEN WARN: Input tokens grew 100% in one turn (1 -> 2). Possible context leak or full replay.

But I can't find if its "for sure" happening when the context leak happens.

`MERIDIAN_DEFAULT_AGENT=pi MERIDIAN_PASSTHROUGH=1 meridian`

Using version `1.43.0`

Thanks.

bryancostanich:
We hit this too and traced it — the "Possible context leak or full replay" warnings and "one session talks about another" line up with the OpenCode-adapter session-resume path. (passthrough/opencode, opus, streaming.)

**1. Stale-session eviction + resume churn.** The proxy maps stateless `/v1/messages` onto stateful Claude Code SDK sessions and resumes them (suffix-overlap "Compaction detected … Allowing resume"); a stale session is evicted and retried fresh ("Stale session UUID, evicting and retrying as fresh session"). On resume, unconsumed/severed stream content from a *prior* request can surface on the *next* request's response stream as `text_delta`s — one session's content in another. We have a byte-level case where a prior user message's tail replayed verbatim as assistant deltas right after a resume.

**2. Deferred-tool loading amplifies it.** With a large tool set, auto-defer (default threshold 15) pushes most tools behind `ToolSearch` and drives per-turn `discovered=` churn keyed to session state; every stale-session eviction re-discovers them, multiplying resume/replay events. In our logs this correlated tightly with the TOKEN WARN lines.

**Mitigation that worked for us:** `MERIDIAN_DEFER_TOOL_THRESHOLD=0` (load all tools up front) removed the churn entirely — `deferred=0`, zero `discovered=`, zero `Stale session UUID` evictions across a long session, prompt cache still ~100%. Worth trying to see if your leaks drop.

**Deeper fix (defense-in-depth):** the streaming emit path forwards `content_block_*` events with no gate requiring the current request's `message_start` first, so replayed/out-of-envelope content gets forwarded rather than dropped. Bounding emission to the current request's `[message_start … message_stop]` (and never emitting non-assistant/stale parts) would stop the leak at the boundary regardless of what the SDK replays on resume.

rynfar:
looking into this

[...]

PR #850, opened 2026-08-19
fix: stop hidden passthrough turns from invalidating cache

chakflying:
## Summary

- limit streaming passthrough to the one client-visible model turn
- treat the resulting `max_turns` boundary as a successful tool handoff when the assistant checkpoint is fully settled
- persist that tool-bearing assistant checkpoint so the next client tool result resumes the same cache lineage
- retain the three-turn non-streaming budget and the existing deferred-tool and advisor allowances

## Previous data flow

Passthrough tools execute in the coding client, but Meridian must expose them through the Claude Agent SDK. Previously, streaming passthrough used a three-turn base budget:

```text
client request
  -> Claude turn 1: thinking + tool_use
  -> Meridian PreToolUse hook captures and denies SDK-side execution
  -> Claude turn 2: receives the synthetic denial and generates a hidden digest, retry, or follow-up
  -> SDK drain/canonical result
  -> Meridian returns only the original client-visible tool_use
  -> client executes the tool
  -> next request rewinds to the tool-bearing assistant with resumeSessionAt
```

The hidden turn was not part of the client conversation. The next request discarded it by rewinding to the visible assistant checkpoint. However, the hidden turn still consumed a model call and could move the automatic prompt-cache breakpoint onto the discarded branch. After the reusable ancestor cache entry expired, subsequent tool steps recreated the growing conversation prefix instead of reading the cache.

A two-turn budget had also failed historically: some fresh `opus[1m]` requests with extended thinking and tool use exhausted the budget while processing the synthetic denial and surfaced HTTP 500. Raising the shared base to three prevented that termination, but did not solve the hidden-branch cache divergence.

## New streaming data flow

For streaming passthrough, the denial means only that the client owns tool execution. It does not need another Claude turn:

```text
client request
  -> Claude turn 1: thinking + tool_use
  -> Meridian streams and captures the complete tool_use
  -> PreToolUse denial becomes iterator-visible
  -> Meridian records the settled assistant UUID and exact tool IDs
  -> maxTurns=1 ends the SDK invocation
  -> Meridian recovers that expected boundary as stop_reason=tool_use
  -> Meridian persists the settled assistant checkpoint
  -> client executes the tool
  -> next request appends the real tool_result at that same checkpoint
```

This keeps the client conversation, SDK resume point, and prompt-cache lineage on the same branch. There is no hidden digest turn to bill or later discard.

## Compatibility and safety

Non-streaming passthrough remains at a base of three turns. Its synchronous path does not yet persist the one-turn checkpoint with the same guarantees, so retaining three avoids reintroducing the historical `opus[1m]` HTTP 500 failure.

Streaming checkpoint persistence is allowed only when all of the following are true:

- termination reason is `max_turns`
- at least one captured tool call can be returned
- the SDK session and tool-bearing assistant UUID are known
- every expected tool denial has settled
- the exact pending tool ID set is available
- the request is not an independent session
- no duplicate-tool abort occurred

Other noncanonical failures continue to evict the session rather than preserving a potentially unsafe checkpoint. Deferred ToolSearch discovery still receives one additional turn, and advisor execution retains its existing three-turn allowance.

## Validation

- full test suite
- TypeScript typecheck
- production build
- provider-backed streaming smoke test covering three sequential client tool calls, stable session continuation, cache reuse, and a final text response without hidden model turns

### DIFF
--- a/src/__tests__/passthrough-early-stop-integration.test.ts
+++ b/src/__tests__/passthrough-early-stop-integration.test.ts
@@ -900,6 +900,50 @@ describe("Integration: passthrough early stop", () => {
     expect(capturedQueryParamsAll[1].options.resumeSessionAt).toBeUndefined()
   })
 
+  it("stream: preserves a settled checkpoint at the one-turn SDK boundary", async () => {
+    const toolTurn = assistantMessage([
+      { type: "tool_use", id: "single-turn-tool", name: "read", input: { file_path: "x" } },
+    ])
+    mockMessages = [
+      messageStart("msg_single_turn"),
+      toolUseBlockStart(0, "read", "single-turn-tool"),
+      inputJsonDelta(0, '{"file_path":"x"}'),
+      blockStop(0),
+      messageDelta("tool_use"),
+      toolTurn,
+      userDenyMessage("single-turn-tool"),
+    ]
+    mockTerminalError = new Error("Claude Code returned an error result: Reached maximum number of turns (1)")
+
+    const first = await post(app, {
+      model: "claude-sonnet-4-5",
+      max_tokens: 400,
+      stream: true,
+      tools: [READ_TOOL],
+      messages: [{ role: "user", content: "read x once" }],
+    }, "es-single-turn-boundary")
+    expect(first.status).toBe(200)
+    const firstBody = await first.text()
+    expect(firstBody).toContain('"type":"tool_use"')
+
+    mockTerminalError = undefined
+    mockMessages = [assistantMessage([{ type: "text", text: "the file says X" }])]
+    const second = await post(app, {
+      model: "claude-sonnet-4-5",
+      max_tokens: 400,
+      stream: false,
+      tools: [READ_TOOL],
+      messages: [
+        { role: "user", content: "read x once" },
+        { role: "assistant", content: [{ type: "tool_use", id: "single-turn-tool", name: "read", input: { file_path: "x" } }] },
+        { role: "user", content: [{ type: "tool_result", tool_use_id: "single-turn-tool", content: "X" }] },
+      ],
+    }, "es-single-turn-boundary")
+    expect(second.status).toBe(200)
+    expect(capturedQueryParamsAll[1].options.resume).toBe("test-session")
+    expect(capturedQueryParamsAll[1].options.resumeSessionAt).toBe(toolTurn.uuid)
+  })
+
   it("stream: closes at turn 1 while digest and canonical result drain invisibly", async () => {
     mockMessages = [
       messageStart("msg_es"),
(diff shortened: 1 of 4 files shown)


rynfar:
Thank you for this — and specifically for the diagnosis I missed.

I tested `MERIDIAN_PASSTHROUGH_MAX_TURNS=1` a few hours ago, watched cache reads drop to zero, and concluded the idea was net-negative. You found *why* they dropped: the streaming catch block never calls `storeSession`, so a turn-cap recovery leaves nothing to resume from. I stopped at the symptom; you found the cause.

I've split it and landed your `server.ts` half in **#853**, cherry-picked so you stay as author. That half is safe, tested, and correct on its own.

The `maxTurns` 3→1 half I'm holding, and measurements make the case for it *stronger* than your PR body does. On #853 (so the store is in place), identical 3-turn tool-calling OpenCode session, billed usage from the subprocess transcripts:

| | API calls | billed input-equiv | upstream time | session lease wait |
|---|---|---|---|---|
| default budget | 12 | 36,554 | 32,747 ms | 9,536 ms |
| `MAX_TURNS=1` | 7 | 21,574 | 17,335 ms | **0 ms** |

−42% calls, −41% spend, −47% upstream time. And the best argument isn't tokens at all: the drain holds the SDK slot **and** the per-session turn lease, so the next request in the session queues behind it — that 9.5 s of lease wait is the user waiting.

Two things blocked it, both reproduced in that run:

1. **The tool-less one-shot 500s.** OpenCode's `title` agent (`tools=0, mc=1`) has no captured tool call, so `canRecoverCapturedToolUses` returns false and the request returns `api_error`. It fires once per session, so this isn't an edge case.
2. **The recovery path records no usage.** 4 of 7 requests logged `cache_read=0 cache_write=0 output=0` while the transcripts show real spend. `maxTurns=1` promotes that path from rare to universal, so `src/telemetry/pricing.ts` prices every passthrough tool turn at $0.00 — you couldn't use the dashboard to confirm your own win — and `checkTokenHealth`, the "cache hit rate 0% on resume" alarm that would catch a broken checkpoint, isn't called there either.

Very happy to take the second half as a follow-up if you want it: carry `lastUsage`'s token fields + `cacheHitRate` into the recovery-path telemetry and call `logUsage`/`checkTokenHealth`; keep absolute headroom for the advisor and deferred-tool budgets rather than stacking their bumps on a base of 1 (#547 was a regression from underestimating exactly that); and let a tool-less turn degrade instead of erroring.

One correction on framing, since it'll matter for the follow-up: these turns don't invalidate the cache. Across 6,857 passthrough sessions they're 0.36% of all cache writes and write *less* per call than a normal turn (682 and 39 vs 2,971), and client-visible turns following a discarded branch write **fewer** cache tokens than those following a clean one (1,008 vs 3,058 per call) — the opposite of prefix recreation. They cost a full-prefix *read* plus their output, and they're waste by construction since `isPostCheckpointCall` discards every tool call from turn 2+ anyway. Worth rewording so the changelog doesn't carry the cache-invalidation claim.

Closing this in favour of #853 purely for the split; `main` requires signed commits so it needed the cherry-pick either way, which isn't something you could resolve from a fork.

PR #855, opened 2026-08-19
fix(server): bound the session tool caches that never released a key

### LINKED ISSUE #854: Unbounded session tool caches never release a key
`sessionDiscoveredTools` and `sessionToolCache` in `src/proxy/server.ts` are plain `Map`s keyed by session ID with no eviction. Every session that ever calls tools adds an entry that is never removed, so a long-running proxy instance grows these maps without bound over the life of the process.

`sessionMcpCache` (server.ts:412) already solves this for the MCP cache by using `LRUMap` sized via `getMaxSessionsLimit()` (`MERIDIAN_MAX_SESSIONS` / `CLAUDE_PROXY_MAX_SESSIONS`, default 1000). The two remaining unbounded maps should follow the same pattern.

mpeter:
## Problem

`sessionDiscoveredTools` and `sessionToolCache` in `src/proxy/server.ts` are plain `Map`s keyed by session ID with no eviction path. Every session that calls a tool adds an entry that lives for the lifetime of the process — on a long-running proxy this grows without bound.

## Fix

Convert both to `LRUMap` sized by `getMaxSessionsLimit()`, matching the pattern already used by `sessionMcpCache` (server.ts:412).

## Changes

| File | Change |
|---|---|
| `src/proxy/server.ts` | `sessionDiscoveredTools`/`sessionToolCache` → `LRUMap<string, ...>(getMaxSessionsLimit())` |
| `src/__tests__/proxy-session-map-bounds.test.ts` | new — LRU eviction, in-limit restore, recency-on-read |

Closes #854

### DIFF
--- a/src/__tests__/proxy-session-map-bounds.test.ts
+++ b/src/__tests__/proxy-session-map-bounds.test.ts
@@ -1,0 +1,177 @@
+/**
+ * The session-keyed maps inside createProxyServer are never deleted from — a
+ * key is added the first time a session is seen and stays for the life of the
+ * process. Bounding them is only observable through eviction, so these tests
+ * pin the limit at 2 and drive a third session past it.
+ */
+
+import { afterAll, beforeEach, describe, expect, it, mock } from "bun:test"
+import { assistantMessage, makeRequest } from "./helpers"
+
+const originalMaxSessions = process.env.CLAUDE_PROXY_MAX_SESSIONS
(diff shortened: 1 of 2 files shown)


mpeter:
Closing as superseded by #864, which merged the same LRU bounds for `sessionDiscoveredTools` and `sessionToolCache` with eviction coverage.

PR #857, opened 2026-08-19
fix(errors): classify an oversized prompt as a 400, not a retryable 500

### LINKED ISSUE #856: Oversized prompt is classified as a retryable 500 instead of a 400
When an upstream request exceeds the model's context window, the SDK error surfaces through `classifyError` as a generic retryable 500 (it falls through to the process-crash branch), so the client retries an identical oversized prompt and fails the same way every time — with no actionable message telling it to trim/compact first.

`classifyError` should detect the context-overflow signals ("prompt is too long", "input length and `max_tokens` exceed context limit", "context length exceeded", etc.) and return a 400 `invalid_request_error` with a message that tells the caller the retry won't help without trimming the conversation.

mpeter:
## Problem

When a request exceeds the model's context window, the upstream error currently falls through `classifyError` into the generic process-crash / retryable-500 path. The client retries an identical oversized prompt and hits the same failure every time, with no message telling it the retry won't help.

## Fix

Detect context-overflow signals (CLI wording, API token-count wording, `max_tokens` phrasing, and the OpenAI-compatible variant) in `classifyError` and return a 400 `invalid_request_error` before the process-crash branch is reached, with a message telling the caller to compact/trim first. Also extends `SdkTermination.reason` with a `context_overflow` variant so `extractSdkTermination`/`formatSdkTermination` name it correctly instead of reporting an unknown process exit.

## How it works for users

A prompt that exceeds context now fails fast with a clear 400 instead of being silently retried as a 500.

## Changes

| File | Change |
|---|---|
| `src/proxy/errors.ts` | `CONTEXT_OVERFLOW_SIGNALS`, new `classifyError` branch (400), `context_overflow` added to `SdkTermination.reason`, matching branch in `extractSdkTermination` |
| `src/__tests__/errors.test.ts` | new — wording variants, ordering vs. process-exit/stderr, never-retryable table, `context_overflow` naming + round-trip through `formatSdkTermination`, `canRecoverCapturedToolUses` default-false case |

Closes #856

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -237,6 +237,12 @@ describe("classifyError", () => {
       expect(canRecoverCapturedToolUses({ ...base, reason: "process_exit" })).toBe(false)
     })
 
+    // An overflow is rejected before generation, so there is nothing captured
+    // to deliver — and replaying the same oversized prompt cannot succeed.
+    it("does not recover a context overflow", () => {
+      expect(canRecoverCapturedToolUses({ ...base, reason: "context_overflow" })).toBe(false)
+    })
+
     // Without captured calls there is nothing to deliver, and outside
     // passthrough the client does not execute tools at all.
     it("requires captured tool calls", () => {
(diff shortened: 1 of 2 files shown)


mpeter:
Ready for review. CI and Docker are awaiting upstream approval to run against this fork PR. The change is isolated to context-overflow classification: oversized prompts return 400 `invalid_request_error` instead of a retryable 500, including the SDK process-exit wording.

rynfar:
Landed in #919 (squash-merged to `main`) — thanks, this was a good catch.

Cherry-picked your commit so you stay the author; GitHub's `Co-authored-by:` trailer carries the credit to `main`. I couldn't merge this branch directly because `main` requires signed commits, which fork branches can't satisfy — a constraint on my side, nothing to do with your change.

Two things worth passing back.

**Your diagnosis was right, and the bug is slightly worse than the issue says.** I confirmed each case against current `main` before touching anything. The 500s were as you described. But when the CLI surfaces an oversized prompt by *exiting*, the bare code-1 exit reaches the process-crash branch and returns **401 — "try `claude login`"**. So an oversized prompt was telling operators to re-authenticate. Your ordering argument — putting the overflow branch above the process-crash branch — is what fixes that, and it's the strongest part of the change.

**I added a follow-up commit anchoring the patterns.** Unanchored, they matched their own strings quoted inside arbitrary text; all four of these classified as 400:

```
"The prompt is too long check did not trigger; this is a network failure"
'The assistant replied: "prompt is too long" is a common error message users see'
'Subprocess stderr: user asked "why does it say prompt is too long?"'
"tool_result: grep found 'context_length_exceeded' in errors.ts:114"
```

A false 400 is the expensive direction here — it tells the client the request is unfixable, so the retry is abandoned and legitimate work is silently dropped. We hit the mirror image of this in #908 last week with the spend-limit patterns.

**Your test is what caught my over-correction.** Anchoring per-line alone rejected the API's own envelope — `API Error: 400 {"type":...,"message":"prompt is too long: N > M"}` — which is a real upstream shape and isn't line-anchored. Your token-count test failed immediately. The pattern now also accepts the phrase when it opens a `message` value, which still rejects the same words in another field or partway through the message. Your "ignores an incidental identifier" test told me you were already thinking about this failure class; the anchor just makes it structural rather than incidental.

Full suite 3232 pass, 0 fail.

Issue #861, opened 2026-08-20
Crossing the auto-defer tool threshold invalidates the whole prompt cache mid-session


rynfar:
## Problem

`createPassthroughMcpServer` decides deferral per request from the live tool count:

```ts
// src/proxy/passthroughTools.ts
const autoDefer = !!(threshold > 0 && coreToolNames && coreToolNames.length > 0 && tools.length > threshold)
const hasDeferredTools = tools.some(t => t.defer_loading === true) || autoDefer
```

Default threshold is 15. A client that crosses that boundary mid-conversation — one tool added or removed — flips deferral for **every** non-core tool at once. That changes:

1. The `_meta: { "anthropic/alwaysLoad": true }` marker on each non-core tool, i.e. the tool definitions themselves. Tools render at position 0 of the prompt, so this invalidates the tools, system, *and* messages cache tiers — a full cold replay of the conversation.
2. `ENABLE_TOOL_SEARCH` in the subprocess env.
3. **`maxTurns`** — since #860, `hasDeferredTools` also gates digest-turn elimination (`query.ts`: `ctx.earlyStop !== false && !hasDeferredTools && ...`). So crossing the threshold silently re-enables the billed digest turn for that request.

Realistic triggers: OpenCode switching modes/agents, an MCP server connecting or dropping mid-session, a plugin toggling a tool. None of these look like a cache-affecting change to the user.

The blast radius is the point — one added tool re-renders the entire tool list, not just its own entry.

## Suggested fix

**Pin the deferral decision per session**, alongside the caches that already exist for exactly this reason (`sessionMcpCache` / `sessionToolCache`, `server.ts` ~485). The first request in a session decides; later requests inherit it unless the tool set changes enough that the MCP server is rebuilt anyway. That matches the existing intent — `sessionToolCache` already exists to stop an intermittently-omitted tool list from busting the cache, and there is already a `tools_changed: MCP server recreated (prompt cache likely invalidates)` log.

Alternatives considered:

- **Hysteresis** (defer above `threshold`, un-defer only below `threshold - N`). No new state, but only helps a client oscillating *at* the boundary — not one that legitimately swings wide.
- **Log and observe first.** Cheapest: emit an explicit `defer_flip` event + telemetry counter so the frequency is measurable before changing behaviour. Reasonable as a first step if the pinning design is contentious.

Recommend pinning, with the flip logged either way.

## Related cleanup

[...]

Issue #767, opened 2026-08-07
OpenCode + Opus: nearly every turn diverges as `modified-history` with prefix overlap exactly `messageCount - 1`, forcing a full fresh replay (Haiku is unaffected)


tetipong2542:
## Environment

| | |
|---|---|
| Meridian | 1.58.3 (lineage code verified byte-identical in 1.60.0) |
| opencode-with-claude | 1.7.0 |
| OpenCode | 1.18.12 |
| Adapter / mode | `opencode` / **passthrough** |
| Model | `claude-opus-4-7` (`opus[1m]`, variant `medium`) |
| Agent | oh-my-openagent "Sisyphus - ultraworker" |
| OS | macOS 15 (darwin arm64) |

## Summary

In a normal 27-turn coding session, **29 of 31 tool-bearing requests were classified `lineage: "new"`**. Every one logged `modified-history`, and the prefix overlap is always *exactly one less* than the stored message count — the last stored message never re-hashes the same way on the following request.

Because each divergence triggers `buildFreshPrompt`, the whole transcript is re-sent as fresh input. Cache hit rate decays 97% → 32% and the session cost roughly 3.9x what it should.

## Evidence

`[PROXY]` log, one line per turn, unbroken for the whole session:

```
Stale session detected (key=ses_0260…): prefix overlap 2/3,   incoming 5 msgs.  Starting fresh replay.
Stale session detected (key=ses_0260…): prefix overlap 8/9,   incoming 11 msgs. Starting fresh replay.
Stale session detected (key=ses_0260…): prefix overlap 10/11, incoming 13 msgs. Starting fresh replay.
...
Stale session detected (key=ses_0260…): prefix overlap 48/49, incoming 51 msgs. Starting fresh replay.
Stale session detected (key=ses_0260…): prefix overlap 50/51, incoming 53 msgs. Starting fresh replay.
```

`overlap == messageCount - 1` on every single line. Per `verifyLineage`'s decision matrix that lands in
"cached prefix changed while growing → diverged", i.e. only the **final** stored message fails to match.

`GET /telemetry/requests` for the same session:

```
msgs  lineage        resume  tools  cacheCreation  cacheRead  hit%
   1  new            False      83         55,816          0    0
   3  continuation   True       83            341     55,816   99
   5  new            False      83          1,739     54,580   97
  11  new            False      83         24,186     54,580   69
  15  continuation   True       83            466     81,167   99
  23  new            False      83         40,669     54,580   57
[...]

PR #871, opened 2026-08-23
fix(passthrough): preserve checkpoint resume continuity

justprosh:
## Summary

- preserve the assistant checkpoint when a complete tool-result batch is followed by queued user input
- make every passthrough checkpoint rewind durable, so delivered client `tool_result` blocks remain in later resume history instead of reverting to the persisted PreToolUse denial
- discard copied rollback UUIDs when the SDK returns a forked session, while retaining the newly observed assistant UUID
- fail closed to structured fresh replay for partial, unknown, duplicate, split, or late results, stray assistant turns, and undo prefixes whose UUIDs were remapped by a fork

### Queued user input forced a fresh replay

OpenCode can append a queued user message while a passthrough tool loop is in progress, producing `user[tool_result] -> user[text]`. The continuation validator rejected every second user message, so the proxy cleared `resumeSessionAt` and replayed the full history in a fresh SDK session.

The validator now coalesces consecutive user turns only after proving that the first turn settles the exact checkpoint batch. Queued string, text-block, and image-block content remains ordered; invalid result sequences still fail closed. Parallel results are matched by `tool_use_id`, as required by the Anthropic protocol.

### A successful resume rewrote an older result

The PreToolUse denial that prevents server-side execution is persisted in the Claude Code transcript. `resumeSessionAt` rewound that denial for one query, but the source transcript retained it. On the next resume, Claude Code serialized the old call as Meridian's 209-byte forwarding marker instead of the client result, invalidating the prompt-cache prefix.

Passthrough checkpoint resumes now use `forkSession: true`. The fork makes the replacement tail durable: the client result becomes ancestry in the returned session, while the denial remains only on the current unresolved tool boundary.

Forking remaps every copied message UUID. Session storage therefore reconciles rollback UUIDs against the session ID actually returned by the SDK: copied UUIDs are cleared on a session change, the current output UUID is retained, and an older undo prefix with no valid boundary goes directly to a structured fresh replay instead of first failing with `missing-message`. This applies to normal stream/non-stream completion and the recoverable streaming `error_max_turns` checkpoint path.

## Production impact

In one live passthrough session before the fixes:

| turn type | n | avg cache hit rate | cache write |
|---|---:|---:|---:|
| resumed | 355 | 0.88 | ~24k/turn |
| checkpoint replay | 44 | 0.36 | ~77k/turn |

That produced 3.4M avoidable cache-write tokens at the 1.25x rate. After queued-message resume was restored, a separate successful continuation still collapsed from a 93.5% cache hit to 30.1% because an older delivered result reverted to the forwarding marker; this PR now addresses both causes.

Related: #767, #861, #869.

## Verification

- `npm test`: 2727 passed, 1 skipped, 0 failed, plus all process-isolated suites
- `npm run typecheck`
- `npm run build`
- `git diff --check`
- focused HTTP regressions cover queued string/multimodal input, repeated and parallel tool rounds, changed fork session IDs, rollback UUID reconciliation, direct fresh undo, and streaming capped-turn checkpoint persistence
- isolated end-to-end HTTP audit against the freshly built CLI with Claude Code 2.1.221 / Agent SDK 0.2.141:
  - four passthrough turns returned four distinct chained session IDs
  - intercepted historical `tool_result` payloads stayed byte-stable; the only prefix change was expected `cache_control` movement
  - the forwarding marker did not reappear in subsequent API request bodies
  - a server-side `bash` sentinel was not executed
  - undo to a remapped old prefix made one fresh request with no `missing-message` attempt

The end-to-end audit used a dummy key in an isolated network namespace with only a loopback interceptor; no external Anthropic request was possible.

### DIFF
--- a/src/__tests__/lineage-unit.test.ts
+++ b/src/__tests__/lineage-unit.test.ts
@@ -15,6 +15,7 @@ import {
   verifyLineage,
   normalizeContextUsage,
   withClientAssistantUuid,
+  reconcileReturnedSessionUuids,
   MIN_SUFFIX_FOR_COMPACTION,
   type SessionState,
 } from "../proxy/session/lineage"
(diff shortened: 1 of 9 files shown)


justprosh:
Post-deployment follow-up: the cache behavior is healthy in our live staging cohort after this fix.

The running build is [`1.62.7-queued-resume.3`](https://github.com/justprosh/meridian/releases/tag/staging-1.62.7-queued-resume.3), which contains the complete change from this PR plus an unrelated profile-aware model-catalog change. I also verified the running bundle, not only the package source: passthrough checkpoint resumes set `forkSession: true` when `resumeSessionAt` is present.

For the UTC window `2026-08-26 10:03:14–11:03:14`, persisted live telemetry contained 206 cache-measurable requests matching:

```text
isPassthrough = true
isResume = true
lineageType = continuation
cacheHitRate != null
```

Results:

```text
status:              206/206 HTTP 200
cache hit, minimum:  92.420%
cache hit, median:   99.737%
cache hit, weighted: 99.649%  (135,196,130 read / 475,965 written)
rows below 90%:      0
```

A representative sustained chain from `messageCount=577` through `596` had 10 continuation requests, all HTTP 200:

```text
cache hit:    99.541%–99.892%
cache read:   578,772–1,725,686 tokens
cache write:  888–3,379 tokens
```

For comparison, the pre-fix failure that led to this PR produced `30.1%` hit, `38,058` read, and `88,171` written on the affected resume. That collapse did not recur in this observed chain or cohort.

This is deliberately a scoped claim: one deployed build and one live observation window, not a same-session A/B or proof for every possible resume path. Within that scope, this is the production-shaped evidence I wanted before calling the fix successful in our environment.

amasover:
This seems to work for me -- I was burning through my usage limit like crazy before I started using this change.

rynfar:
Superseded by #879, now merged as `2ccc48a5`.

Thank you, Aleksey. Your three commits were carried into #879 as patch-identical cherry-picks (`b35362a7`, `f82b2606`, `f798f362`), and the squash commit retains your credit with an explicit `Co-authored-by` trailer. #879 adds the crash-safe lifecycle, cross-process fencing, bounded supported deletion, and exact-final real SDK/OpenCode validation needed for production.

PR #872, opened 2026-08-24
Resume the session when a client only adds or drops content blocks

CrazyCoder:
Two block-level mutations force a full history replay, and a replayed turn that also queued behind the session turn lease returns HTTP 400 `session_turn_conflict`.

## Appended blocks on a tool-result slot

This is #767. The append-only path required every appended block to be a new `tool_result`. OpenCode trails reminder text after the results of a turn, so its trailing user message grows from `[tool_result, tool_result]` to `[tool_result, tool_result, text, text]`. The turn then diverges as `modified-history` with prefix overlap exactly `messageCount - 1`, which is the signature reported in #767.

Appended blocks that are not `tool_result` are now accepted, but only on a slot that already carries a `tool_result`. Appended text on a plain user message is the user editing their own turn, and that still diverges. The existing test for it stays green.

## Blocks injected for one turn, then dropped

A Claude Code `UserPromptSubmit` hook writes to stdout. oh-my-openagent bridges that hook into OpenCode and injects the stdout as a `<user-prompt-submit-hook>` text block. The block rides on the turn that submits the prompt, and it is absent on the next turn.

`message[0]` therefore changes from `[text, text]` to `[text]` in the middle of a conversation. Prefix overlap is 0, so the whole conversation classifies as `unrelated-history`.

I measured this on 1.62.7 with an OpenCode client: 0 `continuation` lineages in 38 requests. The `pi` and `prime` adapters resumed normally over the same period.

`droppedEphemeralBlocksOnly()` resumes when the only differences in the stored region are blocks removed from user turns. Every surviving block must be byte-identical and in its original order.

## Why a resume is safe here

The difference is subtractive. The SDK session holds a superset of the history the client now sends, so a resume cannot skip anything the client believes it sent. That is the opposite of the direction guarded in #689, #692 and #712, where the client sends history the SDK session never saw.

Four conditions must all hold:

- The history grew. An equal or shorter history keeps the classification it has today.
- Every changed slot is a user turn. A changed assistant turn diverges.
- Every changed slot has strictly fewer blocks than the stored slot.
- Surviving blocks are byte-identical and still in their original order.

Both hash arrays must also cover the whole stored region. A short array would read an out-of-range `undefined` as "changed".

## Verification

I replayed the injected-block shape against 1.62.7 and against this branch. Turn 1 carries the injected block, turn 2 drops it and appends an exchange, turn 3 is steady state.

| Turn | 1.62.7 | This branch |
|---|---|---|
| 1, injected block present | new | new |
| 2, injected block dropped | new | continuation |
| 3, steady state | continuation | continuation |

Turn 2 is the case this PR changes. On 1.62.7 it replays the full history against a cold prompt cache.

131 tests pass across `lineage-unit`, `lineage-safety`, `session-lineage`, `lineage-thinking-blocks`, `opencode-title-agent-collision` and `proxy-concurrency-coordination`. Nine of them are new.

Every new guard has a test that fails when the guard is relaxed, and I checked that rather than assuming it. Disabling the new path fails the two positive tests. Relaxing the user-turn check and the ordered-subsequence check fails three negative tests.

## Scope

The append-only change alters an existing condition, so it can admit shapes I did not see in the field. The dropped-block path is new code and cannot change any classification that reaches it today, because every case it accepts currently returns `diverged`.

I did not touch the turn lease or the conflict response. A client that genuinely sends unrelated history still gets the 400.

### DIFF
--- a/src/__tests__/lineage-unit.test.ts
+++ b/src/__tests__/lineage-unit.test.ts
@@ -643,7 +643,7 @@ describe("formatLineageMismatch", () => {
   })
 })
 
-// Shared by the two block-level suites below. Both need a session whose per-block
+// Shared by the block-level suites below, which need a session whose per-block
 // hashes were recorded — that is what makes block-granular classification
 // possible, and a session without them is expected to keep replaying.
 const toolResult = (id: string, content: string) => ({
--- a/src/__tests__/lineage-unit.test.ts
+++ b/src/__tests__/lineage-unit.test.ts
@@ -643,23 +643,26 @@ describe("formatLineageMismatch", () => {
   })
 })
 
-describe("verifyLineage append-only tool-result extension", () => {
-  const toolResult = (id: string, content: string) => ({
-    type: "tool_result",
-    tool_use_id: id,
-    content,
+// Shared by the two block-level suites below. Both need a session whose per-block
+// hashes were recorded — that is what makes block-granular classification
+// possible, and a session without them is expected to keep replaying.
+const toolResult = (id: string, content: string) => ({
+  type: "tool_result",
+  tool_use_id: id,
+  content,
+})
+
+function sessionWithBlockHashes(messages: Array<{ role: string; content: any }>): SessionState {
+  return makeSession({
+    lastAccess: 0,
+    lineageHash: computeLineageHash(messages),
+    messageCount: messages.length,
+    messageHashes: computeMessageHashes(messages),
+    messageBlockHashes: computeMessageBlockHashes(messages),
   })
+}
 
-  function sessionFor(messages: Array<{ role: string; content: any }>): SessionState {
-    return makeSession({
-      lastAccess: 0,
-      lineageHash: computeLineageHash(messages),
-      messageCount: messages.length,
-      messageHashes: computeMessageHashes(messages),
-      messageBlockHashes: computeMessageBlockHashes(messages),
-    })
-  }
-
+describe("verifyLineage append-only tool-result extension", () => {
   it("resumes from only the newly appended parallel tool result", () => {
     const stored = [
       { role: "user", content: [{ type: "text", text: "run both" }] },
@@ -682,9 +685,9 @@ describe("verifyLineage append-only tool-result extension", () => {
       { role: "user", content: [toolResult("call-c", "c-result")] },
     ]
 
-    expect(verifyLineage(sessionFor(stored), incoming)).toEqual({
+    expect(verifyLineage(sessionWithBlockHashes(stored), incoming)).toEqual({
       type: "continuation",
-      session: sessionFor(stored),
+      session: sessionWithBlockHashes(stored),
       resumeFrom: 2,
       resumeContentFrom: 1,
     })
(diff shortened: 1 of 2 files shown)


rynfar (changes requested):
Requesting changes. The appended-text symptom remains relevant, but the current dropped-block path is not safe to merge after #879.

Blocking issues:

1. `droppedEphemeralBlocksOnly()` does not prove that a removed block is ephemeral. It accepts deletion of any hashable block from any incoming user-array slot while history grows. `SessionState.messageBlockHashes` does not retain the cached role or block type, so checking only the incoming role cannot prove the stored slot was a user turn. A cached assistant slot rewritten as a user slot can pass. Returning `resumeFrom: cached.messageCount` then omits the altered slots while the SDK session still contains the removed or role-changed content. With #879, that incorrect superset can be durably published and reused across processes/restarts.

2. The valid `UserPromptSubmit` case is already handled narrowly by #879 in `openCodeAdapter.canonicalizeMessagesForLineage`. It validates the exact envelope, preserves message positions, and leaves the execution body unchanged. Please remove the generic dropped-block path rather than broadening pure lineage around an OpenCode-specific symptom.

3. The append relaxation admits every non-`tool_result` block once the *incoming* prefix appears to contain a tool result. It neither restricts the observed case to text nor proves cached block types. The lineage normalizer is not type-domain-separated for text: text equal to `tool_result:<id>:<content>` hashes like that tool-result block. Type decisions therefore cannot be reconstructed safely from incoming blocks plus stored hashes alone.

Please rebase on current main and retain only a narrowly proven appended-text continuation. It needs HTTP integration through current canonicalization and server wiring, plus stream/non-stream durable-fork, restart/cross-process, queued-lease, OpenCode V1, and exact supported V2 coverage. Add adversarial negatives for ordinary text/tool-result/image/document deletion, stored-role flips, multiple historical drops, and text/tool-result type collisions. The existing pre-#879 green checks do not validate these semantics on current main.

connor-grady:
Two things in support of the appended-block half, from a client with none of the plugins this PR cites.

**The append case reproduces without oh-my-openagent.** The PR justifies it with OMO trailing reminder text, and the review points at #879 as already covering the legitimate case. It does not cover a stock client: `isTransientUserPromptHook` matches only a text block that is entirely a `<user-prompt-submit-hook>` element wrapping JSON with `hookSpecificOutput.hookEventName === "UserPromptSubmit"`, and `sst/opencode` contains zero occurrences of that string. I see the same `overlap == messageCount - 1` divergences on stock OpenCode 1.18.21 with no OMO and no hook bridge, on Meridian 1.62.7 and unchanged on 1.64.0 (`verifyLineage` differs only by a minifier rename between those two bundles). Shapes and details are on #767. That is an argument for the appended-text path that does not depend on this PR's original justification.

**Objection 3 constrains the narrow form too, and the migration is cheaper than it looks.** The type-domain observation is correct: `normalizeContent` emits a text block as raw text and a tool result as `tool_result:<id>:<content>`, so per-block hashes cannot carry type information. Any predicate that infers "the appended block is text" or "this stored block was a tool_result" from `messageBlockHashes` is unsound, including a narrowed one. Prefixing the block type inside `computeMessageBlockHashes` alone fixes it without touching `computeLineageHash` or `hashMessage`, so no stored session is invalidated and nobody pays a cold replay on upgrade. Stale per-block arrays written by an older build fail the existing `every` check and fall through to today's behaviour, which the `cached.messageBlockHashes?.length === cached.messageCount` guard already handles. I have filed that as #887 so it survives whatever happens to this PR.

I am not asking to take this over and I am not disputing the review. I can supply field evidence from a stock client if a narrowed version is attempted.

Issue #707, opened 2026-07-28
models-auth-status.test.ts tests a local reimplementation, not the real auth cache


rynfar:
Found while sweeping for tests that don't exercise the module they claim to cover (the class of problem fixed for `settings.ts` in #703).

## The problem

`src/__tests__/models-auth-status.test.ts` opens with:

> Mocks `../proxy/models` to fully control `getClaudeAuthStatusAsync` behavior, eliminating races from parallel test files that share the module singleton.

It does not import or mock `../proxy/models`. It declares its own `authCache`, `authCacheAt`, `lastKnownGood`, `authIsFailure`, `AUTH_TTL` and `FAIL_TTL`, reimplements the caching logic locally, and asserts against that copy.

All 8 tests would pass if `getClaudeAuthStatusAsync` were deleted.

## Why it matters

The behaviour it claims to cover is auth-status resilience — TTL expiry, the failure path, and the `lastKnownGood` fallback that keeps the proxy usable when an auth check fails. That is a real resilience property with zero real coverage.

It is also one of the files `npm test` runs in isolation, which suggests the module singleton caused cross-file races. Reimplementing the logic locally made the races disappear by no longer touching the module — the coverage went with them.

## Fix

Make the auth cache in `models.ts` injectable (or resettable) so the tests can drive the real implementation deterministically, the way `MERIDIAN_CONFIG_DIR` made `settings.ts` testable in #703. Then rewrite the 8 tests against the real `getClaudeAuthStatusAsync`.

This touches production code on the auth path for testability, so it deserves its own change rather than being folded into unrelated work.

## Related

`src/__tests__/proxy-passthrough-concept.test.ts` (10 tests) has the same shape — simulated SDK messages, no source import — but is honestly labelled "Concept Tests" and reads as documentation rather than a false coverage claim. Lower priority; worth deciding whether it should count as tests at all.

## Sweep method

```bash
for f in src/__tests__/*.test.ts; do
  grep -qE '"\.\./' "$f" || echo "$f"
done
```

Catches static and dynamic imports (many integration tests use `await import("../proxy/...")` after mocking). Across 143 test files this leaves only the two above.

Issue #381, opened 2026-04-17
Claude native binary issue, If running on remote server and accessing via host machine.


semi710:
Command I used to run `meridian`:

```
MERIDIAN_HOST=0.0.0.0 meridian
```

`meridian setup` had ran successfully on remote machine, and `claude` binary is present on remote machine at the location that i get in the error `/home/<remote>/.local/bin/claude`.

It works fine if i use it on remote machine itself.
But i get error on my host machine if i try to connect, 
Below is the connection command.
```
ANTHROPIC_API_KEY=x ANTHROPIC_BASE_URL=http://<tailscale-host>.<tail-random>.ts.net:1234 opencode
```

error
```
Claude Code native binary not found at /home/<remote>/.local/bin/claude. Please ensure Claude Code is installed via native installer or specify a valid path with options.pathToClaudeCodeExecutable.
```

eliasstepanik:
Same Error!

eliasstepanik:
Setting these env vars fixed it for me.   

 environment: 
      - CLAUDE_PROXY_PASSTHROUGH=1
      - MERIDIAN_PASSTHROUGH=1
      - MERIDIAN_WORKDIR=/home/claude

rynfar:
thanks for posting a workaround - let me look into if we can set this automatically

semi710:
@eliasstepanik Thanks for posting the workaround, I can confirm this is working.

For someone else trying `MERIDIAN_WORKDIR=/home/<remote>` It's the path of the user where claude binary is present.

For my case below is the command that i use.
```
CLAUDE_PROXY_PASSTHROUGH=1 MERIDIAN_PASSTHROUGH=1 MERIDIAN_WORKDIR=/home/<remote> MERIDIAN_HOST=0.0.0.0 MERIDIAN_PASSTHROUGH=1 MERIDIAN_PORT=1234 meridian
```

semi710:
Keeping this open till it gets in the native tool itself as mentioned in https://github.com/rynfar/meridian/issues/381#issuecomment-4268135879

rynfar:
Sorry for the delay on this my plate has been full.  I will get to this once I wrap up a copule of other things.

rynfar:
I am finally getting back to this.  I am so sorry @semi710  for taking so long, i had a long back log of things to knock out.

rynfar:
Fixed in #473 (just merged into \`main\`).

[...]

PR #873, opened 2026-08-25
fix(opencode): read the client working directory past the workdir override

CrazyCoder:
An OpenCode client is told its working directory is the proxy's whenever the SDK directory is pinned. The model then reports that path as its own.

I reproduced it against Meridian in Docker with `CLAUDE_PROXY_WORKDIR=/app`. An OpenCode 1.18.22 client on Windows asked the model for its working directory, and the model answered `/app`. The client had sent its real path in the request.

## Where the path is lost

OpenCode reports its directory in an `<env>` block, and `extractClientCwd` parses that shape correctly. The adapter wires the parse to `extractWorkingDirectory` only.

`resolveSdkWorkingDirectory` ranks `MERIDIAN_WORKDIR` and `CLAUDE_PROXY_WORKDIR` above the adapter value, so with an override set `claimedWorkingDirectory` becomes the pinned path and the client's path is gone.

`server.ts` reads `adapter.extractClientWorkingDirectory?.(body) || cwdResolution.claimedWorkingDirectory`. The OpenCode adapter does not implement the first, so the value is the pinned path. `buildCwdNote` compares that against the SDK cwd, finds them equal, and returns an empty string. Nothing contradicts the SDK's own env block, which advertises the proxy's directory.

Replaying a captured request through those four functions, with `exists()` answering as a Linux host would:

| `CLAUDE_PROXY_WORKDIR` | `clientWorkingDirectory` | cwd note |
|---|---|---|
| unset | the client's path | emitted |
| `/app` | `/app` | empty |

Unset, the #381 fallback works as designed: the Windows path fails `existsSync`, the SDK spawns in `process.cwd()`, and `claimedWorkingDirectory` still carries the client path. The override is what removes the fallback's input.

## The change

`openCodeAdapter` implements `extractClientWorkingDirectory`, returning the same `extractClientCwd(body)` the adapter already imports. That value is read before the resolution, so an operator who pins the SDK directory no longer erases it. `piAdapter` does this for the same reason and its comment says so.

The value also feeds fingerprint bucketing at `server.ts`. On a pinned deployment every project collapsed into one key, which is the collision the bucketing exists to prevent. Clients that send an explicit session header were unaffected in practice.

## Verification

The fixture is a real request captured from OpenCode 1.18.22 on Windows through a recording proxy. The working directory is redacted and the tool list removed. Nothing on the parse path reads `tools`. A hand-written shape is what #707 warns against, so the capture is the instrument here.

132 tests pass across `opencode-cwd-note`, `cwd`, `claude-code-cwd-note`, `query-cwd-note`, `adapter-detection`, `fingerprint-unit`, `opencode-title-agent-collision` and `opencode-internal-agent-session-key`. Six are new. `tsc --noEmit` reports no errors.

The regression test pins the pinned-directory case: it asserts the resolution still reports `/app` as both the SDK cwd and the claimed path, and that the note names the client path anyway. Reverting the adapter method fails it.

One test guards the instrument rather than the code. A fixture whose system prompt lost its `<env>` block would make every other assertion in the file pass for the wrong reason.

## Scope

`extractWorkingDirectory` keeps its current precedence. An operator who pins the SDK directory still gets that directory for the subprocess, which is the point of the setting. Only the note and the fingerprint key change.

I did not touch the other adapters. Seven leave `extractClientWorkingDirectory` undefined and rely on the same fallback: `cherry`, `codex`, `crush`, `droid`, `forgecode`, `jcode` and `openai`. Each carries the same defect wherever its client reports a path and an override is set. I have no captured requests for them, and a fix without one would be the invented shape #707 describes.


rynfar (changes requested):
This fixes a real bug and the small adapter change is directionally correct, but the current head needs stronger proof before merge.

Required changes:

1. Rebase on current main. Existing checks predate #879 and #880, so they do not exercise the merged durable-resume or OpenCode V2 tree.
2. Add HTTP integration through adapter detection and current `server.ts`, with a real workdir override. Assert both that SDK `options.cwd` remains the proxy path and that the exact client path reaches the CWD note. The current helper test locally reproduces server wiring, so it can stay green if production stops consuming `extractClientWorkingDirectory`.
3. Cover the #879 durable identity interaction: keyed OpenCode must remain stable; explicitly test/document the one-time headerless key transition from override path to client path, including restart behavior.
4. Prove request-shape compatibility for OpenCode V1 and the exact supported V2 beta. #880 changes headers, not bodies, but that is not evidence that hidden and primary V2 requests always carry this `<env>` shape.
5. Make the test platform-neutral and add meaningful Windows coverage. Hard-coded `/app` plus `path.resolve()` is not portable, and the current Windows smoke does not run this new file. Assert the exact extracted path rather than `toContain`, which can accept a wrongly prefixed path on POSIX.
6. Mark the new OpenCode-specific behavior with the required `NOTE:` comment from `AGENTS.md`.

No production architecture objection: this belongs on `openCodeAdapter`, the override should continue controlling only SDK cwd, and the client path should independently control the note/project identity. Please keep the fix and harden its tests rather than close the PR.

CrazyCoder:
Superseded by #885. The replacement keeps this adapter fix and adds the requested current-main HTTP coverage, V1/V2 fixtures, durable key-transition coverage, Windows CI, and `NOTE:` marker.

Issue #619, opened 2026-07-13
Bug report: fresh-session prompt rebuild causes model self-play (transcript leakage)


duncanam:
Disclosure: I'm a Rust guy, and all I did was point Fable at the issues I was seeing with examples and live session data. Basically, the sessions seemed to get tripped up after an upgrade, where it was self-responding as me, without me ever giving a response (spooky). If this is garbage, please delete this issue report. 

Thanks for working on Meridian! It's made Pi usage tolerable with Anthropic. 

### Symptom

When a client (an agent harness) has a long tool-heavy conversation and Meridian falls off the session-resume path, the model's next response degenerates into self-play: it generates fake `Human: [your bash …]:` tool results and its own `Assistant:` replies inside a single completion, confabulating tool output for many "turns" without any tools executing. The polluted text is then stored as a real assistant message in the client's history, so every subsequent request replays it and the failure is self-reinforcing.

### Root cause

Observed in the published bundle at `dist/cli-vhrprwj4.js`; the source lives under `src/proxy/` — locate the functions by the names below.

1. Normally each request runs `lookupSession(profileSessionId, messages, cwd)` and, on a `continuation` lineage, resumes the underlying SDK session, sending only the new tail messages.
2. When lineage is `diverged`/`new`, or when a resume fails with the "Stale session UUID, evicting and retrying as fresh session" retry path (two occurrences, non-stream and stream), the request is rebuilt via **`buildFreshPrompt(allMessages, sanitizeOpts)`**. (An equivalent inline flattening also exists in the main request handler for the non-resume case — fix both.)
3. `buildFreshPrompt` flattens the entire structured conversation into **one bare text string**:
   - assistant turns → `Assistant: <flattened text>`
   - user turns → `Human: <text>`, with `tool_result` blocks rendered as `[your <toolname> <target>]:\n<output>` via `describeToolCall` / `buildToolUseIndex`
   - turns joined with `\n\n`
4. This string is sent as a single user prompt with **no framing envelope** — nothing tells the model it's replayed context rather than a live transcript format to continue. The model pattern-continues the `Human:/Assistant:` structure, hallucinating both sides. No stop sequence is available on the Agent SDK path to cut it off at `\n\nHuman:`.

There is also a partial variant of the problem in the multimodal branch, where assistant turns are re-sent as user messages with content `[Assistant: <text>]` — same leakage risk, same fix applies.

### Requested fix

[...]

Issue #552, opened 2026-06-26
Claude is really confused with the Edit tool


FelipeSharkao:
I'm just started using this with opencode, using the default passthrough mode

Claude Sonnet 4.6 is *really* confused with the output of the Edit tool. I really don't know if this is an issue, or just a misconfiguration. From its thinking I could gather:

- It expected the output of Edit to be the edited content, not just an "Edit applied successfully.". I can see in [opencode source](https://github.com/anomalyco/opencode/blob/7a17925495ca61b76169737b226b834b61942586/packages/core/src/tool/edit.ts#L190) that it returns the replacement result.
- When calling multiple tools, only one would work (or in case of bash, it seems to get the results concatenated, as it reported). I can see that that *is* in the known limitations, but it still confuses the agent. Maybe add an instruction in the system prompt so it runs one tool at a time?

I'll test again with passthrough disabled (as soon as I have more tokens lol), but I'd really like to passthrough to work just fine.

Here some exerts from its thinking:

> The edit failed - the oldString didn't match exactly. Let me re-read the exact content of responder.ts around line 192.

> Wait, actually the edit to responder.ts applied successfully? The message says "Edit applied successfully." So it seems it was applied. But wait, the previous message said "Could not find oldString in the file." Let me think about what happened...

> Good, it seems some edits are being applied.

> It seems like my edits are being applied. However, I'm not getting direct read results back - instead I'm getting back responses saying "Edit applied successfully." Let me try to get a clear read of the current state of all files.

> I keep getting "Edit applied successfully" but I haven't seen the content of the file after all these edits.

> I'm getting very confused by this conversation. Each of my tool calls is being forwarded to a client, and I get back either the result of a previous tool call or "Edit applied successfully." I can't tell what state the files are in.

From here on, it started using bash to read files because it stopped trusting its own tools

> I still can't see the actual file contents. The responses I'm getting back ("Edit applied successfully") seem to be echoing back results from some queue of previously applied edits. 
> 
> I notice that in the conversation flow, when I've made multiple tool calls at once, I only get one result back at a time, which is causing massive confusion.

After three bash calls in a row to read files

> I can see the current state of all files. The output is mixed up (resolver.ts and visitante.ts are concatenated in the output), but I can see the content.

FelipeSharkao:
Testing with passthrough off for completeness, it of course has no problem editing files, tho, unfortunately but as expected, I can't see the edits or bash results in opencode.

[...]

Issue #496, opened 2026-05-07
Claude fails to use Read/Grep/etc. tools correctly


stefanpartheym:
Since 1-2 weeks im noticing increasingly often, that the communication between AI agent and OpenCode is not working properly at times.
I experienced situations, where Claude asked me to paste contents of a file, or run grep, etc. manually. 

Some recent examples: 

```
Thinking: I'm getting Edit confirmations when I actually requested Read operations—this seems like a tool relay issue. Either something triggered auto-edits or the system misunderstood my request, but either way, four edits apparently went through. I should check the git status to see what actually changed in the working tree.
Tool relay glitch — I asked for reads, system reports 4 edits applied. Verifying actual tree state before doing anything:
```

```
Thinking: I'm noticing a pattern where tool invocations return empty and I keep getting the AGENTS.md file back instead of actual results. It looks like the tools are being forwarded to the client, but then the client is sending back the documentation file as context rather than the tool's output. I should try delegating this research task to a Task subagent to run the exploration in parallel, which is what the AGENTS.md guidance recommends for open-ended work.
Tool execution seems to be getting interrupted. Let me delegate the exploration to a research agent which is the right approach here per the AGENTS guidance.
```

```
Thinking: I've got the SUT locked in, but I'm running into issues reading the spec mock setup and the user-settings.json fixture. The file paths I'm using keep getting rejected, so I'm going to try switching to absolute paths that I know worked before.
Let me try using grep to search for what I need instead.
```

```
Thinking: I'm noticing a pattern where the file reads keep returning the same content — it looks like there's a loop happening with how the tool calls are being intercepted. Let me try delegating this to a build subagent to isolate the issue, though I'm realizing the tool responses coming back might not match what I actually requested.

Tool channel is in a loop returning the same file. Delegating the full fix workflow to a build subagent — its own context, its own tool sequence:
...
```

I noticed, meridian outputs some huge log lines sometimes, that seem to get bigger with every prompt (Not sure if that has anything to do with this):

```
[...]

PR #875, opened 2026-08-25
fix(passthrough): rewrite forwarded denials before each resume so the model never sees them

CrazyCoder:
In passthrough mode, from the third turn of a session, the model is sent the hook's denial instead of the client's real result. This happens for every tool call made in an earlier turn. This PR rewrites that denial in the session transcript before each resume, so the model only sees the real result.

## What the model sees today

Passthrough forwards each tool call to the client. The PreToolUse hook denies the call inside the SDK. The CLI stores that denial in the session JSONL as the call's `tool_result` ("This tool call has been forwarded to the client for execution..."). The client runs the tool and sends the real result on its next request, which Meridian resumes at the assistant checkpoint (`resumeSessionAt`).

That resume cuts a suffix of the transcript. It removes the current turn's denial and never an earlier turn's. On the following resume the CLI's loader walks the `parentUuid` chain, then splices back every `tool_result` whose parent assistant is on the chain. That pass exists for parallel siblings and cannot tell a stale denial from one. The loader keeps the first `tool_result` per id, and the denial is older than the real result. The request body the model receives now carries the denial, and the real result is gone.

I measured this with a recording proxy on `ANTHROPIC_BASE_URL` (`scripts/probe-passthrough-accumulation.mjs`). On turn 3 of a three-call chain, the request body contained the denial for call 1 and no real result for it. Every single-turn probe measured clean, which is why the defect survived the existing suites: it needs three turns to appear.

I also observed it in OpenCode after a turn with parallel `grep`, `glob` and `read` calls. The model reported the results as fabricated, refused to quote output it had received, re-ran the calls, and retracted findings drawn from them. With this PR applied, OpenCode reports every result correctly across the same session.

## Issues that patched the symptom

The following reports describe this defect from the model's side. Each was answered by changing what the model is told, and the denial stayed in the transcript.

- #496: "tool invocations return empty", the model asks the user to paste file contents or run grep by hand.
- #552: the model is confused about which calls ran and reports "a file I never created". PR #580 answered it by rewording the deny text for dropped calls.
- #768: "the persisted `end your turn now` deny sits immediately before the continuation's user turn, and the model obeys it". PR #793 captured the boundary and added `SILENT_TURN_NUDGE`, which tells the model that the earlier deny is discharged.
- #689: the same "results never came back to me" report on the internal-mode resume path.
- #619, #111, #386: the replay envelope and the anti-imitation instruction. Related because the envelope's wording ("context only", "never invent tool output") made the model discard tool results it really received. The second commit here changes that wording.

`server.ts`, `messages.ts`, `passthroughEarlyStop.ts` and `envelopeIntegrity.ts` cite 43 distinct issue numbers. Their comments describe one family: the model denies making calls it made, misattributes results, invents tool output, imitates replayed transcripts. This PR removes the cause and touches none of those workarounds. Which of them are now unnecessary is a separate pass.

## What changes

`src/proxy/passthroughTranscript.ts` (new). Before a checkpoint resume that delivers real results, every forwarded denial whose id has a delivered result is rewritten in place with that result. No row is added or removed. No `uuid`, `parentUuid` or `last-prompt` leaf hint changes, because the loader starts its chain walk from that hint and deleting the row breaks the resume. Only the changed lines are re-serialized; every other line goes back byte for byte. The rewritten row also loses the CLI's row-level denial stamps (`toolDenialKind`, `toolUseResult`). A denial row is recognised as an error `tool_result` carrying `PASSTHROUGH_DENY_REASON`, which is now the single owner of that text. A client result that is itself an error is left alone. If the repair throws, the resume proceeds un-repaired and the turn does not fail.

A repair is only needed on a checkpoint resume. A passthrough tool turn with no checkpoint (early stop off, or its checkpoint refused) is evicted, never stored. The next turn is `lineage=new`, and the old transcript is never loaded again. `passthrough-transcript-integration.test.ts` pins both facts.

`src/proxy/server.ts`, `src/proxy/query.ts`, `src/proxy/passthroughEarlyStop.ts`. The non-stream path had no turn boundary. It released held denies on the first assistant message and froze the checkpoint there, which delivered 1 of 3 parallel calls against the real proxy. Passthrough now requests `includePartialMessages` on both paths and releases denies on `message_delta`. The checkpoint is gated on the completeness oracle the streaming path already used, `trackerCoversStreamedCalls`, which now has one owner instead of two copies. Both paths deliver 3 of 3. `DENY_HOLD_TIMEOUT_MS` is derived from `UPSTREAM_IDLE_MS` instead of a standalone 8 s. `guardUpstreamIdle` owns model-stream liveness, and the shorter timer released mid-generation and truncated the tool set.

`src/proxy/messages.ts` (second commit, independent). The fresh-session envelope now states that the replayed history happened and the tool results are genuine, before it forbids imitating the format. On a live OpenCode session the old wording made the model retract a pytest run, a file read, a glob and a grep as fabrications. It cited the warning as its reason.

## Evidence

`bun scripts/e2e-passthrough-turns.mjs [--stream]` (E41) runs an OpenCode-shaped client through the real proxy and the real SDK. The client makes three dependent reads, or three parallel reads with `PROBE_PARALLEL=1`, then one follow-up turn. The gate asserts three things. The final answer quotes all three results. No forwarded denial remains stored for a delivered id in any resumed session. Every continuation turn reads at least 95% of the previous turn's prompt cache. It needs Claude Max and costs real tokens.

Results on sonnet, 2026-08-25, all four combinations:

| Run | Turns | Quoted | Stale denials | Cache read per continuation |
|---|---|---|---|---|
| chain, non-stream | 4 + follow-up | 3/3 | 0 | 5859 = 3703+2156, 6053 = 5859+194, 6216 = 6053+163, 6369 = 6216+153 |
| chain, stream | 4 + follow-up | 3/3 | 0 | 5795 = 3703+2092, 5978 = 5795+183, 6207 = 5978+229, 6368 = 6207+161 |
| parallel, non-stream | 2 + follow-up | 3/3 | 0 | 5830 = 5355+475, 6299 = 5830+469 |
| parallel, stream | 2 + follow-up | 3/3 | 0 | 5772 = 5294+478, 6244 = 5772+472 |

Each continuation's `cache_read_input_tokens` equals the previous turn's `cache_read + cache_creation` to the token. That includes the follow-up turn, whose prompt the loader builds through the repaired rows. The repair costs no prompt cache. A mutation that writes one extra byte into the rewritten content freezes the follow-up's cache read at 5838, against an expected 6307. The assertion catches that defect.

Without the fix (`git stash push src/proxy/server.ts`), the chain run fails by the third turn: 0/3 quoted, 2 stale denials. The model answers "[Error: tool call forwarded to client, no content returned]" for the first two files.

Unit and integration suites: `passthrough-transcript`, `passthrough-transcript-integration`, `passthrough-early-stop`, `passthrough-early-stop-integration`, `proxy-nonstream-deny-hold`, `messages`, `proxy-replay-envelope`. 152 pass on this branch. `tsc --noEmit` is clean.

## Not in this PR

- Removal of the workarounds cited above. Candidates only; none is removed here.
- Runs on haiku and opus. All measurements are on sonnet.
- The `scripts/probe-passthrough-*.mjs` files print findings and never assert. They are the diagnostics that found the mechanism and are documented in E2E.md as such, separate from the gate.

### DIFF
--- a/E2E.md
+++ b/E2E.md
@@ -93,6 +93,7 @@ kill $(lsof -ti :3456)
 | E38 | [Silent turns (#768)](#e38-silent-turns-768) | **Automated**: `bun scripts/e2e-silent-turn.mjs` — real CLI, SSE mode. Asserts four things per attempt: the client got text or a tool call; recovered content sits BEFORE the terminal `message_delta` (content behind it is dropped by a correct client); exactly one `message_delta` per message; and a third turn after a recovery still resumes. Attribution is read from `/telemetry/logs`, not stdout. Pair `MERIDIAN_DEBUG_FORCE_SILENT_TURN=1` against `MERIDIAN_SILENT_TURN_RECOVERY=0` for the before/after. **Run before any release touching the passthrough tool loop, prompt assembly, or session resume** | 2026-08-11 |
 | E39 | [OpenCode internal-agent session key (#845)](#e39-opencode-internal-agent-session-key-845) | **Manual**, real OpenCode: its `title` agent runs under the USER'S session id, so the user's first turn used to queue behind it and then get HTTP 400 `session_turn_conflict`. Asserts the first turn succeeds, waits ~0ms on the session lease, and every later request is `lineage=continuation`. **Run after any OpenCode upgrade and before releases touching session keys or the turn coordinator** | 2026-08-19 |
 | E40 | [Passthrough digest-turn cap](#e40-passthrough-digest-turn-cap) | **Automated**: `bun scripts/e2e-digest-turn-cap.mjs` — real SDK. Asserts the capped tool turn generates no digest text, costs materially less than uncapped on an identical prompt, still RESUMES at its captured checkpoint, leaves text-only turns returning `success`, and does not truncate parallel tool calls. **Run before any release touching the passthrough tool loop, `maxTurns`, or the early-stop checkpoint** | 2026-08-20 |
+| E41 | [Passthrough multi-turn: one call, one answer](#e41-passthrough-multi-turn-one-call-one-answer) | **Automated**: `bun scripts/e2e-passthrough-turns.mjs [--stream]` — real proxy, real SDK, an OpenCode-shaped client replaying history with tool_results over a chain of dependent calls (`PROBE_PARALLEL=1`: one parallel turn). Asserts the model's final answer quotes every delivered result and never claims a call went unanswered, that no forwarded denial remains stored for a delivered id, and that every continuation turn reads the previous turn's prompt cache in full — the repaired rows must not cost a byte of prefix. **Run before any release touching passthrough resume, the deny hook, or `passthroughTranscript.ts`** | 2026-08-25 |
 
 | P1 | [Profile: List & Auth Status](#p1-profile-list--auth-status) | `/profiles/list` returns profiles with emails, login status, auth timestamps | - |
 | P2 | [Profile: Switch via API](#p2-profile-switch-via-api) | `POST /profiles/active` switches profile; health endpoint reflects new email | - |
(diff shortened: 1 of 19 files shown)


connor-grady:
Filed #877 for a residual path: on undo turns the anchor moves to the rollback uuid, `forkSession` is forced, and the converted messages carry no delivered `tool_result` blocks, so the repair here looks like it no-ops while the fork still persists earlier welded denials. Independent reproduction on 1.62.7 with the opencode adapter, details in the issue.

rynfar:
Superseded by #879, now merged as `2ccc48a5`.

Thank you, Serge. The complete stream/non-stream turn-boundary work and investigation were carried into #879 as the attributed selected derivative `8a095e8b`. The unsupported private JSONL rewrite was deliberately replaced with supported preallocated Agent SDK forks. The squash commit retains your credit with an explicit `Co-authored-by` trailer. Exact-final real SDK and OpenCode validation passed.

Issue #744, opened 2026-08-02
claude-code clients are told the proxy's working directory, so absolute paths write to the wrong tree


rynfar:
## Summary

A `claude-code` client is told the **proxy host's** working directory, not its own. The model then composes absolute paths from it, so a tool call can write to the wrong machine's tree.

Found while verifying #743. Pre-existing and unrelated to that change — see "Not caused by #743" below.

## Reproduced

Real Claude Code CLI pointed at Meridian, client running in a scratch directory:

```
$ cd /private/tmp/.../scratchpad/cc-cwd
$ ANTHROPIC_BASE_URL=http://127.0.0.1:PORT claude -p \
    "Do not use any tools. In one line: what is your current working directory, exactly as your context states it?"

My current working directory, exactly as my context states it, is `/Users/rynfar/repos/meridian`.
```

`/Users/rynfar/repos/meridian` is where the **proxy** runs. The client was in `.../scratchpad/cc-cwd`.

The practical consequence, from an earlier run in the same setup:

```
TOOL_USE: Bash {"command": "echo BANANA >> /Users/rynfar/repos/meridian/log.txt"}
TOOL_RESULT is_error=False
```

The model was asked to append to `log.txt` in its own directory. It built an absolute path from the working directory it had been told, and the line landed in the proxy's repo instead. The tool reported success, and the file the user asked about was never touched.

## Mechanism

`claudeCodeAdapter` deliberately returns `undefined` from `extractWorkingDirectory` (so the SDK subprocess chdirs somewhere valid on the server) and is supposed to surface the client's real path separately, via `extractClientWorkingDirectory` → `buildCwdNote` → a `<meridian-note>` in the system prompt.

That note is absent here. `extractClaudeCodeClientCwd` (`src/proxy/adapters/claudecode.ts:38`) parses the system prompt for:

```ts
systemText.match(/Primary working directory:\s*([^\n<]+)/i)
```

If the client's system prompt doesn't carry that exact phrasing, the parse returns `undefined`, `buildCwdNote` short-circuits to `""`, and the model is left with the SDK's own env block — the proxy path.

Worth confirming which it is before fixing:

1. the CLI version under test doesn't emit that line at all, or
2. it emits it somewhere the extractor doesn't look (e.g. a later `system` block, or different wording).

Either way a silent `undefined` is the wrong failure mode for something this consequential.

## Not caused by #743

`server.ts:853` computes `clientWorkingDirectory` unconditionally — it does not branch on passthrough:

[...]

PR #876, opened 2026-08-25
fix(cwd): disown the CLI's own environment lines, not only the path

CrazyCoder:
When the SDK subprocess and the client run in different directories, the model gets two environment blocks. Only one line of the stale block is disowned. The model reads "Is a git repository: false" from the proxy host as true of the user's machine. It then reports the mismatch to the user as a contradiction.

## What the model sees

The CLI builds its own environment block from the subprocess cwd and emits it on every request, whatever the system prompt already contains:

```
Primary working directory: /app
Is a git repository: false
```

`buildCwdNote` (#744, PR #746) appends an `<env>` block with the client's directory, plus a note that disputes the path. Its comment says the appended block stops the CLI from injecting its own. It does not. The CLI builds that block in `computeEnvInfo` from `getCwd()` and `getIsGit()`, and it reaches the model anyway, as the transcript below shows.

So the model receives both, and the note names only one of the two lines. Measured against a proxy running in `/app` with a client in `C:\projects\example-app`, asked to quote every environment line it can see:

```
[Environment block]
"Primary working directory: /app"
"Is a git repository: false"

[env block]
"Working directory: C:\projects\example-app"

[meridian-note]
"... resides at "/app" on the proxy host, but that is not the user's working directory."
```

The model picked the client directory for paths, which is what the note asks for. It then told the user that the environment block reports `/app` and no git repository, while the real directory is a repository. That second claim comes from a line no one had disowned.

This matters beyond the confusing message. `Is a git repository: false` gates behaviour. A model that believes it is outside a repository does not offer to commit, and does not run `git status` before a destructive command.

## The change

The note now names both of the CLI's lines as proxy-host facts and states the client's directory. It also says where the git answer comes from: the client's own environment block when it sends one, otherwise `git status`. One test covers that.

After the change, the same probe on the same proxy:

```
1. Working directory: C:\projects\example-app
2. Yes, it's a git repository
3. I'm reached through a proxy - the subprocess backing me runs at /app on the
   proxy host, but that's irrelevant to you; all file/path operations refer to
   your machine above.
```

## One wording constraint

An earlier revision of this note ended with "so there is no discrepancy to resolve or report". The model then treated the whole `<meridian-note>` as injected content and said so to the user. It read that sentence as an instruction to explain away a mismatch. The shipped wording states that the two blocks describe two different machines, and stops there. A note that tells the model what not to say to the user reads as manipulation, whoever wrote it.

## Scope

`buildCwdNote` returns an empty string when the client directory is absent or equals the SDK cwd, so a same-host client is unaffected. Only the note text changes; no call site, no control flow.

Verified on the prod image with the plugins mount: 9 of 9 harness cases pass, and the probe above is the deployed behaviour. Unit suites `query-cwd-note`, `opencode-cwd-note`, `claude-code-cwd-note`, `query-git-status-note` and `pi-adapter`: 80 pass. `tsc --noEmit` is clean.

### DIFF
--- a/src/__tests__/query-cwd-note.test.ts
+++ b/src/__tests__/query-cwd-note.test.ts
@@ -31,6 +31,19 @@ describe("buildCwdNote", () => {
     expect(note).toContain("</meridian-note>")
   })
 
+  // The CLI always emits its own environment block from the subprocess cwd,
+  // so the note must disown BOTH of its lines by name, not only the path:
+  // left unmentioned, "Is a git repository: false" is taken as true and the
+  // model reports the contradiction to the user.
+  it("names the CLI's own working-directory and git-repository lines as proxy-host facts", () => {
+    const note = buildCwdNote("/app", "C:\\projects\\example-app")
+    expect(note).toContain('"Primary working directory: /app"')
+    expect(note).toContain('"Is a git repository: ..."')
+    expect(note).toContain("describes that host, not the user's machine")
+    expect(note).toContain("client's own environment block")
+    expect(note).toContain("two different machines")
+  })
+
   it("places the <env> override before the meridian-note so the subprocess sees it first", () => {
     const note = buildCwdNote("/srv/proxy", "/Users/alice/app")
     const envIdx = note.indexOf("<env>")
--- a/src/proxy/query.ts
+++ b/src/proxy/query.ts
@@ -240,26 +240,33 @@ function computePassthroughMaxTurns(
  * Build an addendum that tells the model which path belongs to the real user.
  * Applied when the SDK subprocess runs in one directory on the proxy host but
  * the client is working in a different directory on their own machine
- * (typical of a remote Claude Code → network-proxy setup). Without this note
- * the SDK's env block leaks `sdkCwd` into the model's context and Claude
- * reports that as its working directory.
+ * (typical of a remote Claude Code → network-proxy setup).
+ *
+ * The CLI computes its own environment block from the subprocess cwd and
+ * always emits it, whatever the system prompt already contains: "Primary
+ * working directory: <sdkCwd>" and "Is a git repository: <bool>" describe the
+ * proxy host, and nothing the proxy appends suppresses them. The note has to
+ * name both lines, or the model accepts the git-status line as true, reports
+ * the contradiction to the user, and reasons about the wrong tree.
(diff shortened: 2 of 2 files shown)


CrazyCoder:
See also https://github.com/rynfar/meridian/pull/873.

rynfar (changes requested):
The underlying stale proxy-host git fact is real, and the note should disown it. The current shared wording is not safe enough to merge.

Blocking issue: `buildCwdNote` is shared by all adapters, but the new text assumes the client has its own OpenCode/Claude-style environment block and says “otherwise `git status` tells you.” For a remote Droid/ForgeCode client, or OpenCode/Pi with passthrough disabled, Meridian's Bash/MCP tool can execute that command on the proxy in `sdkCwd`. That reports the same stale proxy-host fact this patch is trying to disown. Please make the wording agent-neutral and avoid an unqualified command, or pass explicit execution-locus/passthrough information into the helper and only recommend a check known to run on the client.

Additional required proof:

1. Rebase on current main and rerun checks; this head predates #879/#880.
2. Add current HTTP/`buildQueryOptions` integration using captured OpenCode V1 and exact supported V2-beta request shapes. Cover stream/non-stream, resume, `clientSystemPrompt:false`, and the workdir-override case from #873.
3. Cover passthrough and non-passthrough remote execution explicitly, so a future wording change cannot send repo inspection to the wrong host.
4. Handle path identity conservatively. Lexical equality of `/app` is not proof that two containers are one machine, while Windows case/separator/trailing-slash variants may describe one path. Avoid categorical “two different machines” claims without evidence.
5. Update the stale `query-git-status-note` test comment that repeats the same false premise, and add adversarial quote/tag/path tests for the inherited client-path interpolation surface.
6. If the text remains agent-specific, add the required `NOTE:` marker or move the policy behind the adapter. Prefer shared, agent-neutral wording.

The direct unit test currently proves authored substrings only; it does not prove production wiring or execution behavior. Please revise rather than close—the core correction is useful once its execution-locus assumptions are removed.

CrazyCoder:
Superseded by #885. The replacement combines this correction with #873, removes the unsafe `git status` instruction and machine claim, distinguishes tool execution loci, normalizes path identity, escapes interpolated paths, and adds the requested integration coverage.

PR #879, opened 2026-08-26
fix(passthrough): make durable resume crash safe

### LINKED ISSUE #827: SDK semaphore permit is released before the subprocess actually exits
Split out of #825 (review of #821).

`runSdkQueryAttempt` releases the semaphore permit in its `finally`, immediately after calling `sdkQuery.close()`:

```ts
if (typeof sdkQuery?.close === "function") sdkQuery.close()
...
lease.release()
```

`close()` is fire-and-forget. The Agent SDK's process transport waits ~2s before SIGTERM and up to ~5s more before SIGKILL, so the child can still be alive when the permit is handed to the next waiter.

**Impact:** on the abort/idle-guard path a wave of cancellations can hold old *and* newly-admitted subprocesses alive at once, so the actual process count transiently exceeds `MERIDIAN_MAX_CONCURRENT` — precisely when the host is already under stress. The code's own comment near the release acknowledges the exit window.

**Possible fix:** await a bounded disposal (e.g. `await Promise.race([closed, timeout(N)])`) before `lease.release()`, so the permit tracks the subprocess rather than the intent to kill it.

Not fixed in #825 because it needs empirical confirmation of the SDK transport's teardown timing and a test that can observe subprocess lifetime rather than generator completion.

### LINKED ISSUE #828: Forced shutdown awaits HTTP teardown, not request settlement
Split out of #825 (review of #821).

`closeServerWithGracePeriod` destroys sockets and awaits only `server.close()`. `ProxyInstance.close()` does not abort or await the per-request controllers, and does not wait for `getInFlightCount()` to reach 0. The CLI then calls `process.exit(0)` as soon as `close()` resolves.

**Impact:** SDK generator cleanup and session persistence are async, so on the hard-deadline path the process can exit before subprocess cleanup and session writes have settled — orphaned children or a lost session store.

**Possible fix:** on the deadline, abort the tracked request controllers and await `getInFlightCount() === 0` (bounded) before resolving, so `process.exit(0)` is only reached once work has actually stopped.

Deferred from #825 to keep that PR's scope to the correctness bugs it could prove with tests.

### LINKED ISSUE #877: Undo welds passthrough denial records onto the live chain and persists them into the forked session
**Environment:** 1.62.7, opencode adapter, passthrough.

**Summary:** On an undo turn the resume anchor moves to the rollback point, `forkSession` is forced, and the fork writes out a chain in which previously-orphaned `PASSTHROUGH_DENY_REASON` records sit between an assistant tool call and its real results. From that write onward the denial is persisted state rather than a per-load reconstruction, and every later fork copies it forward.

**What I saw:** the first two turns were clean, real tool output on the live chain and the denials orphaned as intended. After an `/undo`, the session file written by the fork has the same user message that carries the real results attached to a different parent:

```text
before undo:  REAL <- ASSISTANT_TOOL_CALL   (denial orphaned as a sibling)
after  undo:  REAL <- DENIAL                (denial on the live chain)
```

(uuids replaced with labels). One welded link, on the conversation's first tool call, was enough to poison every subsequent turn of that session.

**Why undo specifically:** `src/proxy/server.ts:1535-1538` collapses `messagesToConvert` to the last user message when `undoRollbackUuid` is set, so the turn carries no `tool_result` blocks. `server.ts:2113` (and `:2990` for the stream path) then resumes at `undoRollbackUuid ?? passthroughToolCallAssistantUuid`, and `src/proxy/query.ts:474` forces `forkSession: true` for any `isUndo` turn. The anchor is therefore downstream of earlier turns' orphaned denials, and the fork serialises them into the new file.

**Relation to #875:** that PR repairs denials before a resume, but its call site is guarded on `resumeSessionId && passthroughToolCallAssistantUuid` and operates on delivered results found in the converted messages. On an undo turn there are no delivered results to key off, and if the previous turn's checkpoint never settled the anchor is null and the guard skips entirely. Undo looks like the residual path even with #875 applied. #871 targets the same failure with a different remedy, so whichever lands should be tested against this case.

**Severity note:** the denial text ends with "Do not retry, do not call additional tools, and do not generate further text — end your turn now". When it resurfaces as the recorded outcome of an earlier call, the model reads it as proof that tool calls never execute. In my session it retracted correct, already-delivered findings as hallucinations and re-issued the same calls, which then hit the exact-duplicate guard and returned nothing, reinforcing the false conclusion each round. The non-error superseded marker chosen in #875 is the right shape for this reason: the failure is instruction-shaped, not merely a missing result.

rynfar:
## Summary
- preserve the passthrough fixes and contributor work from #871 and #875
- replace private transcript mutation with supported preallocated Agent SDK sessions and isolated forks
- add durable generation fencing, cross-process turn coordination, exclusive writer leases, crash recovery, and supported bounded deletion
- complete stream/non-stream checkpoint handling, undo/full-replay safety, cancellation/shutdown fencing, and cross-proxy resume
- ignore validated transient OpenCode `UserPromptSubmit` hooks for lineage only, while leaving the model/tool request unchanged

## Safety invariants
- never read or rewrite Claude private JSONL transcripts
- every Meridian-created transcript is journaled before SDK spawn
- every resume writes only to a fresh supported fork and publishes it by exact CAS
- terminal tool/structured responses follow durable publication or exact source eviction
- active/unjoined SDK writers and uncertain deleters remain durably fenced
- lineage canonicalization preserves message count/order and applies resume indexes to the same canonical block list

## Validation
Implementation SHA `73a966f2` passed:
- local full `npm test`: 2,834 main-phase tests plus every isolated follow-up
- `npm run typecheck`, `npm run build`, and PR-range `git diff --check`
- four-case real Agent SDK matrix: stream/non-stream × parallel/serial, with the real SDK process gate enabled
- actual headless OpenCode 1.18.11: real read, consecutive ordinary continuations, three parallel reads, supported revert/undo, post-undo continuation, and two real Meridian restarts
- 97–100% prior-prefix cache reuse on supported forks; 98% on safe undo replay
- `MERIDIAN_MAX_STORED_SESSIONS=1`: one mapping with exactly current plus direct predecessor
- supported SDK GC: 10 deleted, 0 not found, 0 failed, 0 deferred; all history checks used `getSessionMessages()`
- hosted Linux, actual `windows-latest`, and Docker: https://github.com/rynfar/meridian/actions/runs/33029475400 and https://github.com/rynfar/meridian/actions/runs/33029475408

## Attribution
- Aleksey Proshutinskiy's `b35362a7`, `f82b2606`, and `f798f362` are patch-identical cherry-picks of #871 commits `7ab0768e`, `ca19201c`, and `ea907bfb`.
- Serge Baranov's `8a095e8b` is an attributed selected derivative of #875 commit `394e3002`; it is not claimed as patch-identical because the unsupported private-transcript strategy was deliberately excluded.
- Because this PR will be squash-merged, contributor credit must survive in the generated commit through explicit `Co-authored-by` trailers for both contributors.

Closes #877
Closes #827
Closes #828


rynfar:
## Final exact-head verification

**Final branch head:** `7953d9f74f0888ac9fa42e45851a13be2534c7bb`  
**Implementation head:** `73a966f2ea4421db6a3046aeec2c80200c924d9d`

### Automated gates
- Local full suite: 2,834 main-phase tests passed; all isolated follow-ups passed
- Typecheck and build: passed
- Exact-final hosted Linux + actual Windows: https://github.com/rynfar/meridian/actions/runs/33030148844
- Exact-final Docker: https://github.com/rynfar/meridian/actions/runs/33030148833
- PR-range `git diff --check`, clean worktree, remote-head, private-path, and added-unsafe-cast audits: passed

### Real Agent SDK
All four stream/non-stream × parallel/serial cases passed with the real SDK process gate enabled. Every continuation used a distinct supported fork, read the prior cache prefix, and the active history contained exactly one real answer per delivered tool call and no forwarded denial.

### Actual OpenCode 1.18.11
- real read: `OPENRESULT[exact-73a966f2-proof]`, supported fork, 99% cache
- consecutive ordinary turns: `CONTONE` (99%) and `CONTTWO` (100%), proving transient `UserPromptSubmit` hooks do not break lineage after becoming historical
- full Meridian restart: `CROSSPROXY`, durable continuation, 100% cache
- exactly three parallel reads: `PARALLEL[alpha,bravo,charlie]`, supported fork, 97% cache
- supported revert: `prefix overlap 4/9`, `lineage=undo session=new`, fresh prepared transcript, `UNDONE`, 98% cache
- post-undo continuation: `AFTERUNDO`, 100% cache
- second restart with `MERIDIAN_MAX_STORED_SESSIONS=1`: `BOUNDED`, 100% cache; one mapping retained exactly current + predecessor

### Supported SDK GC
All transcript history checks used `getSessionMessages()`. GC used supported `deleteSession()` semantics:
- 10 deleted
- 0 not found
- 0 failed
- 0 deferred
- deleted histories returned zero messages; pinned current and predecessor retained 9 and 6 messages

### Provenance
- Aleksey Proshutinskiy's three commits are patch-identical cherry-picks from #871.
- Serge Baranov's commit is an attributed selected derivative of #875, not claimed as patch-identical.
- The squash merge body will contain explicit `Co-authored-by` trailers for both contributors.

## Merge result

Squash-merged to `main` as `2ccc48a5d75514b490420b2c6a735a5983107f7e`. The merge tree is byte-identical to verified final branch head `7953d9f7`. Explicit co-author trailers for Aleksey Proshutinskiy and Serge Baranov are present in the merge commit. Post-merge Linux, actual Windows, build, and changelog checks passed: https://github.com/rynfar/meridian/actions/runs/33030438901. Post-merge Docker passed: https://github.com/rynfar/meridian/actions/runs/33030438894.

PR #882, opened 2026-08-27
fix(models): use configured profile auth for context window

justprosh:
## Summary

- resolve the active/configured profile before building `/v1/models`
- pass that profile's id and environment to the Claude auth probe
- preserve the legacy auth context when no profiles are configured

## Why

The model catalog derives its advertised context window from the authenticated subscription. With profiles configured, the endpoint still probed the process-global auth context, so it could advertise 200k for a routed Max profile or 1M for a profile that does not include extended context.

Using the same profile auth context as routing makes discovery reflect the account that will serve requests.

## Validation

- `bun run typecheck`
- `bun test src/__tests__/proxy-health-build.test.ts` (8 passed)
- coverage for legacy, Max 1M, and non-Max 200k contexts
- `git diff --check`


rynfar:

PR #883, opened 2026-08-27
fix: forward SSE keepalive comments in OpenAI-compat streams

justprosh:
## Summary

- forward SSE comment lines from the internal `/v1/messages` stream through `/v1/chat/completions`
- forward the same comments through `/v1/responses`
- cover both adapters while preserving their translated data frames and completion events

## Why

The internal messages stream emits `: ping` every 15 seconds, but both translating adapters filtered every line that did not start with `data: `. During a long thinking phase this could leave a downstream client without any bytes until content appeared, allowing a normal idle timeout to sever a live stream.

SSE comments are valid keepalive frames and are ignored by event consumers, so relaying them preserves the existing response semantics while keeping the connection active.

## Validation

- `bun run typecheck`
- `bun test src/__tests__/proxy-openai-compat.test.ts src/__tests__/proxy-responses-endpoint.test.ts` (42 passed)
- `git diff --check`

### DIFF
--- a/src/__tests__/proxy-openai-compat.test.ts
+++ b/src/__tests__/proxy-openai-compat.test.ts
@@ -501,6 +501,50 @@ describe("POST /v1/chat/completions — streaming", () => {
     expect([...uniqueIds][0]).toMatch(/^chatcmpl-/)
   })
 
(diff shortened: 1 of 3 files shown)


rynfar:

PR #884, opened 2026-08-27
fix(docker): initialize machine identity in runtime image

CrazyCoder:
## Summary

- initialize `/etc/machine-id` in the Alpine runtime image
- validate its 32-character hexadecimal format during the image build
- make the identifier read-only for the runtime user

## Problem

The durable process-incarnation fencing added in #879 requires a Linux
machine ID. `node:22-alpine` provides neither `/etc/machine-id` nor
`/var/lib/dbus/machine-id`, so `captureProcessIncarnation()` returns
undefined. Every `/v1/messages` request then fails with HTTP 500:

```
cannot capture lock owner process incarnation
```

The existing Docker workflow only builds the image, so it does not
exercise this request path.

## Verification

Tests were not run locally. The production failure was reproduced with
a direct request and the OPNsense real-request liveness probe. Adding a
valid machine ID to the running container restored both to HTTP 200.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,12 @@ RUN deluser --remove-home node 2>/dev/null; \
     && mkdir -p /home/claude/.claude \
     && chown -R claude:claude /home/claude
 
+# Alpine does not provide /etc/machine-id. Durable process-owner fencing needs
+# one to distinguish lock owners safely, so create it once in the image layer.
+RUN node -e "process.stdout.write(require('node:crypto').randomBytes(16).toString('hex') + '\n')" > /etc/machine-id \
+    && grep -Eq '^[0-9a-f]{32}$' /etc/machine-id \
+    && chmod 0444 /etc/machine-id
+
 USER claude
 WORKDIR /app
 


CrazyCoder:
The Linux job failed only in `proxy-subagent-support.test.ts` during its concurrent-request case. The Docker build, Windows smoke job, and 2,885 other tests passed. This Dockerfile-only PR does not touch request concurrency.

I cannot rerun upstream Actions as a fork contributor, so I am closing and reopening the PR to request a fresh run.

diwakar-s-maurya:
Using bash to generate the id will be better. I have tested this.

```Dockerfile
RUN cat /proc/sys/kernel/random/uuid | tr -d '-' > /etc/machine-id \
    && chmod 0444 /etc/machine-id
```

rynfar:
Thanks — this was exactly right, and it landed as your commit in #903 (`Co-authored-by` on the squash, so the credit carries onto `main`).

For context on why it came in through a new PR rather than merging here: `main` requires signed commits, so fork PRs sit at `mergeStateStatus: BLOCKED` with every check green and no visible reason. Nothing you could fix from your side. Cherry-picking preserves you as `Author`.

I added two commits on top. One makes the Docker workflow actually serve a request — it only built the image before, which is why a container that 500s on every request shipped to ghcr.io with CI fully green. The other gates `build-push` on that smoke job; my first attempt ran them concurrently, so a tag push would have published `:latest` while the smoke test was still starting.

Review also caught one thing worth flagging on your diff: the build-time `grep -Eq '^[0-9a-f]{32}$'` is line-scoped, but `linuxLocalBootIdentity()` reads the whole file, `trim()`s it, and tests `^[0-9a-fA-F]{32}$`. A file with a banner line plus a valid id passes the build and still 500s at runtime — I verified it. The smoke job now checks it the way the runtime does. Your instinct to validate at build time was the right one; it just needed to match the consumer.

@diwakar-s-maurya — your `/proc/sys/kernel/random/uuid` suggestion is equally correct, and it is what systemd itself does. I kept this version only because it validates the result; both draw from a CSPRNG and both satisfy the consumer regex.

Two deeper issues are filed separately: `hostId` is neither stable across `docker restart` nor unique across containers from the same image, which the machine-id alone cannot fix. Thanks again — this one was breaking every containerized user.

PR #894, opened 2026-08-28
fix: preserve cached input usage in OpenAI responses

justprosh:
## Summary
- preserve Anthropic input and cache usage across streaming message_start/message_delta events
- report full OpenAI prompt/input totals with cached and cache-write details
- apply the same accounting to Chat Completions and Responses, streaming and non-streaming

## Tests
- npm test
- npm run typecheck
- npm run build
- git diff --check


rynfar:

PR #896, opened 2026-08-29
fix: run session garbage collection on Windows

### LINKED ISSUE #895: Session-cache doesn't get garbage-collected on Windows
I started getting this error within an hour of setting up Meridian behind Pi, when natively installed on Windows:

```
Error: 500 {"type":"error","error":{"type":"api_error","message":"session transcript ownership backlog is full"}}
```

My agent pointed out that I could raise the limit with `$env:SESSION_GC_MAX_PENDING = "2048"; meridian`; but claims the issue will persist until it's fixed upstream, or that I'll have to manually wipe out the `session*.json` to clear the error each time it fills up.

Hopefully this is helpful, I have not personally investigated deeply.

amasover:
Fixes #895

## Problem

Session GC is a no-op on Windows because fenced deletion assumes POSIX process groups. Retired transcripts accumulate until the pending backlog fills.

## Fix

- Run fenced SDK deletion on Windows.
- Use `taskkill /PID <pid> /T /F` for timed-out deletion trees.
- Fence PID reuse during join and crash recovery.
- Add Windows regression coverage and run it in CI.

## Testing

- Session lifecycle tests pass.
- Typecheck passes.
- Build passes.

I also tested it locally on my Windows machine and it worked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,13 @@ jobs:
       - run: bun test src/__tests__/durable-file-system.test.ts src/__tests__/windows-durable-storage.test.ts
         shell: bash
 
+      # Session GC's fenced deletion path was a silent no-op on win32 until it
+      # was ported. The ubuntu job runs this file too, but the win32-only
+      # branches (handle-based join, taskkill tree kill, pid-reuse-proof
+      # deleting-claim recovery) only execute on a real Windows runtime.
+      - run: bun test src/__tests__/session-lifecycle-windows-gc.test.ts
+        shell: bash
+
       # Plugin loader exercises `await import(absolutePath)` which on Windows
       # rejects raw backslash paths with `Received protocol c:`. The test
       # writes a real plugin file to a tempdir and loads it, so it actually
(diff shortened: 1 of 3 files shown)


amasover:
Merged latest main into the branch and fixed the conflict in ci.yml. I'd love a review on this if you have time @rynfar, I've been daily driving it for the last couple of weeks and didn't get any errors.

The very latest commit was an "extra fix" according to Claude:

>  **Why 7fe1aca**

> Session GC deletes transcripts from a spawned child process. The child isn't allowed to start until the parent has recorded its identity (pid + OS start id) in the sidecar, so that a crash mid-delete leaves something another process can resolve. The parent signals "recorded, go ahead" by writing a gate file; the child polls for it and exits 75 if it never shows up.

> Capturing that identity on Windows means shelling out to PowerShell, budgeted 10s. We were handing the child the deletion timeout as its gate deadline, so any timeout shorter than the PowerShell call meant the child gave up before the parent could open the gate. Every deletion then fails, and `prepareFork` eventually throws "ownership backlog is full".

> Fix: gate deadline = deletion timeout + probe budget, via a new `processIncarnationProbeBudgetMs()` export (10s win32, 2s elsewhere).

> The full local suite caught this as a flaky failure of `runGc tree-kills and joins a timed-out deletion child`. At a 100ms budget it fails every time, the child dying mid-probe and the parent reporting `cannot capture session deletion executor incarnation`.

rynfar:
sorry for the delay yes i will get this reviewed today and merged.  I just
hadnt had a chance to run this on my windows machine.  I can get this done
today.

PR #898, opened 2026-08-29
feat(openai): support structured output via response_format

diwakar-s-maurya:
## Problem

`/v1/chat/completions` dropped `response_format` entirely. A request asking for schema-valid JSON got prose back. It did not support [structured output](https://developers.openai.com/api/docs/guides/structured-outputs?example=structured-data) well.

```bash
$ curl -s localhost:3456/v1/chat/completions -H "Authorization: Bearer $KEY" \
    -d '{"model":"claude-sonnet-5","max_tokens":200,
         "messages":[{"role":"user","content":"Give me a person with name and age."}],
         "response_format":{"type":"json_schema","json_schema":{"name":"person","schema":{...}}}}'

{"choices":[{"message":{"content":"Here's a person for you:\n\n**Name:** Sarah Johnson\n**Age:** 32"}}]}
```

The Anthropic spelling `output_config.format` was ignored at this endpoint too. Structured output already worked correctly on native `/v1/messages`. Only the OpenAI-compatible route was missing the wiring.

## Fix

Translate `response_format` into the `output_config.format` that the internal `/v1/messages` hop already understands. Enforcement is the SDK's native structured output - schema validation and retry included, so this is plumbing into a working path rather than a second implementation. Streaming works the same way: the SDK buffers until the result validates, then emits it as one delta.

| Request | Before | After |
|---|---|---|
| `response_format: json_schema` | prose, 200 | schema-valid JSON, 200 |
| same, `stream: true` | prose | valid SSE, JSON delta |
| `response_format: json_object` | silently ignored | 400 |
| `response_format` + `tools` | silently ignored | 400 |
| `response_format: text` | 200 | 200 (unchanged) |
| native `/v1/messages` | 200 | 200 (unchanged) |

### Design decisions

**`json_object` is rejected, not approximated.** Anthropic has no schema-less JSON mode. Widening it into a permissive `{"type":"object"}` schema would promise an enforcement the request never gets, so it fails with an actionable
message instead.

**`strict` and `name` are dropped.** The SDK always validates, which is never weaker than `strict` asked for; `name` is a client-side label.

**Errors name the field the caller actually sent.** `parseOutputFormat` takes a dialect, and the OpenAI route validates before the hop - otherwise a client sending `response_format` is told about `output_config.format`, a field it never wrote. Clients using the Anthropic spelling at this endpoint still get `output_config.*` errors.

```console
$ ... "response_format":{"type":"json_object"}
{"error":{"message":"response_format.type: Only 'json_schema' is supported"}}

$ ... "output_config":{"format":{"type":"json_object"}}
{"error":{"message":"output_config.format.type: Only 'json_schema' is supported"}}
```

## Also fixed: `$schema` caused HTTP 500

A schema carrying a root-level `"$schema"` set to anything but the draft-07 URI returned:

```json
{"error":{"message":"Structured output was requested but the SDK returned no structured_output result"}}
```

The model never submitted its result.

**zod v4's `z.toJSONSchema()` emits the 2020-12 URI by default**, so every schema from the zod / Vercel AI SDK toolchain hit this. It is not a draft-version issue: 2020-12 (both spellings), 2019-09, and an arbitrary string all failed; only draft-07 or an absent key worked. **Both endpoints were affected**, since `/v1/chat/completions` forwards the schema to the same handler.

| `$schema` | Before | After |
|---|---|---|
| *(absent)* | 200 | 200 |
| `http://json-schema.org/draft-07/schema#` | 200 | 200 |
| `https://json-schema.org/draft/2020-12/schema` | **500** | 200 |
| `http://json-schema.org/draft/2020-12/schema` | **500** | 200 |
| `https://json-schema.org/draft/2019-09/schema` | **500** | 200 |
| arbitrary string | **500** | 200 |

The keyword only declares which dialect a schema is written in and constrains nothing, so it is dropped rather than translated. Only the root is touched: nested occurrences already round-trip fine. The strip lives in `parseOutputFormat`, the single funnel both endpoints pass through, so no path reaches the SDK unnormalized.

## Testing

`bun run test` - 160 tests pass across the affected files, typecheck clean.
Adds 9 tests: translation mapping, both error vocabularies, tools rejection, and `$schema` handling (root stripped, nested preserved, caller's object not mutated).

Verified end-to-end against a live proxy: nested schemas with `enum` and `integer` constraints are still fully enforced, tool-calling without `response_format` is unaffected, and native `/v1/messages` is unchanged.

## Not addressed

- **`/v1/responses`** has the same gap - it silently drops `text.format`. That is a different request shape and warrants its own change; happy to follow up.
- **Numeric bounds** (`minimum`/`maximum`) are not enforced - upstream moves them into the property description. This matches direct Anthropic behavior and is not a Meridian issue.

### DIFF
--- a/src/__tests__/openai.test.ts
+++ b/src/__tests__/openai.test.ts
@@ -732,6 +732,68 @@ describe("translateOpenAiToAnthropic", () => {
     expect(result!.tools).toHaveLength(1)
     expect(result!.tools![0]!.description).toBe("")
   })
+
+  // --- structured output ---
+
+  const personSchema = {
+    type: "object",
+    properties: { name: { type: "string" } },
+    required: ["name"],
+  }
+
+  it("maps response_format json_schema to output_config.format", () => {
+    const result = translateOpenAiToAnthropic({
+      messages: [{ role: "user", content: "hi" }],
+      response_format: {
+        type: "json_schema",
+        json_schema: { name: "person", schema: personSchema, strict: true },
+      },
+    })
+    // `name` and `strict` have no Anthropic equivalent and are dropped: the
+    // SDK always validates, which is never weaker than strict asked for.
+    expect(result!.output_config).toEqual({
+      format: { type: "json_schema", schema: personSchema },
+    })
+  })
+
+  it("keeps response_format alongside output_config.effort", () => {
+    const result = translateOpenAiToAnthropic({
+      messages: [{ role: "user", content: "hi" }],
+      output_config: { effort: "high" },
+      response_format: { type: "json_schema", json_schema: { schema: personSchema } },
+    })
+    expect(result!.output_config).toEqual({
+      effort: "high",
+      format: { type: "json_schema", schema: personSchema },
+    })
+  })
+
+  it("passes through an Anthropic-shaped output_config.format unchanged", () => {
+    const format = { type: "json_schema", schema: personSchema }
+    const result = translateOpenAiToAnthropic({
+      messages: [{ role: "user", content: "hi" }],
+      output_config: { format },
+    })
+    expect(result!.output_config).toEqual({ format })
+  })
+
+  it("forwards json_object intact so the boundary can reject it", () => {
+    // Not widened into a permissive schema: Anthropic has no schema-less JSON
+    // mode, and accepting it here would promise an enforcement never applied.
+    const result = translateOpenAiToAnthropic({
+      messages: [{ role: "user", content: "hi" }],
+      response_format: { type: "json_object" },
+    })
+    expect(result!.output_config).toEqual({ format: { type: "json_object" } })
+  })
+
+  it("treats response_format text as no constraint", () => {
+    const result = translateOpenAiToAnthropic({
+      messages: [{ role: "user", content: "hi" }],
+      response_format: { type: "text" },
+    })
+    expect(result!.output_config).toBeUndefined()
+  })
 })
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


fwmarcel:
Hit this exact failure mode today in production, wanted to add a real-world repro for context.

**Setup:** Meridian `1.66.0` (official `ghcr.io/rynfar/meridian` image) behind the OpenAI adapter, fronting a [Mealie](https://mealie.io/) instance that uses it as its `OPENAI_BASE_URL` for AI recipe/image features.

Mealie's Python OpenAI SDK client calls `client.chat.completions.parse(...)` (Structured Outputs) to validate the connection and to parse AI responses into a strict Pydantic schema. Against Meridian this comes back as plain prose instead of schema-valid JSON, exactly as described here:

```
ERROR - OpenAI Request Failed. ValidationError: 1 validation error for OpenAIText
  Invalid JSON: expected value at line 1 column 1 [type=json_invalid, input_value="Hello! Yes, I can see yo...nything else with this.", input_type=str]
```

Meridian itself logged a clean 200:
```
[PROXY] ... adapter=openai model=sonnet stream=false tools=0 ... msgs=user[text,image]
[PROXY] ... usage: input=451 output=538 cache=0%
```

So this lines up 1:1 with the root cause in the PR description — `response_format` is silently dropped on `/v1/chat/completions`, the model just answers conversationally, and any client using OpenAI Structured Outputs (not just custom tooling — stock SDK usage like Mealie's) breaks.

This blocks real functionality (Mealie's AI recipe scanning) for anyone running Meridian as a general-purpose OpenAI-compatible endpoint, not just custom agent tooling. Would be great to see this merged/released — happy to help verify against Mealie's flow once it lands if useful.

PR #900, opened 2026-08-30
fix(errors): classify org spend-limit refusals as rate limits

StanChmielewski:
## Problem

When an organization caps Extra Usage, Claude Code refuses with:

```
Claude Code returned an error result: You've hit your org's monthly spend limit ·
ask your admin to raise it at claude.ai/settings/usage
```

`HIT_YOUR_LIMIT = /hit your (?:[\w-]+ )?limit/` allows exactly one qualifier word, and `[\w-]` excludes the apostrophe — so `hit your org's monthly spend limit` matches nothing.

Consequences, in order:

1. `classifyError()` falls through to a generic `api_error`
2. `isAccountFailoverError()` returns false
3. the profile is never marked exhausted

With `MERIDIAN_ROUTING=priority` this is the bad case: `/profiles/list` reports `exhausted: []` while every request fails, and the pool keeps serving the dead account with a healthy one waiting behind it. Same class as #764 and #787 — one qualifier shape further out than the current pattern reaches.

## Why not just widen `HIT_YOUR_LIMIT`

Because `/hit your (?:[\w'-]+ ){0,4}limit/` also matches:

```
you have hit your configured tool call depth limit
```

which `errors.test.ts` already pins as a non-quota error ("does not treat unrelated 'limit' prose as a rate limit"). A wildcard qualifier can't distinguish a quota refusal from any other limit.

## Fix

Anchor on the *kind* of limit instead. `HIT_YOUR_SPEND_LIMIT` requires the word `spend` or `usage` immediately before `limit`, so an arbitrary number of qualifier words is fine without swallowing unrelated limits. Apostrophes (straight and typographic) are allowed for the possessive.

Two tests added for the live wordings: the `org's monthly spend limit` form and the shorter `monthly spend limit` one that the CLI emits when talking straight to the API.

`bun test src/__tests__/errors.test.ts` — 118 pass, 0 fail.

## Verification

Against a live refusal on a Claude Max org that had hit its cap:

before
```
exhausted: []                      # both accounts failing, pool sees nothing wrong
```

after
```
exhausted: [{ id: "max", until: <reset>, reason: "rate_limit_error" }]
profile=max(priority) → 429 → profile=pro(priority) → 200
```

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -620,6 +620,18 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     expect(r.status).toBe(429)
   })
 
+  it("maps the CLI's 'You've hit your org's monthly spend limit' to rate_limit_error", () => {
+    const r = classifyError("Claude Code returned an error result: You've hit your org's monthly spend limit · ask your admin to raise it at claude.ai/settings/usage")
+    expect(r.type).toBe("rate_limit_error")
+    expect(r.status).toBe(429)
+  })
+
+  it("maps the CLI's 'You've hit your monthly spend limit' to rate_limit_error", () => {
+    const r = classifyError("You've hit your monthly spend limit · raise it at claude.ai/settings/usage · your session limit resets 5pm (Europe/Warsaw)")
+    expect(r.type).toBe("rate_limit_error")
+    expect(r.status).toBe(429)
+  })
+
   it("maps the CLI's 'You're out of usage credits' to rate_limit_error without a same-profile retry", () => {
     const msg = "Claude Code returned an error result: You're out of usage credits. /model to switch models."
     const r = classifyError(msg)
--- a/src/proxy/errors.ts
+++ b/src/proxy/errors.ts
@@ -62,6 +62,19 @@ const BILLING_SIGNALS: readonly RegExp[] = [
  *  can't drift into unrelated text that happens to contain "limit". */
 const HIT_YOUR_LIMIT = /hit your (?:[\w-]+ )?limit/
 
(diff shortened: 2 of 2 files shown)


rynfar:
Thanks — this landed as your commit in #908, now on `main`.

`main` requires signed commits, so fork PRs sit at `mergeStateStatus: BLOCKED` with every check green and no indication why. Nothing you could fix from your side. I cherry-picked yours to preserve authorship and added two commits on top; GitHub's `Co-authored-by:` trailer carries your credit onto `main`.

Your reasoning for not just widening `HIT_YOUR_LIMIT` was the most useful part of the report, and it holds up — I reproduced the counter-example. Anchoring on the limit's *kind* rather than the qualifier count is the right shape, and the two live wordings you supplied were what made it verifiable.

What I added, both from review of the merged branch:

1. **Anchoring.** Four qualifier words is a much wider net than the single-word pattern, and unanchored it also matched `you have not hit your monthly spend limit yet` and the phrase merely quoted inside tool stderr. Those mark a *healthy* account exhausted — the same outage, from the other direction. `OUT_OF_USAGE_CREDITS` already anchors for exactly this reason.

2. **A correction to my own commit 1.** Anchoring to the start of the whole message made it blind to the shape the CLI uses most — exiting non-zero and appending the banner to stderr. That fall-through returned `401 "run claude login"` for a quota refusal, while the session-limit banner in the identical shape returned 429. Fixed by allowing `subprocess stderr` as a wrapper and anchoring per-line.

Reading the banner strings out of the bundled CLI binary also turned up two adjacent refusals with the same failure — `You've reached your Fable 5 limit.` and `Your group's usage limit is set to $0` — filed as #909. If you have a live refusal for either, that would help; the `reached` variant needs its wording pinned down before it is safe to widen.

Issue #901, opened 2026-08-31
Rate-limit responses carry no Retry-After, and [1m] benching is profile-wide — one RLM child's 429 cold-caches every concurrent sibling


rynfar:
Prime Agent's RLM harness runs N concurrent children against one Max account through Meridian. Telemetry from real prime traffic (2026-08-17/18) shows one upstream limit multiplying into coordinated failure:

- Each child independently runs the full retry ladder (strip `[1m]` → retry, then 2 exponential retries, plus up to 3 resume-refusal retries), and Meridian returns 429/503 **without a `Retry-After` header**, so client-side backoff (Prime Agent retries 4× on top, pylon-code/prime-agent#24) has nothing to coordinate against.
- `recordExtendedContextRateLimited` benches `[1m]` **profile-wide**: one child's rate limit downgrades every sibling session to 200k, and the model switch cold-caches each of them (observed: `cache_miss: 0% on resume` storms during the limit window).

Asks:

1. Emit `Retry-After` on 429/503 responses — propagate upstream's value when parsed (`oauthUsage.ts` already reads it), otherwise a computed backoff hint.
2. Scope extended-context benching per session (or at minimum make profile-wide benching opt-in), so one session's 1M rate limit doesn't invalidate the prompt cache of every concurrent sibling.
3. Consider distinguishing "account is rate limited" from "session limit hit" in the error body so harnesses can stop retrying the unfixable (Prime side tracked in pylon-code/prime-agent#24).

Umbrella with full evidence and the Prime-side plan: pylon-code/prime-agent#22.

rynfar:
Shipped in #907 (merged): Retry-After on 429/503 with upstream propagation and per-status defaults, and [1m] benching split by trigger — Extra Usage exhaustion stays profile-wide, plain rate limits are session-scoped. Item 3 deliberately skipped (the CLI's 'session limit' is the account window; the wait duration on the wire is the actionable part).

PR #907, opened 2026-08-31
feat: emit Retry-After and scope [1m] rate-limit benching per session

rynfar:
## Problem

Prime Agent's RLM harness runs N concurrent children against one Max account through Meridian. Meridian returned 429/503 with no `Retry-After`, so every child backed off on its own schedule, woke at the same time, and re-hammered a window that was still spent. Worse, a rate limit on an extended-context request benched `[1m]` for the *whole profile*: one child's 429 downgraded every concurrent sibling to the 200k model at the same instant, and the model switch cold-caches each of them, because their cached prefixes were built on the 1M model.

## The fix

**Retry-After on every throttle-shaped response.** A new pure leaf module, `src/proxy/retryAfter.ts`, computes the number in "most authoritative first" order: upstream's own `Retry-After` when it survived into the error, then a hint embedded in the upstream error text (`"retry-after": 30`, `Retry-After: 30`, `retry_after=30`), then the account's own observed window reset from `rateLimitStore`, then a per-status constant — 60s for a rate limit, 5s for overload. Every branch is clamped to `[1s, 24h]`, so no source can produce "retry immediately" or "retry never".

It is wired wherever an error response is built:

- non-streaming failures get a real `Retry-After` header (and a mirrored `error.retry_after` field);
- SSE turns get `error.retry_after` in the error frame — a stream's headers went out with `message_start`, long before the failure existed, so the header cannot reach the client there;
- priority routing's pool-exhausted response names the *pool's* earliest opening, not the last account tried;
- the drain 503, the durable-priority-state 503s, and the cross-process turn 529 all carry the short constant;
- `relayInnerError` forwards the header, so `/v1/responses` and `/v1/chat/completions` callers get it too.

`oauthUsage.ts`'s private `parseRetryAfterMs` moved into the new module rather than being duplicated.

**Bench scoping decision: split by trigger, not blanket session-scoping.** The two reasons to bench `[1m]` have genuinely different blast radii, and the original profile-wide choice was right for one of them:

- **Extra Usage exhaustion stays profile-wide.** `recordExtendedContextUnavailable` fires on "extra usage is required for 1M context" / "out of extra usage". That is an entitlement fact about the *subscription*: every session on the account would fail identically, so making each one prove it costs N failed requests and buys nothing.
- **A plain rate limit is now session-scoped.** `recordExtendedContextRateLimited` benches only the session that earned the 429, keyed by the client's session identity namespaced under its profile. If the account's window really is spent, every session still discovers that — one extra refused attempt each, once — instead of N simultaneous cache misses.

Clients with no session identity (fingerprint-based adapters) still bench profile-wide; there is nothing narrower to scope to, so their behavior is unchanged. Session entries are bounded and self-expiring.

**Item 3 (session-limit vs account-limit in the error body) was skipped.** The Claude CLI's "You've hit your session limit · resets <time>" wording does not mean a per-conversation cap — it is the account's five-hour usage window, which resets. Labeling it "session" in the error body would tell a harness the opposite of the truth, and the thing it actually needs (how long to wait) is now on the response. The genuinely unfixable class, out-of-usage-credits, already classifies distinctly.

## Verification

- `npm test` — exit 0. 3159 tests across the 13 invocations (`npm test` isolates the module-mocking files into their own `bun test` runs): 3158 pass, 0 fail.
- `npm run typecheck` — clean.

New tests: `src/__tests__/retry-after-unit.test.ts` (precedence, clamping, per-status defaults, parse forms, statuses that take no hint), Retry-After integration cases in `proxy-error-handling.test.ts` (header on 429/503, upstream propagation, silence on 402/500/504, SSE frame), session-scope cases in `models.test.ts`, and bench-scope assertions in `proxy-extra-usage-fallback.test.ts`.

Closes nothing on its own — see #901 for the remaining item.

---

Model: Claude Opus. Harness: Claude Agent SDK.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -45,6 +45,7 @@ src/
 │   │   └── forgecode.ts       ← ForgeCode adapter (fingerprint sessions, XML CWD, passthrough)
 │   ├── query.ts               ← SDK query options builder (shared between stream/non-stream paths)
 │   ├── errors.ts              ← Error classification (SDK errors → HTTP responses)
+│   ├── retryAfter.ts          ← Retry-After computation for 429/503/529 (PURE)
 │   ├── models.ts              ← Model mapping, Claude executable resolution
 │   ├── buildInfo.ts           ← Build provenance: source detection, semver compare (PURE)
 │   ├── updateCheck.ts         ← Cached npm registry lookup for the newest published version
@@ -95,6 +96,7 @@ server.ts (HTTP layer)
     ├── adapters/opencode.ts ──► messages.ts, session/fingerprint.ts, tools.ts
     ├── query.ts ──► adapter.ts, mcpTools.ts, passthroughTools.ts
     ├── errors.ts
+    ├── retryAfter.ts
     ├── models.ts
     ├── tools.ts
     ├── messages.ts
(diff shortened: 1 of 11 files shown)


rynfar:
Reviewed against #901 with an independent full-suite re-run in the worktree (`npm test` exit 0) and a read of the load-bearing seams.

The design decisions are right. Splitting the bench by trigger instead of blanket session-scoping is better than what the issue asked for: Extra Usage exhaustion genuinely is a subscription fact and stays profile-wide, while plain rate limits stop punishing siblings. Verified the scoping is coherent end to end — `benchSessionKey` comes from `adapter.getSessionId(c, body)` (for Prime Agent, the `metadata.user_id` identity) and the same value is used at both `mapModelToClaudeModel` read time and both `recordExtendedContextRateLimited` call sites (stream and non-stream), with NUL-separated profile namespacing and a bounded, self-expiring map. The Retry-After precedence chain (upstream header → embedded hint → observed window reset → per-status default, clamped to [1s, 24h]) lives in a pure leaf module per the architecture rules, and the pool-exhausted response naming the pool's earliest opening rather than the last account tried is a thoughtful touch.

The SSE in-frame `error.retry_after` (seconds) is now consumed on the harness side: pylon-code/prime-agent#30 gained a follow-up commit that prefers the in-frame value over response headers on mid-stream errors, so the contract is closed end to end — streamed rate limits through Meridian now carry an actionable wait into Prime Agent's kind-aware retry.

Agree with skipping item 3: the CLI's "session limit" is the account's five-hour window, and labeling it per-conversation would mislead harnesses; the wait duration now on the wire is the actionable part.

Reviewed by Fable 5 via Claude Agent SDK.

PR #920, opened 2026-09-01
fix(errors): classify the credits-era per-tier limit refusal

### LINKED ISSUE #909: Two live quota refusals still fall through to api_error with no failover
Found while reviewing #908. Two refusal wordings that ship in the Claude Code binary today are not classified as account-failover errors, so an account that hits either keeps serving in a priority pool while every request fails and `/profiles/list` reports `exhausted: []` — the same outage #908 fixes for the spend-limit wording.

Both strings are verbatim in `node_modules/@anthropic-ai/claude-code-darwin-arm64/claude`:

```
You've reached your Fable 5 limit.
Your group's usage limit is set to $0 · run /usage-credits to request more
```

Current behaviour:

| message | status | type | fails over |
|---|---|---|---|
| `You've reached your Fable 5 limit.` | 500 | `api_error` | no |
| `Your group's usage limit is set to $0 · run /usage-credits …` | 500 | `api_error` | no |

### Why each misses

**`reached your <model> limit`** — every pattern in `errors.ts` keys on the verb `hit`. `HIT_YOUR_LIMIT` also allows one qualifier word, so the two-token model name (`Fable 5`) would miss even with the right verb.

**`group's usage limit is set to $0`** — matches no rate-limit predicate and nothing in `BILLING_SIGNALS`. It is an entitlement cap rather than a quota window, so `billing_error` is probably the better classification: it fails over via `ACCOUNT_FAILOVER_ERROR_TYPES` without `isQuotaRefusal` sending the cooldown to look up a five-hour reset that does not apply.

### The care needed on the first one

Do not simply widen to `reached your (?:\w+ ){0,4}limit`. The binary also contains a templated `reached your specified …` whose noun does not resolve from a `strings` dump — plausibly a config limit like max turns or cost, which must not mark an account exhausted. That wording needs to be pinned down first, ideally from a live refusal.

### Related

There is a separate question about cooldown tiers: a *monthly* spend cap classified as `rate_limit_error` makes `isQuotaRefusal` true, so `priorityCooldownUntil` reads the account's five-hour window. The account comes back after ~5h and fails again for the rest of the month. Still far better than never failing over, but the cooldown is wrong for monthly caps.

justprosh:
Fixes #909.

`You've reached your Fable 5 limit` currently falls through as `500 api_error`. This change classifies it as `429 rate_limit_error`, so priority routing marks the profile exhausted and fails over to the next account.

`isRateLimitError` remains unchanged: retrying the same account cannot help.

The matcher is intentionally narrow:

- known model tiers, with an optional numeric version and `Claude` prefix;
- only the observed `/usage-credits` and `/model` suffixes;
- multiline banners only under an explicit `Subprocess stderr:` label;
- the known appended beta-warning line produced by captured stderr;
- full-message anchoring to reject quoted, negated, configured, or otherwise incidental prose.

The concurrent proxy test now drains request finalization and instance GC before changing its process-global session store. Requests remain headerless, so the test still covers weak-fingerprint CAS. The target file passed 500 fresh-process stress runs.

Known trade-off: exhaustion is currently profile-wide rather than model-tier-scoped, so a Fable refusal cools the profile for all models. Tier-scoped exhaustion is separate work.

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -730,6 +730,7 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     ["Opus tier", "You've reached your Opus limit"],
     ["versioned Sonnet tier", "You've reached your Sonnet 4.6 limit"],
     ["Claude-prefixed tier", "You've reached your Claude Opus 4.6 limit"],
+    ["space-separated model command", "You've reached your Fable 5 limit /model to switch models."],
     ["multiline subprocess stderr", "Claude Code process exited with code 1\nSubprocess stderr: You've reached your Fable 5 limit. Run /usage-credits to continue or switch models with /model."],
   ])("maps the credits-era per-tier %s to rate_limit_error", (_label, msg) => {
     const r = classifyError(msg)
@@ -749,6 +750,9 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     ["documentation sentence suffix", "Error: You've reached your Fable 5 limit. This is only a documentation example"],
     ["false assertion suffix", "You've reached your Fable 5 limit is false"],
     ["possessive threshold suffix", "You've reached your Fable 5 limit's configured warning threshold"],
+    ["joined run command", "You've reached your Fable 5 limitRun /usage-credits"],
+    ["joined usage command", "You've reached your Fable 5 limit/usage-credits"],
+    ["unspaced punctuated command", "You've reached your Fable 5 limit.Run /usage-credits"],
     ["generic error newline", "Error:\nYou've reached your Fable 5 limit."],
     ["SDK wrapper newline", "Claude Code returned an error result:\nYou've reached your Fable 5 limit."],
     ["unrelated appended stderr", "You've reached your Fable 5 limit.\nSubprocess stderr: unrelated tool output"],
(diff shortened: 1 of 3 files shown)


rynfar:
Thanks for this — the tier enumeration and the anti-false-positive work here were exactly right, and all of it shipped in #929 (`85d18d9c`) with your commits cherry-picked, so you're retained as author on the squash trailer.

I closed this one only because `main` requires signed commits, which means fork PRs sit at `mergeStateStatus: BLOCKED` no matter how green the checks are. Nothing you could fix from your side — the cherry-pick is our standard path in for outside contributions.

One change on top of your commits, for the record. The matcher was anchored to the end of the whole message, but `server.ts` appends captured stderr to `Error.message` before classification:

```js
error.message = `${error.message}\nSubprocess stderr: ${stderrOutput}`
```

so that anchor is unreachable in production. Two live shapes were still missing:

- banner followed by any appended stderr other than the one whitelisted beta warning → `500 api_error`, no failover;
- `stderrLines.join("\n")` labels only its *first* line, and on Team plans the harmless "custom betas" warning is emitted first — so the real banner lands on an unlabelled second line, falls through to the code-1 branch, and returns `401` telling the operator to run `claude login` for what is actually a quota refusal.

The fix was just moving the bound from the message to the line (`/m`, matching the `HIT_YOUR_SPEND_LIMIT` idiom right above it) and tolerating CRLF. Your suffix whitelist and tier enumeration are unchanged, and every negative case you wrote still misses — I kept all of them.

I also verified the patterns against the refusal builder in the shipped CLI (2.1.198) rather than against the issue text: it emits exactly two suffixes, one per interactive/non-interactive copy branch, and both are now pinned in tests. And I added the second wording from #909 (`Your group's usage limit is set to $0 · …`) as a `billing_error`, so the issue closes fully.

Appreciate the care in the original — the narrowness was the right instinct, since a false positive exhausts a healthy profile pool-wide.

Issue #917, opened 2026-08-31
Intermittent CI failures: concurrency tests fail fast, never the same one twice


rynfar:
#915 fixed the largest category of intermittent CI failures — tests that slept through the real rate-limit backoff and tipped over bun's 5s per-test timeout. That accounted for every failure with a `500Xms` duration.

A second, distinct category remains. These fail **fast**, so they are not timeouts:

| test | duration |
|---|---|
| `MCP server per-request lifecycle > should not share MCP server between concurrent requests` | 919ms |
| `Plugin integration — end-to-end via HTTP > loads an absolute-path plugin and applies its onRequest hook` | 2197ms |
| `Integration: Concurrent subagent requests > should handle 3 concurrent requests (parent + 2 subagents)` | 119ms |
| `Fingerprint resume: cross-project safety via lineage > resumes correctly after cross-project rejection` | 89ms |
| `Integration: Non-stream usage propagation > falls back to zeros when no SDK assistant message is produced` | 43ms |
| `queue telemetry under cancellation` | 226ms |
| `Stale UUID retry` | 582ms |

## What is known

- **Never the same test twice.** Across ~6 failing runs, no test failed in two of them.
- **All pass locally**, individually and in the full suite, repeatedly. `npm test` on `main` is 3211 pass / 0 fail.
- **Several are explicitly concurrency tests** — "should not share MCP server between *concurrent* requests", "handle 3 *concurrent* requests". That is the strongest signal in the data.
- One observed failure was concrete rather than a timeout: `falls back to zeros` returned **500** instead of 200.
- The main pass runs ~2900 tests across ~175 files in a single bun process, and 75 test files call `mock.module("@anthropic-ai/claude-agent-sdk")`, which is process-global in bun.

## Two candidate explanations, not yet distinguished

1. **CPU contention on the runner.** Timing-sensitive concurrency assertions lose races when the runner is loaded. Would explain why it is never the same test and why it never reproduces on a developer machine.
2. **Cross-file mock leakage.** 75 files racing on one process-global module mock; file ordering can differ between macOS and Linux, which would explain the platform split.

These are not mutually exclusive, and I have not designed an experiment that separates them. Filing before investigating so the evidence is not lost.

## Why it matters

[...]

Issue #450, opened 2026-04-29
RangeError: Invalid array length on cold-start requests with msgCount=0 (pi adapter)


jesuskannolis:
## Summary

Meridian crashes with `RangeError: Invalid array length` on every request that arrives with no messages in the body (`msgCount=0`). This happens reliably on cold-start sessions — the first request after an upstream client creates a fresh conversation.

## Repro

Send any anthropic-messages request to meridian with an empty `messages` array (or possibly with a system prompt only and no user turn — I haven't isolated which of these triggers it). With openclaw 2026.4.26's session-fresh path, this fires every time a new openclaw session sends its first turn.

## Stack

```
RangeError: Invalid array length
    at Object.start (file:///opt/homebrew/lib/node_modules/@rynfar/meridian/dist/cli-3azh7s3k.js:18117:102)
    at setupReadableStreamDefaultController (node:internal/webstreams/readablestream:2464:23)
    at setupReadableStreamDefaultControllerFromSource (node:internal/webstreams/readablestream:2496:3)
    at new ReadableStream (node:internal/webstreams/readablestream:279:7)
    at file:///opt/homebrew/lib/node_modules/@rynfar/meridian/dist/cli-3azh7s3k.js:18086:26
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async handleWithQueue (file:///opt/homebrew/lib/node_modules/@rynfar/meridian/dist/cli-3azh7s3k.js:18759:14)
    at async dispatch (file:///opt/homebrew/lib/node_modules/@rynfar/meridian/dist/cli-3azh7s3k.js:1573:17)
    at async dispatch (file:///opt/homebrew/lib/node_modules/@rynfar/meridian/dist/cli-3azh7s3k.js:1573:17)
    at async cors2 (file:///opt/homebrew/lib/node_modules/@rynfar/meridian/dist/cli-3azh7s3k.js:3147:5)
```

## Logged proxy line at the crash point

```
[PROXY] <uuid> adapter=pi model=opus[1m] stream=true tools=18 lineage=new session=new active=1/10 msgCount=0 msgs=
RangeError: Invalid array length
```

Always `msgCount=0` and `msgs=` (empty); always `lineage=new`.

## Frequency

In our environment over a ~12h window: 28 RangeError instances out of 592 total `adapter=pi` proxy calls (~5%). 100% correlation with `msgCount=0` cold-start requests; 0% on continuations. Continuation requests on the same adapter work fine and return real output.

[...]

PR #927, opened 2026-09-03
test: isolate the suite from the operator's meridian environment

materemias:
## Problem

Every runtime knob is read through `env()` — `MERIDIAN_<X>` with a `CLAUDE_PROXY_<X>` fallback — so anything a developer exports to run their own proxy silently reconfigures the code under test. The suite inherits it, and CI (which exports none of it) cannot see the difference.

Two cases observed on one machine, both from a normal operator environment:

- **`MERIDIAN_NO_FILE_CHANGES=1`** disables the PostToolUse file-change hook, so the three *"File change visibility: other adapters still track"* cases in `src/__tests__/proxy-file-changes.test.ts` fail on `main` — with no local change in sight.
- **`MERIDIAN_TELEMETRY_PERSIST=1`** makes the process-global telemetry store the real `~/.config/meridian/telemetry.db`. The suite is written against the in-memory store and calls `telemetryStore.clear()` / `diagnosticLog.clear()`, which issue `DELETE FROM metrics` / `DELETE FROM diagnostic_logs` — so running `npm test` **destroys every metric and diagnostic log the operator's own proxy has recorded**. Redirecting `MERIDIAN_CONFIG_DIR` does not help: `src/telemetry/index.ts` resolves its path from `env("TELEMETRY_DB")` and otherwise hard-codes `~/.config/meridian/telemetry.db`.

The second one is data loss, and it is silent.

## Fix

Strip the `MERIDIAN_*` / `CLAUDE_PROXY_*` namespace in `src/__tests__/preload.ts`, before any module import. The preload then sets the values the suite owns (config dir, session dir, SDK process gate) as it already did.

CI runs with none of these set, so this is precisely what makes a local run mean the same thing as a CI run. Tests that need a knob set it themselves, in-process, after the preload.

## Verification

`npm test` on this branch: all 13 invocations green (2982 + 251 tests, 0 fail). Before it, on the same machine, `main` reported 3 failures in `proxy-file-changes.test.ts` and wiped the local telemetry database.

### DIFF
--- a/src/__tests__/preload.ts
+++ b/src/__tests__/preload.ts
@@ -7,8 +7,25 @@ import { mkdirSync } from "node:fs"
 import { tmpdir } from "node:os"
 import { join } from "node:path"
 
-// Auth middleware reads this at request time; clear it so tests don't need API keys
-delete process.env.MERIDIAN_API_KEY
+// Take the operator's whole configuration namespace away from the suite.
+//
+// Every runtime knob is read through `env()`, i.e. `MERIDIAN_<X>` with a
+// `CLAUDE_PROXY_<X>` fallback, so anything a developer exports to run their own
+// proxy silently reconfigures the code under test. Two real examples from this
+// machine: `MERIDIAN_NO_FILE_CHANGES=1` turns off the PostToolUse hook and
+// fails the three "other adapters still track" cases in
+// proxy-file-changes.test.ts, and `MERIDIAN_TELEMETRY_PERSIST=1` makes the
+// global telemetry store the real ~/.config/meridian/telemetry.db — which the
+// suite then DELETEs, because tests call `telemetryStore.clear()` and
+// `diagnosticLog.clear()`. (Redirecting the config dir does not save it:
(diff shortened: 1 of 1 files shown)


rynfar:
Your `test` job is red for a reason that has nothing to do with this PR — please rebase onto `main` and it should clear.

The 11 failures here (`Retry-After (#901)` ×5, `Empty messages array (#450)`, `Passthrough tool_use` ×5) are #917, which was diagnosed and fixed earlier today in #935. The transcript ownership ceiling is bounded per session root, and the test preload pointed all ~179 files at a single root — so prepared transcripts accumulated until it filled, after which every request in the process returned `500 session transcript ownership backlog is full` regardless of which file issued it. That is why unrelated tests fail with impossible statuses, like a 429 test receiving 500.

This branch predates that fix, so it is still running against the broken harness. `git rebase origin/main` and push, and the cluster should disappear.

Apologies for the delay in getting you CI signal on this — the approval was held while the flake made every result untrustworthy.

PR #934, opened 2026-09-04
DO NOT MERGE: temporary #917 diagnostic

rynfar:
Throwaway draft to get a Linux CI run that prints the body of the 500 the #917 cluster returns.

Two `console.error` calls in `proxy-error-handling.test.ts` and `proxy-passthrough-tool-use.test.ts`, fired only when the status is already wrong. No source changes. **Close without merging** once the log is captured.

Context: three hypotheses have failed to reproduce on macOS (SDK mock leakage — shipped as #931 and did not fix it; partial `../proxy/models` stubs; rate-limit store leakage). The `[PROXY]` log shows the retry starting but never what it failed with, so this captures that.

### DIFF
--- a/src/__tests__/proxy-error-handling.test.ts
+++ b/src/__tests__/proxy-error-handling.test.ts
@@ -168,6 +168,10 @@ describe("Error classification", () => {
       const app = createTestApp()
       const res = await post(app, BASIC_REQUEST)
 
+      if (res.status !== 429) {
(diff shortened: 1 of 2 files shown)


rynfar:
Diagnostic captured what it was for. Both halves of the cluster returned:

```
500 {"type":"api_error","message":"session transcript ownership backlog is full"}
```

Root cause and fix in #935: the transcript ownership budget is bounded per session root, and the test preload pointed all ~179 files at a single root keyed only by pid. Closing unmerged.

Issue #946, opened 2026-09-05
Meridian sessions not resumable by host → full-transcript replay loop burns tokens


filipporovelli:
<h2>Summary</h2><p></p><p>When <strong>OpenCode (via the Meridian proxy)</strong> is used as the coding agent inside an</p><p>Orca worktree that <strong>also</strong> runs Orca background ("bg") jobs, the Orca background</p><p>agent repeatedly <strong>fails to resume</strong> its Claude Code session. On each failure it</p><p>falls back to <strong>injecting the entire prior conversation transcript</strong> (~200k–300k</p><p>characters) into a brand-new session as the first user message, then <strong>respawns</strong>.</p><p>This produces a loop of duplicate full-context runs and burns a very large number</p><p>of subscription tokens for a single task.</p><p></p><p>Native <strong>Claude Code</strong> and <strong>Codex</strong> sessions in the same setup resume normally and</p><p>never trigger this. The problem correlates <strong>only</strong> with OpenCode/Meridian.</p><p></p><h2>Environment</h2><p></p><ul><li>macOS (Darwin 25.6.0)</li><li>Orca (background job daemon; jobs template <code>bg</code>, backend <code>daemon</code>)</li><li>Claude Code CLI v2.1.259</li><li>OpenCode routed to a Claude subscription via the <strong>Meridian</strong> proxy</li></ul><p>  (<code>@rynfar/meridian</code>) + <code>opencode-with-claude</code> plugin, baseURL <code>http://127.0.0.1:3456</code></p><ul><li>Session managed by Meridian; Orca resumes background jobs with native</li></ul><p>  <code>claude --resume &lt;id&gt;</code></p><p></p><h2>Impact</h2><p></p><p>Observed on a single worktree on one day (2026-09-04). The same task looped in</p><p><strong>3 separate episodes</strong>:</p><p></p>
Episode | Time (local) | Respawns | Approx. weighted tokens*
-- | -- | -- | --
1 | ~09:22–09:50 | ~8 | ~3.4M
2 | ~10:10–10:11 | ~7 (mostly empty respawns) | ~0.1M
3 | ~14:04–14:33 | ~16 | ~4.1M
 

[...]

PR #947, opened 2026-09-05
fix: resume Claude Code trailing system turns

justprosh:
## Summary

- preserve passthrough checkpoint resume when Claude Code closes a complete tool-result delta with a trailing `system` reminder turn
- deliver the reminder as ordinary user text after the native `tool_result`, never as an SDK system prompt
- keep the exception scoped to the `claude-code` adapter and fail closed for leading, repeated, non-final, non-text, blank, partial, unknown, and split-echo forms

## Root cause

With Claude Code's `mid-conversation-system` feature enabled (beta `mid-conversation-system-2026-04-07`), the CLI sends the next tool continuation as a semantically stable prefix plus three messages:

```text
assistant[tool_use] -> user[tool_result] -> system[text]
```

The final message is a `<total_tokens>`-style mid-conversation system reminder. Lineage correctly classified the request as a continuation, but the later checkpoint validator rejected every `role=system` message, cleared `resume`, and replayed the complete history in a fresh SDK session.

## Client scope

The shape is gated by the **feature**, not by the client version: reproduced with the real CLI on 2.1.259 and 2.1.261 with the feature enabled. With the feature off the same clients fold the reminder into the user `tool_result` and the generic path resumes without any opt-in. The exception therefore stays limited to this exact shape rather than treating all Claude Code system messages as resumable.

## Live verification (real Agent SDK, real Claude Max)

A throwaway Meridian built from this branch's `dist/` on a Linux host, passthrough mode, one live Max profile, driven by the real `claude -p` (`ANTHROPIC_BASE_URL` at the proxy, `--allowedTools "Bash(printf:*)"`, Haiku, `mid-conversation-system` forced by env). Turn 2 on the wire was the same in every run: `user[string] → system[string] → assistant[thinking,tool_use] → user[tool_result] → system[text]` (the echo carries a thinking block; the grammar tolerates it and a unit test pins that). Verdict read from `/telemetry/requests`.

| build | runtime `claude-agent-sdk` / bundled `claude-code` | client CLI | turn 2 lineage | `isResume` | SDK session | answer |
|---|---|---|---|---|---|---|
| clean `upstream/main` | 0.2.119 / 2.1.259 | 2.1.259 | continuation | **false** | new (full replay) | confused by replay framing, fixture missing |
| this branch | 0.2.119 / 2.1.259 | 2.1.259 | continuation | **true** | resumed at checkpoint | exact fixture |
| this branch | 0.3.261 / 2.1.261 | 2.1.259 | continuation | **true** | resumed at checkpoint | exact fixture |
| this branch | 0.3.261 / 2.1.261 | 2.1.261 | continuation | **true** | resumed at checkpoint | exact fixture |

## Validation

- clean `upstream/main` with the test-only diff: the HTTP regression fails because `options.resume` is `undefined`
- focused unit and HTTP integration suites: `100 pass, 0 fail`; full `npm test`: `3166 pass, 0 fail`
- `npm run typecheck`, `npm run build`, `git diff --check`
- Linux CI, Windows smoke, Docker smoke/build
- independent review of adapter scoping, content ordering, and fail-closed guards; follow-ups landed: agent-neutral option name `allowTrailingSystemReminder` in the leaf helper (the `claude-code` gate lives only in `server.ts`), whitespace-only reminders fail closed, and new tests for a system message between two result turns, blank/non-string reminder content, an extra unknown tool_use id in the echo, two trailing reminders (fresh replay, `isResume=false`), and the same shape under a non-claude-code adapter (no opt-in)

This is related to the full-history replay symptom in #946, but it is a distinct direct Claude Code adapter failure and does not claim to fix that issue's OpenCode/Orca cross-tool session ownership case.

### DIFF
--- a/src/__tests__/passthrough-early-stop-integration.test.ts
+++ b/src/__tests__/passthrough-early-stop-integration.test.ts
@@ -165,6 +165,25 @@ async function post(app: any, body: any, sessionHeader = "es-session", extraHead
   }))
 }
 
+/** Live Claude Code request: session identity in metadata.user_id, claude-cli
+ * UA, and no x-opencode-session header. */
+async function postClaudeCode(app: any, body: any, sessionId: string, extraHeaders: Record<string, string> = {}) {
+  usedSessionKeys.add(sessionId)
+  return app.fetch(new Request("http://localhost/v1/messages", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+      "x-api-key": "dummy",
+      "user-agent": "claude-cli/2.1.259",
+      ...extraHeaders,
+    },
+    body: JSON.stringify({
+      metadata: { user_id: JSON.stringify({ session_id: sessionId }) },
+      ...body,
+    }),
+  }))
+}
+
 describe("Integration: passthrough early stop", () => {
   let app: any
   let savedPassthrough: string | undefined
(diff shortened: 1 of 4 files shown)


justprosh:
Updated the description with a live verification on the real Agent SDK and Claude Max (clean main vs this branch, SDK 0.2.119 and 0.3.261, clients 2.1.259 and 2.1.261): the trailing-system shape is gated by Claude Code's `mid-conversation-system` feature, not by client version, and the branch resumes the checkpoint in every combination while clean main falls back to a full replay. Review follow-ups (neutral option name, blank-reminder fail-closed, extra fail-closed/adapter-gate tests) are in the last commit; CI is green.

justprosh:
Superseded by #949, which incorporated this branch — thank you. The three later revisions that #949 does not carry (blank-reminder fail-closed, feature-gated wording instead of a version boundary, extra fail-closed tests) are in #951.

PR #951, opened 2026-09-05
fix: fail closed on blank trailing reminders and scope the gate to the feature

justprosh:
## Summary

Follow-up to #949, which incorporated #947. Three small things from the later revisions of that branch that #949 does not carry:

- **whitespace-only trailing reminders fail closed** — they passed the gate and would reach the API as blank text blocks, which it rejects with 400 on the resumed turn
- **comments state the actual gate** — the trailing-system shape comes from Claude Code's `mid-conversation-system` feature (beta `mid-conversation-system-2026-04-07`), not from a client version. Verified live on the real Agent SDK: claude-cli 2.1.261 with the feature on emits the same `assistant[thinking,tool_use] -> user[tool_result] -> system[text]` delta as 2.1.259 and resumes through this path; with the feature off both fold the reminder into the `tool_result`. The "newer clients (2.1.261) fold the reminder … and need no opt-in" sentence was an artifact of a probe run with the feature off
- **tests pin the remaining fail-closed edges** — a system message between two result turns, blank and non-string reminder content, an extra unknown tool_use id in the echo, two trailing reminders through HTTP (`resume` undefined, telemetry `isResume=false`), thinking/text blocks in the echo tolerated, two text blocks delivered in order

No behaviour change other than rejecting blank reminders. The option keeps its name.

## Validation

- focused unit + HTTP suites: `123 pass, 0 fail`; full `npm test`: 0 failures
- `npm run typecheck`, `git diff --check`
- live evidence for the feature-gate claim is in the description of #947 (real Agent SDK 0.2.119 and 0.3.261, clients 2.1.259 and 2.1.261, clean main vs fix)

### DIFF
--- a/src/__tests__/passthrough-early-stop-integration.test.ts
+++ b/src/__tests__/passthrough-early-stop-integration.test.ts
@@ -1286,6 +1286,92 @@ describe("Integration: passthrough early stop", () => {
     })
   }
 
+  // Fail-closed twin: a second trailing system breaks the one-reminder
+  // contract, so the continuation must fall back to a fresh replay.
+  it("stream: two trailing system reminders fail closed to a fresh replay", async () => {
+    const sessionId = `cc-delta-fc-${TEST_RUN_ID}`
+    const headers = { "x-meridian-agent": "claude-code", "x-opencode-session": sessionId }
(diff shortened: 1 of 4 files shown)


justprosh:
Already landed in #950 (82164d2) with the same tests and wording — closing as superseded. Thanks for folding it in.

PR #962, opened 2026-09-06
fix(errors): classify the per-tier refusal that ends in prose

### LINKED ISSUE #909: Two live quota refusals still fall through to api_error with no failover

justprosh:
`REACHED_YOUR_TIER_LIMIT` (1.67.0, the reimplementation of #920 that closed #909) accepts the banner only when the line ends at `limit.`, or continues with `/usage-credits` or `/model to switch models`. A Claude Max pool receives a fourth shape, and it is the only one that arrives there:

```
Claude Code returned an error result: You've reached your Fable limit. Switch to another model to continue.
```

The prefix is already anticipated; the suffix is prose rather than a slash command, so the refusal falls through to `500 api_error` and `ACCOUNT_FAILOVER_ERROR_TYPES` never sees it.

Observed live on 1.66.0 through a gateway fronting a pool of Claude Max profiles: five consecutive Fable turns, five identical refusals from the same profile, no failover — while three other profiles in the same pool still had Fable window left (29%, 79%, and one untouched). Re-checked against the pattern as released in 1.68.0 by lifting the regex out of the published bundle and running the real string through it:

```
false | Claude Code returned an error result: You've reached your Fable limit. Switch to another model to continue.
true  | You've reached your Fable 5 limit.
true  | You've reached your Fable 5 limit. /model to switch models.
```

The suffix is enumerated like the existing ones rather than admitted as a wildcard tail, because the negative cases turn on exactly that distinction: a documentation sentence continuing past the banner must not exhaust a healthy profile pool-wide. A new negative case pins that boundary next to the positives — `"...limit. Switch to another model to continue, the docs say, but the account is healthy"` still classifies as `api_error`.

The line anchor, the tier enumeration and the `/m` bound are untouched.

Verified: `bun test src/__tests__/errors.test.ts` — 194 pass, 0 fail; `tsc --noEmit` clean. The same patch was also built into a bundle and the released artifact re-checked, so the behaviour is confirmed on the compiled output, not only the source.

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -910,6 +910,13 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     ["SDK wrapper newline", "Claude Code returned an error result:\nYou've reached your Fable 5 limit."],
     ["trailing newline", "You've reached your Fable 5 limit.\n"],
     ["CRLF line ending", "You've reached your Fable 5 limit.\r\n"],
+    // The prose suffix, not a slash command. Observed live on 1.66.0 through a
+    // gateway fronting a Claude Max pool: every Fable turn came back in this
+    // exact shape, classified api_error, and failed nothing over while three
+    // other profiles in the pool had Fable window left.
+    ["prose switch suffix", "Claude Code returned an error result: You've reached your Fable limit. Switch to another model to continue."],
+    ["prose switch suffix, bare banner", "You've reached your Fable 5 limit. Switch to another model to continue."],
+    ["prose switch suffix without the trailing clause", "You've reached your Opus limit. Switch to another model."],
   ])("maps the credits-era per-tier %s to rate_limit_error", (_label, msg) => {
     const r = classifyError(msg)
     expect(r.type).toBe("rate_limit_error")
@@ -931,6 +938,9 @@ describe("classifyError: session/usage limit phrasings (live-observed)", () => {
     ["joined run command", "You've reached your Fable 5 limitRun /usage-credits"],
     ["joined usage command", "You've reached your Fable 5 limit/usage-credits"],
     ["unspaced punctuated command", "You've reached your Fable 5 limit.Run /usage-credits"],
+    // The prose suffix is enumerated, not a licence for any tail: a sentence
+    // that merely starts like it must still fall through.
+    ["prose switch suffix continuing into documentation", "You've reached your Fable 5 limit. Switch to another model to continue, the docs say, but the account is healthy"],
   ])("does not classify credits-era per-tier %s as a rate limit", (_label, msg) => {
     expect(classifyError(msg).type).toBe("api_error")
   })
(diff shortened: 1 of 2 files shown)


rynfar:
Incorporated in #974 and now on main as `94e88cf0`, with your commit
cherry-picked so your Author and AuthorDate are preserved, and a

Your diagnosis and your judgement call both held up. I re-derived the claim
against main before touching it: the prose-suffix string returns `false` from
the released pattern while the two known shapes return `true`. And enumerating
the suffix rather than allowing a wildcard tail was the right instinct — I
threw six negatives at it, including three you hadn't written, and the boundary
holds.

One thing worth telling you, because it affects the case you actually observed.
A new live gate (`scripts/e2e-tier-refusal-failover.mjs`, E2E.md E44) drives
the real failover path with a genuine Claude Max profile as the fallback. On
your commit alone it **fails**: the refused attempt is still recorded 500 and
nothing fails over.

The reason is the deployment you reported it from. A gateway or API-key profile
never delivers the banner bare — the SDK splices the upstream status in front:

```
Claude Code returned an error result: API Error: 400 You've reached your Fable limit. Switch to another model to continue.
```

That numeric status sits between the accepted wrappers and the banner and
defeated the line anchor for *every* suffix, including the two that already
worked. So the classifier returned `rate_limit_error` for the bare string while
a live request still 500'd — which is exactly the gap unit tests cannot show.
A separate maintainer commit allows exactly three digits immediately before the
banner; a 4-digit lookalike, a `Retried 3 times:` preamble, and a status ahead
of a non-quota qualifier all still fall through.

With both commits, the gate passes streaming and non-streaming: the refusing
profile records 429 and a real Claude Max profile answers.

Thanks for the report — the live 1.66.0 capture and the bundle re-check are what
made it actionable without guesswork.

PR #963, opened 2026-09-06
fix(codex): keep auto-defer off for Codex requests

justprosh:
## Problem

`codexTransforms` runs after the shared OpenCode transform and inherits its `coreToolNames` (`read, write, edit, bash, glob, grep`). Codex never sends tools with those names, so as soon as a Codex request crosses the auto-defer threshold (default 15 — trivially, once MCP namespaces are carried through, see #964) the core set matches nothing and **every** tool is deferred, `exec_command` and `apply_patch` included. Observed with Codex Desktop 0.153.4: `deferred=283/283 tools (core: read,write,edit,bash,glob,grep)`.

Deferral has a second, larger cost in passthrough: `computePassthroughMaxTurns` lifts the single-turn cap when `hasDeferredTools` is set (a ToolSearch discovery needs a turn), so every tool-calling turn also generates the SDK's discarded digest turn on the full context. Measured on a 680k-token Codex session: cache reads of 1×/2×/3× the context per turn (683k / 1349k / 2022k) with a 97–100% hit rate, versus one read per turn before MCP tools arrived. The SDK transcripts contain no `ToolSearch` call at all — the extra reads are digest turns, not discovery.

## Fix

The codex transform sets `coreToolNames: undefined`, which per `AgentAdapter.getCoreToolNames` disables auto-defer for this agent. Codex never sets `defer_loading` itself, so `hasDeferredTools` stays false, the cap stays at 1, and a tool turn is one model call again. Loading the ~250 namespace tool schemas costs ~125k cached tokens once (437k chars in the captured request) — far less than one extra full-context read per turn.

An earlier revision of this PR named Codex's built-ins as the core set instead; that kept `exec_command` loaded but still paid the digest turn on every tool call, so it was replaced.

## Tests

`src/__tests__/codex-transform-core-tools.test.ts` runs the OpenCode + codex transform pipeline with 40 tools and asserts passthrough is forced, no core set is exposed, and `createPassthroughMcpServer` reports `hasDeferredTools === false`. `bun test` and `tsc --noEmit` green.

Independent of, but most useful together with, #964.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/aleksei/dev/meridian/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/aleksei/dev/meridian/node_modules
--- a/src/__tests__/codex-transform-core-tools.test.ts
+++ b/src/__tests__/codex-transform-core-tools.test.ts
@@ -1,0 +1,28 @@
+/**
+ * The codex transform must name Codex's own built-ins as core tools: the
+ * inherited OpenCode list matches nothing Codex sends, so auto-defer deferred
+ * every tool (exec_command included) and each turn paid a discovery round.
+ */
+
+import { describe, it, expect } from "bun:test"
+import { codexTransforms } from "../proxy/transforms/codex"
+import { openCodeTransforms } from "../proxy/transforms/opencode"
+import type { RequestContext } from "../proxy/transform"
+
+function runCodexPipeline(body: Record<string, unknown>): RequestContext {
+  let ctx = { body, adapter: "codex" } as unknown as RequestContext
+  for (const t of [...openCodeTransforms, ...codexTransforms]) {
+    if ((t.adapters === undefined || t.adapters.includes("codex")) && t.onRequest) ctx = t.onRequest(ctx)
+  }
+  return ctx
+}
+
+describe("codex transform core tools", () => {
+  it("forces passthrough and names Codex built-ins as the always-loaded set", () => {
+    const ctx = runCodexPipeline({ model: "m", messages: [], tools: [{ name: "exec_command" }, { name: "mcp__iskron_bridge__iskron_orient" }] })
+    expect(ctx.passthrough).toBe(true)
+    expect(ctx.coreToolNames).toContain("exec_command")
+    expect(ctx.coreToolNames).toContain("apply_patch")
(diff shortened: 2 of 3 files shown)


rynfar:
Incorporated in #975 and now on main as `dabd969b`, with your final commit

Your analysis held up under independent checking. Your test fails on main with
exactly the leak you described, and I measured the consequence directly rather
than taking the numbers on faith: 40/40 tools deferred with `exec_command`
among them, and `maxTurns` going 1 → 4 because `singleTurnHandoff` requires
`!hasDeferredTools`.

Abandoning your first revision was the right call, for the reason you gave:
any non-empty core set still sets `hasDeferredTools` and still lifts the cap.
`coreToolNames: undefined` is also the documented idiom — `AgentAdapter`'s own
comment says "Return undefined to disable auto-defer for this agent" — and the
codex adapter defines no `getCoreToolNames`, so the leak was purely the
inherited transform and the fix belongs exactly where you put it.

There was no Codex E2E coverage in this repo at all, so your report also got
one: `scripts/e2e-codex-auto-defer.mjs` (E2E.md E45) drives a real proxy and
real SDK with 40 Codex-shaped tools. Pre-fix it reproduces your diagnostic
verbatim, including a detail worth having:

```
deferred=40/40 tools (core: read,write,edit,bash,glob,grep)
discovered=1 (exec_command) session_total=1
```

That second line means deferring `exec_command` also cost a ToolSearch round
trip before Codex's primary tool could be called.

One note on the digest-turn cost. I could not get it to show at probe scale —
a 40-tool request on a short prompt produced one model call both before and
after — so the gate reports the turn count rather than asserting it, and E2E.md
records why. Your 680k-token measurement is the evidence for that part; I just
didn't want a check that looks meaningful and discriminates nothing.

Finally, something for your own branches rather than this PR. The intermediate
commit `b9bca255` tracks a `node_modules` symlink pointing at
`/Users/aleksei/dev/meridian/node_modules`. Your follow-up `6781ff43` removes
it, so your PR head is clean — but anyone checking out or cherry-picking that
middle commit gets a dangling symlink over their own install. It replaced mine
before I noticed. Only your final commit was incorporated, so main is
unaffected. A `node_modules` line in `.gitignore` on that branch would prevent
the recurrence.

Thanks — the measurements and the fact that you'd already tried and rejected
the obvious alternative are what made this quick to land.

PR #964, opened 2026-09-06
fix(responses): carry Codex namespace and custom tools through to Claude

justprosh:
## Problem

Codex 0.15x (Desktop 0.153.4 captured) no longer sends MCP tools as flat `function` entries. Every MCP server arrives as one `{type:"namespace", name:"mcp__<server>", description, tools:[…]}` entry with the tool definitions nested inside, and its freeform `apply_patch` as `{type:"custom", format:{grammar}}`. `translateResponsesToAnthropic` keeps only `type:"function"`, so all of them are dropped silently: a Codex session on a Claude model sees none of its MCP servers (in the captured request, 25 top-level tools, 12 of them namespaces holding ~250 tools) and no `apply_patch`. The design spec scoped Phase 1 to function tools and this shape did not exist when it was verified against Codex 0.144.

Two more gaps surface as soon as an MCP tool is actually called:

- Codex expects the call back as `function_call {name, namespace}` (its router resolves `ToolName::new(namespace, name)`; a flattened name without `namespace` never matches), and `custom_tool_call {name, input}` for custom tools.
- `function_call_output.output` is a string for shell-style tools but an **array of content items** for MCP tools (`FunctionCallOutputBody` is an untagged `Text | ContentItems`). Passed verbatim it reaches Anthropic as `tool_result.content[0].type = "input_text"` → `400`, killing every turn after an MCP call.

## Fix

- Nested and custom tools are exposed to Claude under deterministic aliases (`<namespace>__<name>`, sanitized to `[A-Za-z0-9_-]`, hashed past Claude's 64-char limit minus the `mcp__oc__` passthrough prefix — a test pins the two constants together). Custom tools become a function with a single `input` string carrying the freeform payload; the grammar is folded into the description.
- `buildResponsesToolAliases(request.tools)` is pure and deterministic; the route builds it once and passes it in `ResponsesCtx`. Both emitters (non-stream and SSE) map an aliased `tool_use` back to `function_call {name, namespace}` or `custom_tool_call {name, namespace?, input}`; custom tools emit no `function_call_arguments.*` events (Codex takes the item from `output_item.done`). Replayed history items with a `namespace`, and `custom_tool_call` / `custom_tool_call_output`, land on the same aliases so the next turn's tool set matches its history.
- `function_call_output` content items map to `tool_result` blocks: `input_text` → text, data-URL `input_image` → image (reusing `parseDataUrlImage`), anything else → an omission note; an empty list → `""`.
- Server-side tool kinds (`web_search`) are still dropped — Claude has no client-side counterpart here.

## Verification

- `src/__tests__/openai-responses-namespaces.test.ts` (new): alias budget/sanitizing/hashing, alias table, request flattening, history replay, non-stream and SSE emitters for namespaced and custom tools, content-item outputs. Full `bun test` and `tsc --noEmit` green.
- Live: Codex Desktop 0.153.4 → Meridian → Claude calls an MCP tool (`iskron_realm` on a stdio server) and the turn completes; before the change the model had no MCP tools, and with only the first half of the change the turn after the call died with the `input_text` 400.

Deployed on a fork build since 2026-09-06. Pairs with #963 (Codex core tools for auto-defer) — without it, a Codex request with ~260 tools crosses the defer threshold with an empty core set.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/aleksei/dev/meridian/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/aleksei/dev/meridian/node_modules
--- a/src/__tests__/openai-responses-namespaces.test.ts
+++ b/src/__tests__/openai-responses-namespaces.test.ts
@@ -185,3 +185,46 @@ describe("createResponsesSseTranslator with aliases", () => {
     expect(done).toEqual({ type: "custom_tool_call", id: "fc_toolu_p", call_id: "toolu_p", name: "apply_patch", input: "*** Begin Patch\n*** End Patch\n", status: "completed" })
   })
 })
+
+describe("function_call_output content items", () => {
+  // MCP tool results come back from Codex as content items, not a string
+  // (FunctionCallOutputBody::ContentItems); a verbatim pass-through reached
+  // Anthropic as `tool_result.content[0].type = "input_text"` → 400.
+  it("maps input_text and data-URL input_image items to tool_result blocks", () => {
+    const png = "data:image/png;base64,iVBORw0KGgo="
+    const r = translateResponsesToAnthropic({
+      model: "m",
+      tools,
+      input: [
+        { type: "message", role: "user", content: [{ type: "input_text", text: "go" }] },
+        { type: "function_call", name: "iskron_orient", namespace: "mcp__iskron-bridge", arguments: "{}", call_id: "call_1" },
+        { type: "function_call_output", call_id: "call_1", output: [
+          { type: "input_text", text: "ГРАФ: merkazim [60]" },
+          { type: "input_image", image_url: png },
+          { type: "encrypted_content", encrypted_content: "opaque" },
+        ] },
+      ] as never,
+    })!
+    const result = (r.messages.at(-1)!.content as unknown as Array<Record<string, unknown>>).find((b) => b.type === "tool_result")!
+    expect(result.tool_use_id).toBe("call_1")
+    expect(result.content).toEqual([
+      { type: "text", text: "ГРАФ: merkazim [60]" },
+      { type: "image", source: { type: "base64", media_type: "image/png", data: "iVBORw0KGgo=" } },
+      { type: "text", text: "[Unsupported encrypted_content tool output part omitted]" },
+    ])
+  })
+
+  it("keeps string outputs as strings and turns an empty item list into an empty string", () => {
+    const r = translateResponsesToAnthropic({
+      model: "m",
+      input: [
+        { type: "function_call", name: "shell", arguments: "{}", call_id: "call_1" },
+        { type: "function_call_output", call_id: "call_1", output: "a.txt" },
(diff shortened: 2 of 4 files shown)


rynfar:
Incorporated in #976 and now on main as `907a00ee`, with all three of your
commits cherry-picked so your Author and AuthorDate are preserved, plus a

I captured the request rather than taking the shapes on faith: codex-cli
0.153.4 pointed at a recording endpoint with a real stdio MCP server attached,
isolated `CODEX_HOME`. It matches your description exactly:

```
by type: {"function":10,"namespace":2,"web_search":1}
 type=namespace  name=multi_agent_v1   nested=5
 type=namespace  name=mcp__iskron      nested=2
```

and through the translator, before and after:

```
ON MAIN     reached Claude: 10   namespaced tools present: NONE
WITH FIX    reached Claude: 17
```

Your report was understated in one respect worth recording: the drop is not
limited to a user's MCP servers. Codex's own `multi_agent_v1` namespace goes
with it, so `spawn_agent` / `wait_agent` were unavailable too — a Codex session
on Claude lost sub-agents silently.

The `function_call_output` content-item finding is the one I'd single out. It is
the difference between the feature being incomplete and being unusable, and it
is not something you would predict from the Responses schema — `output` being an
untagged `Text | ContentItems` only shows up when an MCP tool actually returns.
It has its own turn in the new gate.

Your alias design held up under review: deterministic, sanitized, hashed past
the budget, and `RESPONSES_TOOL_ALIAS_MAX = 64 - "mcp__oc__".length` pinned
against `PASSTHROUGH_MCP_PREFIX` by a test — the same duplicate-constant-plus-
cross-check pattern `passthroughEarlyStop.ts` uses on purpose. It also composes
correctly with #969, which landed earlier today: an MCP server literally named
`oc` yields aliases beginning `mcp__oc__`, and the passthrough layer now aliases
those rather than double-stripping them.

New gate: `scripts/e2e-codex-namespace-tools.mjs` (E2E.md E46), real proxy and
real SDK, both modes. Pre-fix it fails with `tools=1` and no call returned.

Two things I could not verify, recorded rather than glossed:

- **`apply_patch` as `{type:"custom"}` is not live-covered.** It did not appear
  in a default-config capture, so that path rests on your unit tests. E2E.md
  says so explicitly.
- A streaming check failed for me at first and it was my harness, not your
  emitter: `function_call` arguments arrive as delta events, so replaying the
  completed item alone gives an argument-less call and the model simply calls
  the tool again. I confirmed against the raw SSE that your emitter was correct
  before drawing any conclusion. Noted in the gate so it doesn't mislead the
  next reader.

Finally, the same branch-hygiene note as on your other PR: the middle commit
`432bead6` tracks a `node_modules` symlink to `/Users/aleksei/dev/meridian/node_modules`.
Your `73f4ddf5` removes it, so your head is clean, but cherry-picking through
that commit clobbers a local install. A `node_modules` entry in `.gitignore`
would stop it recurring.

Thanks — this was the most consequential of the five, and the captured request
plus the Rust router detail (`ToolName::new(namespace, name)`) is what made it
reviewable without guesswork.

Issue #655, opened 2026-07-17
Codex /v1/responses Phase 2: cross-turn reasoning continuity


rynfar:
Follow-up to #475 (PR #654), which shipped **Phase 1** of the OpenAI Responses API for Codex: text + function tool calling, streaming + non-streaming, forced passthrough — all live-verified on Codex 0.144.

## What Phase 1 deliberately defers

**Cross-turn reasoning continuity.** Phase 1 does not carry Claude's private thinking across Codex turns:

- Codex uses the Responses API's reasoning mechanism: the model emits `reasoning` items with an opaque, provider-encrypted `encrypted_content` blob, which the client echoes back next turn to restore the model's chain of thought. Only the originating provider can decrypt its own blobs.
- Claude produces **signed `thinking` blocks** (`{thinking, signature}`) — a different format and trust model. The two reasoning-persistence schemes are not interchangeable, so Claude's thinking can't round-trip through Codex's `encrypted_content` channel natively.

Phase 1's handling: `reasoning` items are dropped on input, and an **empty placeholder** `reasoning` item is emitted on output solely to stop Codex from hanging (it waits indefinitely for a reasoning item it requested — see PR #654). Each turn still reasons with full context (including tool results); only the *private* chain-of-thought is not persisted turn-to-turn.

## Phase 2 goal

Forward real reasoning continuity by packing Claude's `{thinking, signature}` into the `reasoning.encrypted_content` envelope (opaque to Codex, so it echoes verbatim), then unpacking on the next turn to restore a valid Claude thinking block.

## Open empirical questions (must verify before building)

1. **Does Codex echo the blob byte-for-byte?** If it re-encodes or truncates `encrypted_content`, the signature won't survive.
2. **Does Claude accept a reconstructed signed thinking block under fresh replay?** The Responses API is stateless (`store: false`), so every turn is a fresh replay with **no SDK session resume** — a signature minted in one turn's session may not validate when replayed standalone. This is the crux; needs a controlled probe.

## Fallback if signatures can't survive fresh-replay

Reasoning-as-plain-text: surface a human-readable reasoning summary (display only, no continuity). Strictly a UX nicety, not true continuity.

## Pointers

- Design spec: `docs/superpowers/specs/2026-07-08-codex-responses-api-design.md` (Phasing section)
- Phase 1 translation + placeholder: `src/proxy/openaiResponses.ts`
- `reasoningRequested()` already detects when Codex asked for reasoning — the hook point for Phase 2.

PR #965, opened 2026-09-06
fix(responses): key a Codex turn by its thread, not the parent's cache key

justprosh:
## What breaks today

Codex Desktop spawns a subagent as its own thread — own instructions, own tools, own conversation — but hands it the **parent's `prompt_cache_key`**. Since #655 that key is the whole identity `/v1/responses` has, so two live conversations land on one session key: the subagent's first turn rebinds the key to its own SDK session, and the parent's next turn then reads as a rewrite of it.

Observed live on Codex Desktop 0.144.4 against a Claude Max pool:

```
18:40:51  c602aaff  lineage=continuation  tools=226  msgCount=137   (46s, ok)
18:41:38  17889774  lineage=new           tools=91   msgCount=1     sessionWait=483ms
18:41:38  17889774  SESSION RECOVERY: previous conversation available. Run: claude --resume 032a5a78
18:41:45  58dce5f6  lineage=diverged                 msgCount=139   sessionWait=6360ms → 400
```

The `msgCount=1`/`tools=91` turn is the subagent; `032a5a78` is the SDK session the parent had just used, which is how the shared key shows up server-side. The parent's next turn queued 6.4s behind it, lost the session-turn race and was refused with *"This session advanced while the request was waiting."* When it wins the race instead, the outcome is quieter but worse: each side replays the full conversation into a fresh session at 0% cache.

The same shape has a second door: Codex runs **compaction** (`request_kind: compact`) as a request against the same thread id, with the whole history and no tools — its binary also names `title`, `summary`, `review`, `memory`. Keyed on the thread alone, a compaction lands on the live conversation's session as a rewrite of it; the conversation's next real turn then reads as `undo` and starts a fresh session at 0% cache. Reproduced through the gateway with a compaction-shaped request between two turns.

## The fix

Codex sends `x-codex-turn-metadata` both as a request header and inside `client_metadata`, and it names the thread and the request:

```json
{"session_id":"01a077c4-…","thread_id":"01a07806-…","thread_source":"subagent","request_kind":"turn"}
```

Three signals, each independently sufficient for its case:

- **`thread_id` keys the session.** For a thread the user drives it equals `prompt_cache_key`, so no existing session is re-anchored by this change; for a spawned thread it is distinct, which is exactly the case the cache key cannot express.
- **`thread_source` declares a concurrent flow** (`x-meridian-source: fork-codex-<source>`), so a spawned turn is admitted rather than refused even if a client ever reuses one id across flows.
- **`request_kind` other than `turn` gets its own key beside the thread's** (`<thread>:<kind>`) and declares its own flow (`fork-codex-<kind>`), so a compaction neither queues on the conversation's lease nor rebinds its session. A kind Codex has not sent yet is handled identically — the property that matters, not being the turn, is shared. A plain turn is keyed exactly as before.

`fork-` and not `subagent-` is deliberate: which tier a Codex subagent runs on is the client's choice, and a concurrency declaration must not silently rewrite the model mapping.

Clients that send no metadata (Codex ≤ 0.14x, anything else on this route) keep the `prompt_cache_key` path unchanged.

## Tests

- `resolveCodexThreadIdentity` unit tests in `openai-responses.test.ts` — header/body precedence, user vs. subagent thread, unparseable metadata, `thread_source` sanitizing, `request_kind` compact/turn/missing/unknown.
- `proxy-responses-endpoint.test.ts` — parent turn, subagent turn, parent turn again: the subagent gets its own SDK session and the parent still resumes the one it left; a user thread that sends metadata keeps resuming as before.

### DIFF
--- a/src/__tests__/openai-responses.test.ts
+++ b/src/__tests__/openai-responses.test.ts
@@ -10,6 +10,7 @@ import {
   translateResponsesToAnthropic,
   translateAnthropicToResponses,
   createResponsesSseTranslator,
+  resolveCodexThreadIdentity,
 } from "../proxy/openaiResponses"
 import type { AnthropicContentBlock } from "../proxy/openai"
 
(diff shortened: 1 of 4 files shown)


rynfar:
Incorporated in #977 and now on main as `03fe5716`, with both your commits
`Co-authored-by` trailer on the squashed commit.

I verified your central premise from a real capture rather than taking it on
trust: codex-cli 0.153.4 sends `thread_id` **equal to** `prompt_cache_key` for a
user-driven thread. That equality is the whole safety argument — keying on the
thread cannot re-anchor a session an existing client already established — and
it holds.

A new gate (`scripts/e2e-codex-thread-identity.mjs`, E2E.md E47) reproduces your
compaction case end to end. The discriminating measurement:

```
pre-fix   parent's turn after a compaction on its thread -> lineage=undo
with fix                                                 -> lineage=continuation
```

which is your reported symptom exactly. E41 all four modes pass alongside it.

Two things I recorded rather than glossed. A genuine `thread_source: subagent`
request could not be produced locally — `codex exec` offers
`multi_agent_v1.spawn_agent` but does not spawn, so that shape needs Desktop's
interactive flow; the gate uses the verified metadata wire format there, and the
subagent half rests on your Desktop 0.144.4 capture. And two of the gate's nine
checks are guards rather than discriminators: pre-fix, each sibling's first turn
also opened its own SDK session, so "the sibling got its own session" passes
either way.

Your PR needed a rebase against #976 and #978, which landed in
`openaiResponses.ts` first. Both conflicts were purely additive — tool aliases
and thread identity share no declarations — so both blocks were kept.

Thanks. The `request_kind` insight in particular is what made the compaction
half tractable; keying a side request beside the thread rather than on it is a
cleaner split than anything I would have reached for.

Issue #933, opened 2026-09-04
npm test is flaky on CI (~1 in 3 runs) and blocked the 1.67.0 npm publish


rynfar:
### Summary

`npm test` is nondeterministic on CI. It blocked the 1.67.0 npm publish: the release job created the tag and GitHub Release, then the `publish` job's `npm test` failed with 12 unrelated failures, so `npm publish` never ran. 1.67.0 was live everywhere except npm until the job was re-run.

### Evidence it's flaky, not a regression

A different, non-overlapping set of tests fails on each run of `main`:

| Commit | Failing tests |
|---|---|
| `ce7a9dc8` | SDK concurrency limiter, Error classification Retry-After (5), Empty messages array, Passthrough streaming/non-streaming tool_use |
| `d406404f` | Stale UUID retry, OpenCode title agent, graceful shutdown, Session lineage pruning, Extra usage fallback |
| earlier | Rate-limit retry backoff, Plugin integration e2e, MCP server per-request lifecycle |

The failures span concurrency, plugins, MCP, sessions, and error classification — no common subsystem. The same commit content passed CI on `6c85fcd5` and failed on `ce7a9dc8`, where the only diff is the version bump and changelog. The full suite passes locally (exit 0).

Recent `main` CI conclusions: `ce7a9dc8` fail, `6c85fcd5` pass, `d406404f` fail, `85d18d9c` pass, `1ea97d01` pass, `3ff4c0a5` pass, `0f38de84` pass, `00a2f82d` fail — roughly a 1-in-3 failure rate.

Note `d406404f` is #931 ("give every file its own module mocks instead of racing for one"), which was intended to address cross-file mock leakage and itself failed CI on `main`. So whatever remains is not fixed by per-file mocks alone.

Several failures are timing-shaped (`this test timed out after 5000ms`), suggesting runner contention or real cross-test interference rather than pure mock leakage.

### Why it matters

Releases are gated on this suite. A flaky gate on the publish step produces exactly the partial state seen here — tag and GitHub Release created, npm not published — which is silent unless someone checks `npm view`. It also trains contributors to dismiss red CI, which is how a real failure gets waved through (see the #921 PR body, where a contributor reported `proxy-stale-uuid-retry.test.ts` failing on a clean `origin/main` worktree and reasonably assumed it was pre-existing).

### Suggested directions

- Identify whether failures are ordering-dependent (does `--randomize` / a fixed seed reproduce them?)
[...]

PR #966, opened 2026-09-07
fix(responses): keep a mid-conversation developer message out of the system prompt

justprosh:
## What breaks today

Anthropic caches the prompt as one prefix in the order **tools → system → messages**. `/v1/responses` folds every `developer`/`system` item into the Anthropic `system` prompt, wherever it sits in the input. So a developer message that first appears mid-conversation rewrites the system block on that turn and invalidates every cached token behind the tools — the entire history.

Codex emits exactly such messages as ordinary conversation events: `<image_resize_notice>` after every `view_image`, `<model_switch>` and `<collaboration_mode>` on a model change, `<app-context>` when the app refreshes. On a long Codex Desktop thread against a Claude Max pool, every cache collapse over a day — 14 of 14 — followed one of these by one to three seconds, with the tools block (~125k) the only part still hitting and 240k–584k tokens re-written each time:

```
20:16:18  b5701fe5  msgs=343  cache=100%  read=697k  write=434
20:16:26  d7eb087d  msgs=345  cache=18%   read=125k  write=575k   <-- 2s after <image_resize_notice>
20:16:57  0e068f2a  msgs=347  cache=100%  read=700k  write=1k
```

An A/B through the proxy reproduces it: two identical conversations, the second turn of one carrying a developer note in the history.

```
A control      msgs=3  continuation  cache_read=30k  cache_write=58   cache=100%
B developer    msgs=3  continuation  cache_write=30k                  cache=0%
```

## The fix

A developer/system item that **precedes** the conversation is still the harness preamble and still folds into `system`. One that arrives **after** the conversation has started stays in the history as a user text block at its own position — the prefix before it remains cached and the turn costs only the note's own size. Codex replays the item at the same position on every later turn, so the placement is stable.

A tool result that lands in the same user message is filed ahead of the note, since tool results must lead their message (a developer note can fall between two outputs of one batch).

After the change the same A/B hits 100% on both sides.

## Tests

`openai-responses.test.ts`: leading developer/system items still fold (existing tests unchanged); a mid-conversation developer item is inlined at its position and absent from `system`; a tool result arriving after an inlined note is filed ahead of it.

### DIFF
--- a/src/__tests__/openai-responses.test.ts
+++ b/src/__tests__/openai-responses.test.ts
@@ -40,6 +40,66 @@ describe("translateResponsesToAnthropic", () => {
     expect(r.messages[0]!.role).toBe("user")
   })
 
+  // Anthropic caches the prompt as one prefix, tools → system → messages.
+  // A developer message that appears mid-conversation and gets folded into
+  // `system` rewrites the system block on the turn it first shows up and
+  // invalidates the entire cached history behind it. Observed live with
+  // Codex's `<image_resize_notice>`: 240k–584k tokens re-written per image.
+  it("keeps a mid-conversation developer message in the history, not in system", () => {
+    const r = translateResponsesToAnthropic({
+      model: "m",
+      instructions: "You are Codex.",
+      input: [
+        { type: "message", role: "developer", content: [{ type: "input_text", text: "House rules." }] },
+        { type: "message", role: "user", content: [{ type: "input_text", text: "look at the picture" }] },
+        { type: "message", role: "assistant", content: [{ type: "output_text", text: "Looking." }] },
+        { type: "message", role: "developer", content: [{ type: "input_text", text: "<image_resize_notice>resized</image_resize_notice>" }] },
+        { type: "message", role: "user", content: [{ type: "input_text", text: "and now?" }] },
+      ],
+    })!
+    expect(r.system).toBe("You are Codex.\n\nHouse rules.")
+    expect(r.system).not.toContain("image_resize_notice")
+    expect(r.messages).toEqual([
+      { role: "user", content: [{ type: "text", text: "look at the picture" }] },
+      { role: "assistant", content: [{ type: "text", text: "Looking." }] },
+      { role: "user", content: [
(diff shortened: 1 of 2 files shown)


rynfar:
Incorporated in #978 and now on main as `4a031b97`, with your substantive commit
`Co-authored-by` trailer on the squashed commit — this repo squashes with a blank
body, so that trailer is what keeps your credit attached. Your empty
`ci: retrigger` commit was not carried over.

Your safety split is the part I want to single out. I checked the premise
against a real capture rather than assuming it: codex-cli 0.153.4 sends the
harness preamble as a **leading** `developer` item
(`input roles: ["developer","user","user"]`), so keeping that folding into
`system` and inlining only later notes is exactly right — inlining the preamble
would have moved every existing client's cache prefix.

New gate: `scripts/e2e-responses-developer-cache.mjs` (E2E.md E48), an A/B with
real cache counters. Worth telling you how my first attempt at it was wrong,
because it bears on how your numbers should be read: I initially asserted hit
**percentage**, and it passed both before and after the fix. Hit rate does not
scale down to a probe — your collapse was on a ~700k thread where `system` sits
behind a ~125k tools block, so losing everything behind it cost 240k-584k, while
in a ~6.5k probe the same bug only moves the rate 98% → ~86%. Re-written tokens
are the invariant:

```
pre-fix   A cache_write=99    B cache_write=769   ->  7.8x   FAIL
with fix  A cache_write=117   B cache_write=141   ->  1.2x   PASS
```

So your A/B result reproduces; it just needs the right quantity to be visible at
small scale. E2E.md records that reasoning so the gate doesn't get "fixed" back
into a no-op later.

One more thing, unrelated to your change but useful to you: your retrigger note
— `failover-request-id timed out at 5s` — is a real, recurring flake, and you
are the second independent sighting. That test has no explicit `it` timeout, so
bun's default 5s applies and expires on a contended runner. I hit it twice today
on docs-only diffs and recorded it against #917 / #933 with the mechanism. You
were right that it is unrelated to your change.

Thanks — the tools → system → messages ordering argument and the timestamped
1-3s correlation are what made this diagnosable rather than a guess.

Issue #1009, opened 2026-09-10
Uncaptured-tool recovery for capped passthrough turns (deferred from #980)


rynfar:
Split out of contributor PR #980 while incorporating the Polytoken adapter. The adapter shipped separately; this commit did not.

## What it is

Contributor commit `f185e76e` by @jakewimmer, "feat(proxy): recover capped passthrough turns with streamed-uncaptured tools".

Reported failure shape, from the contributor's own deployment (sessions `46466398` / `ee5c8ac9`, 2026-09-10):

> an abort landing between stream completion and tool dispatch makes the CLI yield `max_turns_reached` WITHOUT running PreToolUse, leaving captures empty although every tool_use block fully streamed. The existing recovery requires `captured>0`, so those turns surfaced the raw cap error and killed the client turn.

The change adds an opt-in streaming recovery behind `MERIDIAN_PASSTHROUGH_UNCAPTURED_TOOL_RECOVERY=1`: a per-client-index tracker of genuinely forwarded starts/deltas/natural stops, with eligibility requiring `maxTurns=1`, no dropped/duplicate/forced-single/early-stop state, no cancellation of any kind, an open envelope, and every streamed block complete with a declared client tool name. Recovery evicts the session mapping before emitting the terminal `stop_reason:"tool_use"` pair and publishes no `resumeSessionAt` checkpoint.

## Why it was deferred, not rejected

- **It is the only commit in #980 that does not apply to current main.** It conflicts in `src/proxy/server.ts` with #998, which rewrote the same early-stop region after #983's regression (#996). Resolving that conflict is a real re-derivation against changed code, not a mechanical merge.
- **The contributor states it is not yet validated in production:** "Default OFF until canaried on the affected deployment." Shipping an unvalidated recovery path into the subsystem that just went through #983 → #996 → #998 warrants its own change and its own gate.
- **Non-streaming parity is deliberately deferred**, so the behavior would be asymmetric between response modes.
- It is unrelated to the Polytoken adapter and would have been squashed under that changelog entry.

The diagnostics groundwork it builds on (`abort=<cause>` in `sdk_termination`, contributor commit `016eb53c`) is also not yet merged — see the sibling split.

## What landing it needs

- Rebase onto current `main` and re-derive the `server.ts` hunks against post-#998 early-stop handling.
- A gate that reproduces the abort-window shape rather than asserting the recovery in isolation. The existing `scripts/e2e-capped-turns.mjs` is the natural place.
[...]

Issue #476, opened 2026-05-03
Multiple Different Passthrough Adapters


Serpentiel:
Passthrough adapter is nice, but it would be awesome if it'd be possible to define multiple ones, and switch between them e.g. via headers, e.g. `x-meridian-agent = passthrough-1`, `x-meridian-agent = passthrough-2`, etc. Or UUID based ones.

This way it would be possible to have multiple passthrough adapters with different settings, e.g. one that has some specific setting enabled, and another that doesn't have it.

Maybe it makes sense as well to have those adapters dynamically configurable instead of having a strict set pre-installed, so that by default there are none configured, or it is configured with "passthrough" adapter only. And then user is able to create new adapters from admin view, also with different flavors like OpenCode, Crush, LiteLLM, passthrough, etc. Or leave pre-defined ones, but make them removable. This way it could also be possible to define `MERIDIAN_PASSTHROUGH=1` mode per adapter instead, so that for OpenCode it's set to `1` and for another adapter, it's the other way, rather than globally how it is defined now.

rynfar:
Interesting idea, I like it. Let me churn on this a bit.  This is an excellent way to extend meridians capabilities.

Serpentiel:
Thanks @rynfar! I was also thinking that instead of hardcoding different flavors such as OpenCode, Crush, etc, maybe a dynamic framework could be introduced which allows to dynamically configure match rules for headers / `User-Agent` / etc.

rynfar:
You've got some excellent ideas @Serpentiel - these will both take a bit of work but I really do love the ideas.  Its becoming one of my complaints about having to test various harnesses, having to swap things around.  This is actively becoming frictional for me and something I need to solve.  I don't have a ton of time today to work on this, but I do tomorrow.  I will start digging into this!  Thank you for the ideas!

rynfar:
Shipped in #616 (merged, rides the next release) — @Serpentiel thank you for the proposal and especially the dynamic-match-rules refinement; both made it in, with credit.

What landed:

- **Named instances** of any base adapter with per-instance feature overrides: `~/.config/meridian/adapter-instances.json` or `MERIDIAN_ADAPTER_INSTANCES` (JSON)
- **Per-instance passthrough** — your `MERIDIAN_PASSTHROUGH=1` per-adapter ask, exactly: `"lite-plain": { "base": "passthrough", "passthrough": true }`
[...]

PR #1010, opened 2026-09-10
feat: add Polytoken harness adapter

rynfar:
Incorporates @jakewimmer's #980 — a native adapter for [Polytoken](https://polytoken.dev), an Anthropic-Messages coding agent that runs its own tool loop.

Contributor commits are cherry-picked with Author and AuthorDate preserved:

| original | here |
|---|---|
| `68a0092e` feat: add Polytoken native adapter | `6878e515` |
| `1bbf7a4b` docs: clarify resume behavior | `64fc6e68` |
| `f22856f4` docs: drop the live-loop qualifier | `639af5e2` |
| `9b7b21d4` docs: correct polytoken claims | `dd84557e` |
| `0edf2c63` fix(detect): match the Polytoken UA | `58518d61` |
| `76f30fa9` test(polytoken): pin contracts | `58e307a8` |
| `acc0c661` test: drain streaming responses | `2a878eea` |

## What is deliberately not here

#980 bundled three commits unrelated to the adapter. Splitting them keeps one changelog entry per PR and keeps a passthrough-subsystem change from hiding behind an adapter headline.

- `f185e76e` uncaptured-tool recovery — the only commit that does **not** apply to current `main`; it conflicts with #998's rework of the same early-stop region. The contributor states it is "default OFF until canaried on the affected deployment" and that non-streaming parity is deferred. Tracked in **#1009**.
- `1b2ba3a9` recover visible empty capped streams, and `016eb53c` classify abort causes in `sdk_termination`. Both apply cleanly and are green; they are held for their own PR so each gets its own changelog line and its own gate. Parked on `codex/polytoken-extras`.

The contributor's reported "1 failed" full-suite result does not reproduce: that flake was fixed by #997, which is now in the base.

## Maintainer correction: a runnable gate

#980's E2E section was a manual procedure — a Docker image swap, a personal systemd unit, and a budget gateway, driven by scripts the contributor deliberately did not commit. Nobody else could run it, and it overshot its own request budget (14 against a cap of 12).

Replaced with the mechanism this repository already has for this exact failure class. Polytoken is added to `scripts/e2e-client-detection.mjs` and its real 0.8.6 headers are recorded in `src/__tests__/fixtures/client-headers.json`, so `client-detection-fixtures` pins the adapter in CI and a client-side header change surfaces as a git diff. That is the #733 bug class — a client quietly changes what it sends and gets rerouted — and a PR whose detection keys on a User-Agent plus a native header is precisely what the fixture exists to protect. `x-polytoken-session` joins the redacted-value set, so a re-capture that only changes the session id produces no diff.

The live section now documents the loop actually exercised here: a disposable Meridian on an isolated port and a temporary Polytoken config directory. No Docker, no systemd, no shared origin.

## Validation

Live against the real **Polytoken 0.8.6** (macos-arm64, sha256 `71353a6d…0793e7`, verified against the published `SHA256SUMS.macos`) and real Claude Max on `claude-haiku-4-5`, through a disposable Meridian on port 3468 with an isolated session store.

**Client-owned tool loop.** `polytoken exec` reading a fixture file returned `LINES=4` — correct — in **four** client round-trips:

```
adapter=polytoken model=haiku stream=true tools=35 lineage=new          session=new
adapter=polytoken model=haiku stream=true tools=35 lineage=continuation session=446feb22
adapter=polytoken model=haiku stream=true tools=35 lineage=continuation session=2ab4b49b
adapter=polytoken model=haiku stream=true tools=35 lineage=continuation session=be371b2e
```

The read executed on the Polytoken side: the proxy's own working directory has no such fixture, so a proxy-executed read could not have produced that answer. Four round-trips rather than one is what shows the client drove the loop. Native-key resume works — `lineage=continuation` from turn 2 on a stable `x-polytoken-session`.

**Mandatory passthrough.** Repeated with `MERIDIAN_PASSTHROUGH=0` set globally: identical result, still four client round-trips, still `LINES=4`. The global setting cannot hand Polytoken's loop to the SDK.

**Detection controls**, all as documented:

| request | adapter |
|---|---|
| `user-agent: PolytokenImpostor/1.0` | `opencode` — token-boundary match rejects it |
| `x-polytoken-session: "   "` | `opencode` — whitespace is not a match |
| `x-polytoken-session: probe-sess-1` | `polytoken` |
| `user-agent: Polytoken/0.8.6`, no header | `polytoken` |

Real captured wire identity: `user-agent: Polytoken v0.8.6`, `x-polytoken-session`, `accept: text/event-stream`.

Gates: `npm test` **3871 pass / 0 fail / 1 pre-existing skip** (bun 1.3.14), `npm run typecheck`, `npm run build`.

## Behavior change worth knowing

Detection precedence changes for adapter instances (#476): a valid `x-polytoken-session` now outranks automatic instance match rules. Explicit `x-meridian-agent` selection still wins over both. A request carrying the native header that would previously have matched a configured instance now resolves to `polytoken`.

Closes #980


jakewimmer:
Thanks for putting this together! I switched my original PR to a draft after I started hitting oddities with tool calls, but I think my latest commits (that you included) address the issue.

PR #1030, opened 2026-09-11
fix(session): take the bookkeeping off the request path

justprosh:
## Symptom

On a long-lived single-proxy deployment (many conversations a day, ~500 turns/hour at peak) 13–16% of turns end in

> 504 — Request timed out. The operation may have been too complex. Try a simpler request.

Shrinking the request does not help, because the request was never the problem: the proxy's own session bookkeeping is. Four defects, each read off the running process, each independent, together compounding.

---

### 1. The deletion backlog never drains — and then blocks every new conversation

`deleteSession` **throws** when the transcript file is already gone: a success in disguise. The child exits 1, and the parent tries to recognise the case by matching `Session <id> not found` against `output.slice(-4_000)` — but a Node crash report fills that tail with minified vendor source and leaves the verdict at the head, so the match never fires.

Entries stay `retired` and are retried with backoff forever (**254 attempts** on one resource) until `pending` reaches `maxPending`. From that moment **every new conversation** fails with `ownership backlog is full`.

**Fix.** The child reports "nothing to delete" by **exit code** (`69`; `75` is the gate timeout), and the parent turns it into a typed `TranscriptAlreadyAbsentError` that the existing `notFound` path accepts — no prose re-parsed by a string matcher. Only the SDK's session-specific wording counts as absence, in the child exactly as in the parent (one shared phrase, passed to the child by env): a generic "not found" can be the SDK or the child failing to load, and stays a retryable failure rather than tombstoning a transcript still on disk. Child output now keeps **head and tail**, so the next verdict is not hidden by truncation either.

### 2. Every store read parses the whole file, synchronously — sometimes under the lock

`sessions.json` grows with conversation history: **26 MiB** on this deployment, three quarters of it per-message hashes. `JSON.parse` of it costs **135 ms**. `lookupSharedSessionResult` and friends do it several times per request, and the GC pin provider does it **while holding the lifecycle lock**, so every deletion claim blocks that lock for a full parse. This is where the 2 s lock budget goes.

**Fix.** An in-process read cache keyed by file *identity* — `{path, ino, mtimeMs, size}`, taken from the same fd the bytes are read from, so a rename landing mid-read can never pair one file's identity with another's content. Identity keying is exact because every writer publishes through `rename`, hence a new inode. Writers still parse fresh and republish the cache after `writeStore`.

Callers must treat the served document as immutable, so: the store copies locators in on write, and the two `server.ts` call sites that handed a **stored** locator into a lifecycle call that fills the caller's object back in place now pass a shallow copy.

### 3. An unarmed lease has no TTL — and fences the conversation until restart

`releaseActiveTranscriptLease` runs in a `finally` and itself waits on the lifecycle lock; when that wait times out, the lease survives. `pruneDeadActiveLeases` collects an unarmed lease only when its *owner* is dead — and the owner is the proxy itself, which is alive. The conversation then fails with `already has an active SDK writer` until a restart. Observed: **30 leaked leases, 28 older than 15 minutes**, none with a live executor.

**Fix.** Retire an unarmed lease once it is older than its TTL — the turn watchdog plus margin, as **its own knob**, not the prepared grace, which a deployment may deliberately size small. By the file's own invariant an unarmed lease never started a physical writer, so one that outlives the watchdog can only be a release that failed. Publication leases are covered too: in the same state they block retirement and deletion of their resource.

### 4. Pretty-printed bookkeeping files

`JSON.stringify(doc, null, 2)` on two machine-read-only files costs ~20–25% of the bytes and of the serialisation CPU spent under the lock: sidecar **1610 → 1216 KiB**, store stringify **72 → 48 ms**.

### And: stop blaming the request

Lock waits and backlog-full failures now answer **503 `overloaded_error`** naming the actual reason ("the retirement backlog is full", "a bookkeeping lock is busy") instead of a 504 that sends the operator to shrink a context unrelated to the failure. Absolute host paths stay out of the response body; the diagnostic log keeps them. New operator knob `SESSION_GC_LOCK_WAIT_MS` (default unchanged) for deployments that would rather wait than fail.

---

## Tests

Full suite green (`npm test`), `tsc --noEmit` clean. New coverage:

- **deletion verdict**, against a real child process: the exit-code absence lands as `notFound` and tombstones; a genuine child failure still lands as `failed` and retries;
- **output clipping**: the head survives;
- **read cache**: consecutive reads reuse the document; an in-process write is visible immediately; a store replaced out of band by another process is picked up; a deleted store is not served from the stale cache; no caller-owned locator is ever aliased into the cached document (neither through `storeSharedSession` nor `attachSharedTranscriptLocator`);
- **leases**: a stale unarmed lease no longer blocks `acquire` while a fresh one still does; a stale publication lease no longer blocks retirement.

### DIFF
--- a/src/__tests__/errors.test.ts
+++ b/src/__tests__/errors.test.ts
@@ -423,6 +423,37 @@ describe("classifyError", () => {
     })
   })
 
+  describe("session bookkeeping saturation", () => {
+    it("classifies the lifecycle lock wait as proxy load, not a request timeout", () => {
+      const result = classifyError("timed out waiting for /var/lib/meridian/.cache/meridian/session-gc.json.lock")
+      expect(result.status).toBe(503)
+      expect(result.type).toBe("overloaded_error")
+      expect(result.message).toContain("a bookkeeping lock is busy")
+      // The raw message carries absolute host paths; the client sees none.
+      expect(result.message).not.toContain("/var/lib")
+      expect(result.message).not.toContain(".lock")
+    })
+
+    it("classifies the ownership backlog limit as proxy load", () => {
+      const result = classifyError("session transcript ownership backlog is full")
+      expect(result.status).toBe(503)
+      expect(result.type).toBe("overloaded_error")
+      expect(result.message).toContain("the retirement backlog is full")
+    })
+
+    it("classifies the ownership capacity limit as proxy load", () => {
+      const result = classifyError("session transcript ownership capacity is full")
+      expect(result.status).toBe(503)
+      expect(result.type).toBe("overloaded_error")
+    })
+
+    it("still classifies an unrelated timeout as a request timeout", () => {
+      const result = classifyError("connection timed out")
+      expect(result.status).toBe(504)
+      expect(result.type).toBe("timeout_error")
+    })
+  })
+
   describe("server errors", () => {
     it("detects 500 status codes", () => {
       const result = classifyError("HTTP 500 from API")
--- a/src/__tests__/proxy-session-store.test.ts
+++ b/src/__tests__/proxy-session-store.test.ts
@@ -133,10 +133,8 @@ describe("Shared session store", () => {
   })
 
   it("never aliases a caller-owned locator into the cached document", () => {
-    // The read cache serves the document the last write published. If that
-    // document shared a locator object with its caller, any later mutation of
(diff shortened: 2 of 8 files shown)


rynfar:
Incorporated into main via #1036 (596a0d8363a88302e52f33841286e22d1c6177fc) with authorship preserved. Thank you @justprosh!

Issue #1024, opened 2026-09-11
OpenCode title + primary turn collide on one SDK session: "session advanced while the request was waiting"


calebdw:
> ## Maintainer status, 2026-09-11
>
> **Root cause: the Meridian plugin is not installed.** OpenCode's hidden
> title/summary agent runs under your own session id. Meridian tells it apart
> from your conversation using headers that **its own plugin** stamps;
> `opencode-with-claude` does not stamp them, so both concurrent streams look
> like one conversation and one loses the commit race.
>
> Reproduced with your models (Opus primary, Haiku title, ~1s apart, one session
> id):
>
> | setup | primary | title |
> |---|---|---|
> | no Meridian plugin headers (your shape) | 200 | **400** `session advanced while the request was waiting` |
> | Meridian's plugin headers present | 200 | 200 |
>
> With the plugin the title runs detached as `source=subagent-title agent=subagent`
> concurrently with `agent=primary`. **`meridian setup` is the fix for your
> machine**, and it should not disturb your `small_model` workaround.
>
> Your note that 1.62.1 works elsewhere fits: the stricter session-advance check
> landed later, so the same collision used to be absorbed as a cold replay
> instead of failing.
>
> **We also changed Meridian so this degrades instead of failing** — merged as
> `2e118a92` ([#1031](https://github.com/rynfar/meridian/pull/1031)), unreleased
> at the time of writing. A plugin-less OpenCode client is now treated the way
> the `pi` adapter already is: it has no per-flow signal to send, so its
> concurrent turns are admitted and the loser is reclassified rather than
> refused. Same reproduction after the change: **200/200**. Turns are still
> serialized one-at-a-time per session key and the loser still runs fresh rather
> than merging two histories; the remaining cost is a cold prompt cache, which
> the warning now names accurately.
>
> A plugin-equipped client keeps the strict 400, because for it a collision is a
> defect rather than an unavoidable protocol shape.
>
> **Kept open** pending @calebdw confirming on their machine. Two smaller points
> from the report are not addressed here: the hardcoded `:3456` fighting
> `opencode-with-claude`'s random port is theirs to fix, and in our reproduction
> the *title* took the 400 where theirs lost the Opus stream — which side loses
> is timing-dependent, the mechanism is identical.

### Summary

With OpenCode talking to Meridian as an Anthropic-compatible `baseURL`, the first message of a new session fails with:

```
AI_APICallError: This session advanced while the request was waiting.
[...]

PR #1031, opened 2026-09-11
fix(opencode): admit a plugin-less client's concurrent turns instead of refusing them

rynfar:
Stops a plugin-less OpenCode client failing the first turn of every new session.

Refs #1024, reported by @calebdw.

## The shape

OpenCode fires a Haiku `agent=title` stream and the primary turn about a second
apart on **one** OpenCode session id. With Meridian's plugin the hidden turn
says so in a header and is detached. A client running a different OpenCode
plugin sends no such header, so both streams look like one conversation, one
loses the commit race, and it took a hard 400:

```
This session advanced while the request was waiting.
```

Reproduced live on the 1.71.0 candidate with the reporter's own models, then
again here with the fix:

| headerless, Opus primary + Haiku title | before | after |
|---|---|---|
| primary | 200 | 200 |
| title | **400** | **200** |

## Why this shape, not a bigger change

This is precisely pi's situation — one session id for the main turn and its side
calls, no per-flow signal available — and the code already says so:

> An adapter can declare the same fact for its whole protocol when the client
> has no per-flow signal to send: pi carries one session id for the main turn
> and its side calls alike, so the loser of that race is reclassified rather
> than refused.

So this reuses that mechanism rather than inventing one, and applies it **only**
to requests that carry no plugin signal. A plugin-equipped OpenCode client keeps
the strict guard, because for it a collision is a defect rather than an
unavoidable protocol shape. Deliberately **not** done: setting
`runsConcurrentTurnsPerSessionKey` on the whole adapter, which would have
relaxed the guard for correctly configured users too.

The decision reads the same pure predicate the pluginless warning already
used — extracted as `isPluginlessOpenCodeRequest` — so the warning and the
behaviour cannot drift apart.

This is not inferring which stream is the title. That was tried and reverted
before, and the reasons are still in `pluginless-opencode-warning.test.ts`:
"tool-less, one message" is equally the first turn of an ordinary chat.

## What does not change

Serialization is untouched: one turn at a time per session key, `maxActiveQueries`
still 1. The loser still runs **fresh** rather than resuming the winner and
merging two histories. The cost is a cold prompt cache — slower, billed as
uncached input — which is what 1.62.1 did before the strict check existed, and
what the reporter experienced as "working".

The warning still fires, because `meridian setup` is still the real fix. Its
text now names the cost that remains instead of predicting a 400 that no longer
happens, and a test pins that it no longer says "400".

## Validation

`npm test` 3927 pass / 1 skip / 0 fail on bun 1.3.14 (+5), typecheck, build.

New coverage, all of it absent before:

- `isPluginlessOpenCodeRequest` — UA case-insensitivity, both agent modes, and
  negatives including `opencode2`, `crush`, `Polytoken` and `my-opencode/1.0`
  (prefix, not substring).
- HTTP-layer: a plugin-less concurrent pair is admitted, still serialized, loser
  runs fresh, and emits no `session_turn_conflict` telemetry. **Fails on the
  tree without this change** — verified by reverting the one-line condition.
- HTTP-layer control: with `x-opencode-agent-mode` present the loser still takes
  the 400 and still records the conflict. Passes both with and without the fix,
  which is what proves the change is scoped.

Live, real SDK and Claude Max: the reported shape now returns 200/200 with the
warning still emitted; the plugin-equipped control is unchanged at 200/200 with
no warning; all four E41 modes pass.

### DIFF
--- a/src/__tests__/pluginless-opencode-warning.test.ts
+++ b/src/__tests__/pluginless-opencode-warning.test.ts
@@ -11,7 +11,9 @@
  *   seq 2  tools=10  msgs=1  x-session-affinity: ses_fe4c…   ← the user's turn
  *
  * Both attempts of that run returned HTTP 400 `session_turn_conflict` after a
- * ~8s wait. The agent-scoped session key that fixes this reads the plugin's
+ * ~8s wait. Since #1024 that collision is admitted and reclassified instead —
+ * the loser replays against a cold cache — so the exposure is a cost rather
+ * than a failure. The warning still fires, because the fix is still the plugin. The agent-scoped session key that fixes this reads the plugin's
  * `x-opencode-agent-mode`, so it cannot reach a client that sends none.
  *
  * Inferring the agent from request shape was tried and reverted — "tool-less,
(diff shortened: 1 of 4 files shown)


calebdw:
Thanks for putting a PR together---but what do you mean by "plugin-less" opencode?

PR #1041, opened 2026-09-14
fix(session): recognize structured Environment CWD format, strip system-reminder noise from fingerprinting

reForMatt:
## What

Three fixes to Droid session fingerprinting, all addressing the same user-visible symptom: `ListAgents`/`Workflow` tools appearing on one turn and vanishing on the next within the same conversation.

1. **`extractDroidCwd()` recognizes the current structured `# Environment` block.** It previously only matched the old shell-transcript style (`% pwd\n/path`). Current Droid CLI builds emit `Primary working directory: /path` inside a structured `# Environment` block instead, so extraction silently returned `undefined` for essentially all current sessions. Both formats are matched now, so older stored sessions and clients pinned to an older CLI build keep resolving correctly.

2. **`<system-reminder>` blocks are stripped before hashing the conversation fingerprint.** These blocks are client-injected, per-machine environment noise (tool schemas, skills, connectors) — not conversation identity. They're identical across conversations in the same project and routinely exceeded the 2000-char hash window, so distinct conversations collided on one fingerprint key.

3. **Fingerprint seeding walks forward to the first task-bearing user message.** Droid opens with a reminder-only environment message and delivers the actual request in a later user message. Seeding from the first user message conflated every conversation in a project onto one session key. Now the seed comes from the first user message whose reminder-stripped text is non-empty; a history where every user message is reminder-only keeps the old raw-first-message behavior.

## Why

A missing/colliding fingerprint degrades session identity resolution and forces `reason=unrelated-history` full-history replays. A fresh SDK session from that replay gets a freshly-resolved deferred-tool manifest — that's why tools like `ListAgents` could appear on one turn and be gone on the next within what the user experienced as one conversation.

Confirmed against a live `meridian.err.log` (adapter=droid): one fingerprint key was observed serving both a 33-message and a 2-message history, and reproduced synthetically with a reminder-dominated hash window.

## Validation

- New unit tests: 4 fingerprint tests (system-reminder stripping) + 3 fingerprint tests (droid-shaped seeding) + droid-adapter tests for the structured `# Environment` CWD format
- Full `npm test`: 3953 pass / 0 fail (final run; ran green across all 3 commits)
- Typecheck + build clean across all 3 commits
- Live E2E on a local daemon with the real Droid binary and captured wire-format request bodies:
  - Same-cwd, different-task requests that previously logged `reason=unrelated-history` now resolve as clean new sessions
  - A two-turn droid-shaped conversation now resumes as `lineage=continuation`
  - Fingerprint resume works correctly under both the old and new CWD formats
- A/B against pristine `main` (evidence log: `meridian-ab-evidence.log`)
- Live soak on the daemon with monitoring (`~/bin/meridian-cache-report`): low-hit-mains near zero, writes only on `lineage=new` — the healthy signature

## Caveat

External cross-model review was attempted but all three critics failed on infra this session (rejected / auth-failed / no-output). The independence pass behind this PR is the author's own A/B control against pristine `main` plus a fresh-redo QA pass, not an external verdict. Flagging this so reviewers weight it accordingly.

## Known limitations (pre-existing or by design, not regressions introduced here)

- Editing the opening message of a conversation still starts a fresh session (client-side behavior, not something this fix touches).
- `droid exec` repeat-prompts intentionally start fresh — by design.
- Compaction that summarizes away the opening prompt still costs one fresh replay — pre-existing class of issue, not introduced or fixed here.
- Title-call payload writes remain the floor for what triggers a session write.


rynfar:
Incorporated into main via #1052 with contributor authorship and commit metadata preserved. Thank you @reForMatt!

PR #1042, opened 2026-09-14
fix(session): strip system-reminder noise from lineage hashing

reForMatt:
**Depends on #1041** — this branch is stacked on top of `fix/droid-cwd-structured-env-format`, so the diff below includes that PR's 3 commits until #1041 merges. Only the last commit (`4c2ac62`, "strip system-reminder noise from lineage hashing") is new here.

## What

Strips client-injected `<system-reminder>` blocks before hashing message content in `hashMessage()` / `semanticContent()` (`src/proxy/session/lineage.ts`), mirroring the stripping `getConversationFingerprint()` (`fingerprint.ts`) already does.

## Why

`fingerprint.ts` decides which cached session bucket a conversation lands in. `lineage.ts` decides whether that bucket's stored history can actually resume. Before this fix, only the first job stripped reminder noise — `hashMessage`/`semanticContent` hashed the raw text, including dynamic per-request content (file diagnostics, git status, tool manifests).

`measurePrefixOverlap()` breaks at the first hash mismatch, and message 0 of a Droid conversation is typically reminder-only. If that reminder's dynamic content shifts between turns — which it does, by design, every turn — message 0 never hashes the same way twice, prefix overlap collapses to 0 forever, and the conversation is classified `diverged: unrelated-history` on every single turn: a full-history replay every turn, indefinitely, even though it had already been bucketed to the correct session key.

This is the direct sibling of the fingerprint bug fixed in `fix/droid-cwd-structured-env-format`: that PR fixes which bucket a conversation lands in; this one fixes whether the bucket, once correct, can actually be used.

## Evidence

Live `meridian.err.log` on the running fork (after the fingerprint fix, before this one): session key `9e62d8f3` stayed stable across 29+ consecutive turns (fingerprint bucketing working correctly) but every single turn still logged prefix overlap `0/N` and `diverged: unrelated-history` (lineage resume still broken). Other keys with reminder content that happened not to shift between turns resumed fine as `lineage=continuation` — consistent with a bug that only fires when the reminder's dynamic payload actually changes, which is why it survived the fingerprint fix's own tests.

## Validation

- New test file `lineage-reminder-noise.test.ts`: 5/5 pass, covering:
  - identical hash for reminder-only messages whose reminder payload differs
  - reminder noise contributing nothing to the hash once stripped (task text still discriminates)
  - plain-string content with an inline reminder
  - `verifyLineage()` staying resumable when only the reminder payload changes between turns
  - `measurePrefixOverlap()` reaching full overlap across a growing conversation despite a changing reminder
- All fingerprint-adjacent suites re-run clean: `degraded-fingerprint-warning`, `fingerprint-unit`, `session-fingerprint-context` — 42/42 pass
- Full `bun test`: 3981 tests, 15 failures — all in 5 files unrelated to this change (`models-auth-status`, `sticky-routing-integration`, `session-lifecycle-process`, `mcp-grep-tool`, `fix-bun-exports`); reproduced the same failures running those files in isolation and traced them to pre-existing cross-test mock leakage (a missing `node:child_process.spawnSync` export, an auth-status mock bleeding into an unrelated grep-tool test) — confirmed unrelated to `lineage.ts`/`fingerprint.ts`
- `tsc --noEmit`: clean
- Build: all bundled entry points pass `node --check` (using the real Homebrew `node`, not a shim)

## Caveat

Same as the sibling PR: external cross-model review wasn't available this session. This is an author A/B/isolation pass, not an independent verdict.


rynfar:
Incorporated into main via #1052 with contributor authorship and commit metadata preserved. Thank you @reForMatt!

PR #1048, opened 2026-09-16
fix(openai): resume a session-keyed chat completions conversation instead of packing its history

justprosh:
## What it fixes

`/v1/chat/completions` packs every prior turn into a `<conversation_history>` system block and runs each request on a fresh SDK session; only Jcode (#784) keeps its real messages and resumes. For a stateless OpenAI-compatible client that *does* carry a stable conversation key this means the whole history is re-sent as fresh system text on every request, the prompt cache never sees a continuation, and every tool step of a multi-step turn is a fresh session. Measured behind a gateway that maps one `x-session-id` per conversation thread onto `x-session-affinity`: the cache share sat at the tool-definition floor (~60%) and fell with thread length.

## What changes

- When the generic `openai` adapter resolves a session key (`x-opencode-session` ?? `x-session-affinity`, scoped by `x-opencode-agent-name` under `x-opencode-agent-mode: subagent` — the `getSessionId` it already inherits from the OpenCode adapter), the handler keeps the real messages and forwards exactly those headers plus `user-agent` on the internal `/v1/messages` hop, so the inner handler resolves the same key, verifies lineage and applies the pluginless-OpenCode degradation (#1024) as it does on the native route. The adapter tag stays `openai`; Jcode's branch is byte-identical.
- Consecutive `tool` results of one assistant turn coalesce into a single user message — the shape the Messages API expects and the only one `coalesceCompleteToolResultContinuation` accepts — so a turn answering parallel tool calls resumes at the checkpoint instead of replaying.
- Unkeyed requests keep the packing. One visible difference for them: the packed request's final message now carries the whole result batch as real `tool_result` blocks instead of one result folded into the history text.
- Docs: the keyed contract in `docs/agents.md` (serialized turns per key, append-only history, an identical-body retry classifies `replayed-request` and replays fresh, the 400 a lost race returns) and a README row.

## Two things for the maintainer to decide

1. `x-session-affinity` becomes conversation identity at this endpoint for every client that sends it, not only OpenCode-family ones. `detect.ts` treats it as a generic stickiness header (Crush sends it), so a front proxy that sets one affinity per *client* rather than per conversation would now share one lineage across conversations and, with overlapping turns, meet `session_turn_conflict`. The docs say so. If you would rather gate it — an opt-in env, or accepting `x-session-affinity` only under an OpenCode user-agent — say which and I will change it.
2. One-time cost for in-flight keyed conversations (Jcode) whose stored lineage was hashed on the old split-result shape: the first post-upgrade turn carrying a parallel-result batch classifies `modified-history` or `undo-gap` (one fresh replay) or `compaction` (resumes correctly). No path reaches a 400.

## Validation

- `npm run typecheck`; `npm test` green except two pre-existing `session-lifecycle-windows-gc` EPERM failures on this machine (identical on `main`).
- New tests fail on the base tree: the keyed resume tests, the user-agent forwarding test and the subagent-scoping test in `proxy-openai-compat.test.ts`; the coalescing unit test in `openai.test.ts`; the three keyed tool-loop tests (single result, two parallel results, streaming) in `passthrough-early-stop-integration.test.ts`. The parallel one was additionally run with only the translator change stashed: it fails with `resume` undefined, so the coalescing is load-bearing.
- Live: deployed as a fork build behind our gateway; cache-share numbers from the journal will follow, which is why this is a draft.


rynfar:
Thank you @justprosh! Incorporated into `main` via #1055 with your commit and authorship preserved, conflict in README resolved, and proven live against Claude Max.

